PR #198, opened 2014-11-11
Accept string functions for output.file filename parameter

Signez:
Sometimes, it may be cool to have different filenames depending on other things than metadata.

With this patch, we can simply use a function as name, that will be evaluated on output start. We can still use metadata or time interpolation in the string returned.

This patch use `to_string_getter` which is already used in [operators/video_text.ml](https://github.com/savonet/liquidsoap/blob/3118f0299b894f4f24f92e14e868bd92d898adc9/src/operators/video_text.ml#L158).

**Rationale:** At [SynopsLive](http://synopslive.net/), we want to create new records file on demand. These records should have an understandable name, composed by a show slug given by an external process and timestamp of record beginning. To do this with current liquidsoap, we had to inject metadata, but it have weird side effects: as metadata is not known when output is starded, we need to do this: 
1. Use `reopen_with_metadata=true`,
2. Start the output,
3. Inject metadata using `insert_metadata`,
4. Ignore the fact that in the meantime, Liquidsoap created a near-empty file with a wrong name.

**Disclamer:** I never wrote any Ocaml before, so feel free to nitpick. However, this compile and run successfully at home.

### DIFF
--- a/src/outputs/pipe_output.ml
+++ b/src/outputs/pipe_output.ml
@@ -42,7 +42,7 @@ let pipe_proto kind arg_doc =
        Some "Encoding format." ;
 
        "",
-       Lang.string_t,
+       Lang.string_getter_t 1,
        None,
        Some (arg_doc ^
              " Some strftime conversion specifiers are available: \
@@ -80,7 +80,8 @@ class virtual piped_output p =
           raise (Lang.Invalid_value (format_val,"Unsupported format"))
   in
   let source = Lang.assoc "" 3 p in
-  let name = Lang.to_string (Lang.assoc "" 2 p) in
+  let name = Lang.to_string_getter (Lang.assoc "" 2 p) in
+  let name = name () in
   let kind = Encoder.kind_of_format format in
 object (self)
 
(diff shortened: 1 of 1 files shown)


toots:
Hi @Signez ! I like the idea of this PR but I have to think about it because the reload & metadata mechanism already make it complex to follow what's going on with file title.. Maybe also @dbaelde would have something to say about this.

dhannyz:
@toots i think this one would simplify my process of saving DJ live shows as an archive quite a lot.
I currently use a few different pieces of info for creating the archive file name, including the ice-name header, dj name, connection time, station name. On top of that some handling for momentary disconnects so that a dropout reconnection merges the archive into the same file.
Currently im relying on some dynamic source stuff, storing a reference to the source's stop function, relying on `on_start` and `on_stop` of a `output.dummy` to create the `output.file`

toots:
I agree that the metadata update mechanism sucks. I'll see what I can do before the release.

toots:
Merging this. Thanks for the contribution @Signez and sorry it took so long to include it, must be the longest running PR we've ever merged..

Signez:
Haha! Glad that it was useful after all.

A few bits of trivia: the webradio I created this feature for (and I mentioned in the PR description), SynopsLive, closed 4 years ago – but I am still recommending Liquidsoap to anyone that ask me how we were able to run it!

Liquidsoap is a great piece of software and I hope its development never ends. \o/

toots:
Thank you!

PR #206, opened 2015-01-04
add example to string.escape documentation

mcfiredrill:
I propose we add inline examples to the documentation for using the functions. Reading a lot of posts on the mailing lists and talking to people on IRC, it seems like users could really benefit from just seeing an example of the function being used right there. It would be easier to make a connection from the function signature to how to actually use it.

I am trying to add an example to the documentation comment, I'm not sure exactly how this will look when generated.

My inspiration from this is admittedly the rubyonrails docs, where they usually have examples inline api.rubyonrails.org

### DIFF
--- a/src/lang/lang_builtins.ml
+++ b/src/lang/lang_builtins.ml
@@ -286,7 +286,7 @@ let () =
        "mimes",Lang.list_t (Lang.string_t),
        Some (Lang.list ~t:Lang.string_t []),
        Some "List of mime types supported by this decoder \
-             for decoding streams."; 
+             for decoding streams.";
        test_arg;
        "",Lang.string_t,None,Some "Process to start."]
       Lang.unit_t
@@ -663,7 +663,9 @@ let () =
                       ~descr:"Escape special charaters in a \
                               string. String is parsed char by char. \
                               See @string.utf8.escape@ for an UTF8-aware \
-                              parsing function."
+                              parsing function.\n\n \
+                              Example:\n \
+                              meta = string.escape(meta)"
                       ~escape ~escape_char:Utils.escape_char ;
   let escape ~special_char ~escape_char f s =
     Utils.escape_utf8 ~special_char ~escape_char f s


bradisbell:
Examples would be great!  I've been messing around with Liquidsoap for a couple years and the lack of examples (and sometimes even full descriptions) makes it hard to get started.  Examples provide context of usage as well as samples for syntax.  Along those lines, I think the examples should show expected output for a given input.  Your example for `string.escape` is good if you already know what `string.escape` does, but does not provide enough context for someone who does not.  Maybe something along the lines of:

```
meta = "Test string line 1
Test string line 2"
print string.escape(meta) #Prints: Test string line 1\nTest string line 2
```

There should also be a second example in this case for the alternative usage of `string.escape` where a list of pairs is used for replacement.

I would contribute to the documentation myself, but unfortunately I do not have the expert viewpoint which would be required.

Moonbase59:
+1 for examples … please. I'm also new to LS and fighting all the time, trying to find code snippets on the net and the like. Concise short examples would be just great!

dbaelde:
Thanks for the suggestion, @mcfiredrill. If you run "make doc" you should be able to check how it looks in HTML, and if you run "liquidsoap -h string.escape" check how it looks in raw text. If you add markup in your example (i.e. @meta = string.escape(meta)@) it'll be nicer in HTML but will be displayed with markup in raw text.

I agree with @bradisbell that your proposed example is not so helpful yet, though. @bradisbell I do not think you need to be an expert to add useful examples in the documentation. In fact this is the kind of contribution where "normal" users would be most useful.

mcfiredrill:
Thanks @dbaelde, I'll check it out.
I agree its not really good enough yet, I will add some more information.

Right, honestly it sounds like @bradisbell knows enough to add examples. :smile:

bradisbell:
I never did figure out what `string.escape` did, specifically.

mcfiredrill:
I think a lot of the tests in `script/tests` might be a good starting point to create some examples.....

toots:
I'm closing this. We'll start reworking the doc next month.

PR #216, opened 2015-05-09
Added support for multiple source shoutcast protocol extension.

toots:


src/outputs/icecast2.ml
```diff
@@ -338,7 +344,11 @@ class output ~kind p =
 
   let host = s "host" in
   let port = e Lang.to_int "port" in
-  let user = s "user" in
+  let user =
+    match protocol, s "user" with
+      | Cry.Http _, "" -> "source"
+       | _, user -> user
```
dbaelde: Broken indentation ;)

src/outputs/icecast2.ml
```diff
@@ -305,18 +308,21 @@ class output ~kind p =
...
+      | Cry.Http _, name, mount when name = no_name && mount = no_mount ->
         raise (Lang.Invalid_value
                  (List.assoc "mount" p,
-                  "Either name or mount must be defined"))
-      else
-        mount
-    else
-      name
+                  "Either name or mount must be defined for icecast sources."))
+      | Cry.Icy, name, _ when name = no_name -> Printf.sprintf "sc#%i" icy_id
```
dbaelde: This doesn't seem to be backwards compatible: does it make sense if the protocol is icy but not v2?

toots: Yeah. I have just relaxed the constraints when using `icy` so that you don't have to pass either a `mount` or `name`. Now, you should simply be able to do `ouput.icecast(protocol="icy",...)` and omit `mount`, `name` and `icy_id`.

Issue #203, opened 2014-12-18
playlist.reload() dos not reload empty playlist


giflw:
When I create a playlist using a file (playlist.pls) and try to reload an empty playlist, liquidsoap keep playing old files instead of loading the empty playlist. The same applies to directory playlists (#196).

When I make a playlist empty I need that it plays blank until I reload it (if it is the only source) or be skeeped (if it is in a multiple sources configuration, like rotate).

Those are my expected results:

Single source (playlist should play blank when empty:

```
pls = playlist("playlist.pls")
out(pls)
```

Multiple sources (playlist should skipped when empty (or skip its turns)):

```
songs = playlist("songs.m3u")
jingles = playlist("jingles.pls")
s = rotate(weights=[1,3], [jingles, songs])
out(s)
```

In multiple sources configuration, if all playlist are empty, it should play blank.

toots:
So, a fallible playlist should accept empty playlists on reload. See #218 for that. If you need the source to play blank when receiving an empty playlist, you will need use `mksafe()`.

dbaelde:
Note that the playlist is fallible here so the proposed change won't be useful. The usual workaround, if one really wants an empty playlist is to make it non-empty but containing a broken URI, e.g., "/path/to/nowhere.ogg". I admit this is ugly and I am open to discussing a change of behavior, but I wonder if we have the time this deserves before a release.

PR #218, opened 2015-05-13
Accept empty playlists when reloading fallible playlists.

toots:
Partial fix for issue #203


toots:
@dbaelde: doest that look good to you?

dbaelde:
Your proposal preserves infallibility, so that's good. However it changes a behavior on which people may be relying currently: on a fallible playlist (most of the cases) a failed reload used to have no effect, now it empties the playlist. This should not be taken lightly. I don't have a clear picture of how broken playlists appear in practice, though. Moreover, some new features (reload on change) could justify a change of behavior, pushing people to use the new features where appropriate.

giflw:
I can use an example to make more clear to usefulness of this kind of behavior.

Suppose one have two `playlist` in a  `rotate`: one for music and one for spots. Those playlist are marked to reload each round. So, as I know, in each rotation they will try to reload. Suppose there are no spots to play during some hours, so a empty playlist will be given. After that time window, new spots appear, and than, spots are played again.

To not change current behavior, could be added a parameter like `accept_empty` with `false` as default.

dbaelde:
@giflw did you consider using a dummy URI (e.g. "/path/to/nowhere") to obtain an empty playlist. As dumb as it sounds, I suspect several people use it since I've been mentioning it several times over the years. It might be preferable that you also use this workaround, instead of changing the behavior of a very important source. A new and cleaner behavior should be reserved for a major release, for which backwards compatibility is not promised.

toots:
I think this behavior makes total sense to me and is more natural. I'm merging this so we can start testing it.

PR #219, opened 2015-05-14
Cleanup deprecated APIs.

toots:


smimram:
Just to make sure, is it supposed to still compile with older versions of OCaml?

toots:
Yeah, it should compile with all version above `3.12.1`. You guys okay with making `3.12.1` the minimal supported version for now?

PR #227, opened 2015-05-28
Fix parens.

zackp30:
I *think* this works (it compiles for me at least). I had tried several OCaml versions.

Disclaimer: I don't know OCaml myself.


toots:
Thanks!

zackp30:
You're welcome! :smiley:

PR #242, opened 2015-08-06
Update to more recent inotify.

smimram:
Apparently the "init" function was renamed to "create"...

I discovered this while compiling with opam. The problem here is that Debian still has 1.0, so it would require an update there.


smimram:
Just to be clear: with this commit we are not Debian-compatible anymore, but we become opam-compatible. @dbaelde @toots can I merge?

Issue #233, opened 2015-06-24
Installing Liquid Soap on Jessie with libaacplus-ocaml-dynlink failed


sdt105:
Hi,

i've installed Liquid Soap on Debian Jessie via apt-get (V 1.1.1). After that i added deb-multimedia.org in my sources.list file and installed libaacplus-ocaml-dynlink.

But when i'm using aacplus in liquid soap now, the aacplus encoder cannot be found ("No encoder found for that format").

What's wrong with that and is it possible to install Liquid Soap on Debian Jessie without the need for recompile?

Thanks for an answer!

Edit: version of libaacplus-ocaml-dynlink is 0.2.1

shortyxs:
Hi there,
I'm recentyl Meeting this Problem and can't finde any solution. It seems anybody who met this Problem before, gave up working on it. Is there any way to get %aacplus or %fdk-aac working on jessie?`
I checked the Folders, that liquidsoap log mentions and it seems all required files are presen.

Thanks and best regards
Malte

toots:
Hi @shortyxs I would recommend installing liquidsoap with `opam` as describe here: http://liquidsoap.fm/download.html Then, if your Debian distribution is recent enough (`>= jessie`), you should be able to compile liquidsoap with fdkaac enabled using the `opam` `fdkaac` package and the Debian `libfdk-aac-dev` package.

shortyxs:
Hi @toots ,
I am system administrator not a software developer, so I have no clue how to compile an application. Why can't I just apt-get install as any other application?
As mentioned already, I have the exact same problem as the threat srarter. I also did the exact same steps on the exact same versions and I did that before I found there is an open issue here on Github.

I tried the opam guide, but any opam command fails with an error.

PR #256, opened 2015-08-27
Native avi support

smimram:
This is nice and enables external for video too, see #233. For instance, we can encode avi with
```
output.file(%external(process="ffmpeg -i pipe:0 -f avi pipe:1",video=true),"/tmp/test.avi",s)
```

### DIFF
--- a/src/encoder/encoder.ml
+++ b/src/encoder/encoder.ml
@@ -85,6 +85,26 @@ struct
 
 end
 
+module AVI =
+struct
+  type t =
+    {
+      samplerate : int;
+      channels : int;
+      duration : float option
+    }
+
+  let to_string w =
+    let duration = 
+      match w.duration with
+        | None -> ""
+        | Some d -> Printf.sprintf ",duration=%f" d
+    in
+    Printf.sprintf
+      "%%wav(samplerate=%d,channels=%d,%s)"
+      w.samplerate w.channels duration
+end
+
 module Vorbis =
 struct
 
@@ -457,6 +477,7 @@ struct
   type t = {
     channels            : int ;
     samplerate          : int ;
+    video               : bool ;
     header              : bool ;
     restart_on_crash    : bool ;
     restart             : restart_condition ;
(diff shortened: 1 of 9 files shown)


src/encoder/encoder.ml
```diff
@@ -85,6 +85,20 @@ struct
...
+struct
+  type t =
+    {
+      samplerate : int;
+      channels : int;
+    }
+
+  let to_string w =
+    Printf.sprintf
+      "%%wav(samplerate=%d,channels=%d)"
```
dbaelde: Typo here: it should be `%avi`.

src/encoder/encoder.ml
```diff
@@ -470,10 +485,11 @@ struct
         | Metadata        -> "restart_on_metadata"
         | No_condition    -> ""
     in
-    Printf.sprintf "%%external(channels=%i,samplerate=%i,header=%s,\
+    Printf.sprintf "%%external(channels=%i,samplerate=%i,video=%s,header=%s,\
                               restart_on_crash=%s,%s,process=%s)"
       e.channels
       e.samplerate
+      (string_of_bool e.video)
```
dbaelde: Why don't we use `%b` instead of `string_of_bool` here?

src/encoder/external_encoder.ml
```diff
@@ -143,45 +150,45 @@ let encoder id ext =
...
         h.log#f 3
           "Error while reading data from encoding process: %s"
           (Printexc.to_string e) ;
         stop (); 
         (if h.params.restart_on_crash then
-          reset_process start_process h
+            reset_process start_process h
          else
-          raise e);
+            raise e);
```
dbaelde: The changeset would have been easier to read without those whitespace changes mixed with actual stuff. Plus, the new indentation on this line does not make any sense!

src/encoder/external_encoder.ml
```diff
@@ -193,41 +200,49 @@ let encoder id ext =
...
-    let start = Frame.audio_of_master start in
-    let b = AFrame.content_of_type ~channels frame start in
-    let len = Frame.audio_of_master len in
-    (* Resample if needed. *)
-    let b,start,len =
-      if ratio = 1. then
-        b,start,len
+    let sbuf =
+      if h.params.video then
+        Avi_encoder.encode_frame ~channels ~samplerate:h.params.samplerate ~converter:h.converter frame start len
```
dbaelde: This line is too long. Too long for my terminal but also for the split diff view of github: one more reason to keep 80 char lines for readability!

dbaelde:
I did not test the feature at all but only had a quick look at the code. I found one typo and complained about whitespace. Other than that the changes seem completely backward compatible so as far as I'm concerned it's OK to merge. Thanks!

PR #281, opened 2016-01-06
Enhance external requests file extensions

toots:
Issue a preliminary `HEAD` request to fetch content-type, if provided. Switch to `curl` for consistency.

This is a little bit involved so not sure it should go into the immediate next release.

@dbaelde, @smimram thoughts?


bradisbell:
This seems like a hacky solution.  Technically speaking, there is no guarantee that the response for a request will be of the same type for a subsequent request.  But, this is definitely better than the old method of detecting based on extension.  I'm not familiar with the internals... is there no way to inspect the response header when the real `GET` request for media occurs?

toots:
Actually by definition, the `HEAD` request must return the same data as the `GET` request but without the request body: http://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html

```
The HEAD method is identical to GET except that the server MUST NOT return a message-body in the response. The metainformation contained in the HTTP headers in response to a HEAD request SHOULD be identical to the information sent in response to a GET request. This method can be used for obtaining metainformation about the entity implied by the request without transferring the entity-body itself.
```

My concern is that this is a rather big change to use a different binary and also that the `HEAD` request would always happen before checking for an obvious file extension. Maybe I'll revisit that part..

bradisbell:
I realize by spec that's what should happen, but I'm thinking of the case where a resource may disappear in between the `HEAD` and `GET` request.  An edge case for sure... I'm only suggesting that if there is a way to get the `Content-Type` response header at the time of `GET` request, then that's what I think should be done.

Regarding file name extension, I think that file name extension should be ignored completely.  The `Content-Type` header should be deemed authoritative when it comes to HTTP.  Now, if you get `application/octet-stream` back, sure, do what you have to do to figure out the content type from the file name where possible.

I think you are right though that switching binaries is a breaking change.  A good one, but a breaking change none the less.  Not everyone has `curl` installed.

toots:
I think the `HEAD` request is a good compromise to find the file's extension. It is being used the way it was intended to by the `HTTP` specs, i.e. to prepare required ressources ahead of the actual request.

You're right on how `Content-Type` should be authoritative. In this case, we have a correct implementation of the file extension I believe.

As for the change of binary, `curl` is quite common these days, it's provided by default on OSX and very common with other OSes. Still, not a trivial change indeed.

toots:
And, of course, now reading the code I realize that we already had a solution in place.. Hugh!

Would you like to try adding this to your script:

```
set("extproto.use_server_name", true)
```

And see if that works for you.

toots:
Not sure I really like this `use_server_name` anymore tho hehe..

toots:
Gonna merge this. I think it's a much better solution, let's test it now!

Issue #284, opened 2016-01-07
Add external.input.video


smimram:
Now that there is native AVI support, see #256, we should add `input.external.video` (and the same for output), so that the following would work for instance:
```
s = input.external.video("ffmpeg -f lavfi -i testsrc -t 5 -f avi pipe:1")
output.sdl(faillible=true,s)
```
Right now `input.external` only supports audio.

toots:
Whoa. I totally missed the avi encoder PR, nice!

smimram:
Thanks! It won't be for next release, though.

smimram:
This should be fixed with the merge 68ddd47ac17bbda082532447bbf070aba01403ea.

PR #285, opened 2016-01-08
Preliminary implementation of AVI decoder.

smimram:
Implement an AVI decoder so that we can do
```
s = input.external.video("ffmpeg -f lavfi -i testsrc -f avi -an -vcodec rawvideo pipe:1")
output.sdl(fallible=true, s)
```
as requested in #284. Some debugging and fine-tuning is still needed on this PR.


smimram:
This is actually working with my latest patch on ocaml-mm!

PR #326, opened 2016-05-01
rotate.sequence

toots:


toots:
@dbaelde. any potential issue with using the same source in two branches of a switch?

PR #349, opened 2016-08-20
Add lib64 location for ladspa plugins

hairmare:
Some modern distros seem to install ladspa plugins into `/usr/lib64/ladspa` (I'm look at you centos). Adding an additional path to the search seems sensible since the paths are hardcoded anyhow.


smimram:
I am fine with that, thanks for the PR! For future reference, note that these paths can also be changed via the `LIQ_LADSPA_PATH` environment variable.

hairmare:
Thank you very much for merging. I was aware of the environment variable but figured the PR would help more people than adding documentation in my project.

PR #352, opened 2016-09-21
Make track_sensitive a bool getter.

toots:


toots:
@dbaelde any argument against or in favor of merging this? I don't see why not..

toots:
Merging this, seems fine to me.

PR #365, opened 2016-11-18
Add peak operator.

### LINKED ISSUE #364: feature request: server.max
I do not know if this is the right way/place to submit a wish.

We are using rms.stereo to log and plot loudness of our streams.
To include peaks into this picture it would be great to have something working in the same way, that can be used with telnet/remote control, but returning just the maximum values for the configured window. this would allow us to have all (rms, peak, range in a single picture)

Best regards, Peter

smimram:
Adapt RMS operator so that it can also output peak, fixes #364.


toots:
Looks good to me!

liquidpeter:
thanks for this!
It works fine.
BR, Peter

`set("server.telnet", true)
set("server.telnet.port",9876)
set("server.telnet.bind_addr","127.0.0.1")

radio=input.http(id="input.http", "http://ice.rosebud-media.de:8000/88vier-low");

duration = interactive.float("duration", 1.)

rs = rms.stereo(duration=duration, radio)
r = fst(rs)
radio = snd(rs)

ps = peak.stereo(duration=duration, radio)
p = fst(ps)
radio = snd(ps)

def measure(_) = "#{fst(r())} #{snd(r())} #{fst(p())} #{snd(p())}" end
server.register("measure", measure)

out(radio)`

smimram:
Great! :)

PR #374, opened 2016-12-23
Update unbuffered alsa example

hairmare:
It looks like the example was a bit outdated. I've updated the documented log output to what liquidsoap was telling me and changed it to use the frame.audio.size setting to get rid of 'there is no configuration key named "frame.size"' warnings.


toots:
Thanks!

PR #376, opened 2017-01-07
SSL support for harbor

toots:


toots:
Looks like it's working! Needs some polish and it's ready to go!

toots:
This is ready to roll!

PR #391, opened 2017-02-04
Account for buffer contents when seeking backwards (#390)

### LINKED ISSUE #390: Beginning of songs are cut too much when using cue_cut
Hi, we are currently renewing our studio automation software and liquidsoap really safes our day. Thanks to everyone contributing to this fine piece of software.

I recently discovered a small problem when using cue_cut. We use silan (https://github.com/x42/silan) to detect of silent parts at the beginning and the end of audio files, and then use this information with the cue_cut operator. When the time of the cue_in point is small (< 0.04s, smaller than the frame length) which quite often _is_ the case, not just a couple of milliseconds, but around half a second is cut off.

It looks like this has happened to others as well: https://forum.sourcefabric.org/discussion/16854/begin-of-songs-are-cut/p1

Tested with:

    $ liquidsoap --version
    Liquidsoap 1.2.0
    Copyright (c) 2003-2016 Savonet team
    Liquidsoap is open-source software, released under GNU General Public License.
    See <http://liquidsoap.fm> for more information.

    $ liquidsoap 'out(cue_cut(single("annotate:liq_cue_in=\"0.001\":test.flac")))'

I think that the fseek function in the file decoder forgets to account for the contents in the pre-buffer when seeking backwards. See https://github.com/savonet/liquidsoap/blob/1.2.1/src/decoder/decoder.ml#L429-L433

smlz:
I _think_ this fixes #390. (I don't really know OCaml)

Tested with the flac and mad decoders.

Seek is still off by a little (-81 samples for mad, +21 samples with flac), but it is not audible any more.

### DIFF
--- a/src/decoder/decoder.ml
+++ b/src/decoder/decoder.ml
@@ -426,17 +426,13 @@ struct
     in
     let fseek len = 
       let gen_len = Generator.length gen in
-      if len < 0 then
-       begin
-        Generator.clear gen;
-        decoder.seek len
-       end
-      else if len > gen_len then 
+      if len < 0 || len > gen_len then
         begin
          Generator.clear gen;
          gen_len + decoder.seek (len-gen_len)
         end
        else
+        (* Seek within the pre-buffered data if possible *)
         begin
          Generator.remove gen len;
          len


toots:
That looks great to me. Merging. Thanks!

PR #407, opened 2017-03-16
Proper parsing of string litterals.

toots:
I think this is fine. I'm not sure what `incrline` is for so I'm leaving this here for review..

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -55,6 +55,8 @@ Bugfixes:
 
 - Fix crash in external (download) protocol.
 
+- Fix parsing of "\\" and '\\' strings (#368)
+
 
 1.2.1 (01-07-2016)
 ========
--- a/scripts/externals.liq
+++ b/scripts/externals.liq
@@ -23,7 +23,7 @@ def enable_external_mpc_decoder() =
       get_channels(file)
     else
       # Otherwise test file extension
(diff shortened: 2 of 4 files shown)


smimram:
Do we really have to double backslashes in liquidsoap files? Apparently we now have to type \\# instead of \#, etc. This is both heavy and non backward-compatible...

toots:
Can you give me an example, @smimram ? I can't reproduce here:
```
# This is a test

x = "this is \
  another test # \ \""

print(x)
```

```
% ./liquidsoap /tmp/test.liq
Warning: illegal backslash escape in string.
this is another test # \ "
```

Simple backlash will work but generate a warning. I've tried to model it after OCaml's string parsing. We use `"%S"` to print strings so it makes sense to also parse them the same way so that we're idempotent here. I'm open to suggestions, though, of course :-)

smimram:
I did not really try but saw lots of changes such as
```
 -    l = string.split(separator='([^=]+\s*=\s*"(\\"|[^"])*")\s*,\s*',s)
 +    l = string.split(separator='([^=]+\\s*=\\s*"(\\"|[^"])*")\\s*,\\s*',s)
```
where `\s` becomes `\\s`...

toots:
This will be fixed when #408 is merged.

PR #408, opened 2017-03-17
Unicode parser/lexer with sedlex/menhir

toots:
Well, this is fun:
```
% cat /tmp/test.liq
# This is a test

%define ע "test"

ש = "This is a #{ע}"

print(ש)

print("That's all, folks!")
% ./liquidsoap /tmp/test.liq
This is a test
That's all, folks!
```


toots:
Well, this is fun:
```
% cat /tmp/test.liq
# This is a test

%define ע "test"

ש = "This is a #{ע}"

print(ש)

print("That's all, folks!")
```

```
% ./liquidsoap /tmp/test.liq
This is a test
That's all, folks!
```

smimram:
This is really cool, but do we want even more dependencies? That being said with opam it should be ok...

toots:
Yeah, with opam I don't think that this is a big deal.

The plus of switching to these guys is fancy unicode support but also better long-term support for lexing/parsing. Also, the pre-processor code is much, much cleaner.

This will need a long observation time, though.

The plan, I think would be to merge this right after the next release and let it simmer for a while..

Issue #446, opened 2017-05-16
icy metadata update example not working on HEAD


mcfiredrill:
Hi I just compiled the latest git HEAD, the example for using `icy.update_metadata` on this page is not working anymore.
http://savonet.sourceforge.net/doc-svn/icy_metadata.html

```
#!/usr/local/bin/liquidsoap

def icy_update(v) =
  # Parse the argument
  l = string.split(separator=",",v)
  def split(l,v) =
    v = string.split(separator="=",v)
    if list.length(v) >= 2 then
      list.append(l,[(list.nth(v,0),list.nth(v,1))])
    else
      l
    end
  end
  meta = list.fold(split,[],l)

  # Update metadata
  icy.update_metadata(mount="/mystream",password="hackme",
                      host="myserver.net",meta)
  "Done !"
end

server.register("update",namespace="metadata",
                 description="Update metadata",
                 usage="update title=foo,album=bar,..",
                 icy_update)

```

The error is
```
At line 18, char 47:
  this value has type
    [(((default:_)->_)*_)] (inferred at line 9, char 20)
  but it should be a subtype of
    [(string*_)]
```

Which I assume means `meta` is the wrong type.

I did a git blame on `src/lang/builtins_cry.ml` where `icy.update_metadata` is defined, and it doesn't look like anything has changed recently.

Did something change about how one of the `list` functions works?

hairmare:
Since 1.3.0 the `list.nth` function needs a `default` argument to be able to infer is return type. This also affects other list functions like `list.hd`. 

Before 1.3.0 those functions always returned  a string type  so they where basically only useable with lists of strings. With 1.3.0 you can help the list functions figure out they should return a string as follows.

```diff
diff --git a/doc/content/icy_metadata.txt b/doc/content/icy_metadata.txt
index bf7d7f53..9230926e 100644
--- a/doc/content/icy_metadata.txt
+++ b/doc/content/icy_metadata.txt
@@ -41,7 +41,7 @@ def icy_update(v) =
   def split(l,v) =
     v = string.split(separator="=",v)
     if list.length(v) >= 2 then
-      list.append(l,[(list.nth(v,0),list.nth(v,1))])
+      list.append(l,[(list.nth(v,0,default=""),list.nth(v,1,default=""))])
     else
       l
     end

```

[...]

PR #447, opened 2017-05-16
update icy_metadata doc with list.nth default argument

mcfiredrill:
see https://github.com/savonet/liquidsoap/issues/446

Before 1.3.0 list.nth didn't require a default argument to indicate they
should return a string, so this example won't work out of the box on
1.3.0.


toots:
Thanks!

PR #460, opened 2017-06-23
Support replaygain aside mp3gain

d4h3r0:
Since mp3gain is not available in Debian 8 / Jessie I have tweaked the source to have support for replay gain. Command line modification works, haven't tested on my stream yet.


toots:
Looks great, thanks!

PR #552, opened 2018-05-12
Adapt to new ocaml-ffmpeg API

gndl:
Maybe this adaptation of Liquidsoap to the new API of Ffmpeg may interest you.
Note that on my repository, the CI fails with the following error :
**fatal: reference is not a tree: 53a6ce816b36a811d7fb5d6c3a98375896732403**.
I do not understand the cause of this error.


toots:
Fixed in 3272e1a9!

Issue #544, opened 2018-04-23
Allow printing to a server connection


davehayes:
I was unable to find any print command that prints arbitrary text to a unix domain socket server connection. Thus I am unable to use the ```on_end``` callback to print a message to the server when the song is about to change. This forces my other software to poll the server repeatedly rather than wait for an IO event. 

Is it possible to add a command that will print text to a server connection?

toots:
Hi,

Would you mind describing the functionality a bit more? Are you talking about an existing server connection or a connection you would also open and close?

davehayes:
So here's the basicis of the script I'm using:
```
set("log.file.path","liquidstream.log")
set("log.stdout", true)
set("server.socket", true)
set("server.socket.path", "/tmp/liquidserver")
set("server.socket.permissions",0o660);

defaultsource = blank(duration=15.0)
serverqueue = request.queue(id="mainreq")
serverqueue= amplify(1., override="REPLAYGAIN_TRACK_GAIN", serverqueue)
content = fallback([ serverqueue, defaultsource ])
crossplay = smart_crossfade(content)
output.icecast(
  %vorbis(channels=2, quality=0.3),
  ...
)
```

So I have an external program that opens the unix domain socket that this script created. Right now, in this external program, I push commands to it and read responses using a simplistic model (no select multiplexing yet, etc.). At the moment, due to this simple model, I poll liquidsoap through this socket to see how big the song queue is and how long the current song has left to play. 

What I'd like to accomplish is to move to an event driven model. To do that, I have to somehow get the `on_end` callback to send a message to my external program everytime the currently playing song ends. To do that, I have to have a way to print arbitrary text to the unix domain socket from within that callback. Then I can move away from polling liquidsoap and become purely event driven. 

Does this help?

toots:
Done now! See: http://www.liquidsoap.info/doc-dev/server.html

davehayes:
Awesome! I like the condition variable stuff too. Thank you!

PR #568, opened 2018-06-27
[WIP] Interactive server commands

toots:
Tentative implementation for #544. Still needs proper documentation and such.

Test script:
```
set("log.file",false)
set("log.stdout",true)
set("server.telnet",true)
set("server.timeout",-1.)

x = server.wait()
wait = fst(x)
resume = snd(x)

value = ref ""

def wait(_) =
  def rec fn () =
    def after () =
      value = !value
      if value == "exit" then
        "All done!"
      else
        server.partial_write(after=fn,"#{value}\n")
      end
    end
    wait(after)
  end
  fn ()
end

def send(v) =
  value := v;
  resume()
  "Ok!"
end

server.register(usage="wait",description="Wait for some data","wait",wait)
server.register(usage="send <data>",description="send some data. Send \"exit\" to end waiting clients","send",send)

output.dummy(blank())
```

Usage: 
First client:
```
% telnet localhost 1234
Trying ::1...
telnet: connect to address ::1: Connection refused
Trying 127.0.0.1...
Connected to localhost.
Escape character is '^]'.
wait
foo
bar
exit
All done!
END
```

Second client:
```
% telnet localhost 1234
Trying ::1...
telnet: connect to address ::1: Connection refused
Trying 127.0.0.1...
Connected to localhost.
Escape character is '^]'.
send foo
Ok!
END
send bar
Ok!
END
send exit
Ok!
END

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -225,6 +225,7 @@ $(call conditional_compilation,synth,$(W_DSSI),synth/dssi_op.ml)
 builtins = lang/lang_builtins.ml lang/builtins_json.ml \
            lang/builtins_harbor.ml lang/builtins_dtools.ml \
            lang/builtins_time.ml lang/builtins_callbacks.ml \
+					 lang/builtins_server.ml \
            $(if $(W_SSL),lang/builtins_https.ml lang/builtins_harbor_ssl.ml) \
            $(if $(W_OSX_SECURE_TRANSPORT),lang/builtins_https_secure_transport.ml lang/builtins_harbor_secure_transport.ml)
 
--- a/src/lang/builtins_server.ml
+++ b/src/lang/builtins_server.ml
@@ -1,0 +1,103 @@
+(* -*- mode: tuareg; -*- *)
+(*****************************************************************************
+
+  Liquidsoap, a programmable audio stream generator.
+  Copyright 2003-2018 Savonet team
+
+  This program is free software; you can redistribute it and/or modify
+  it under the terms of the GNU General Public License as published by
+  the Free Software Foundation; either version 2 of the License, or
+  (at your option) any later version.
+
+  This program is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details, fully stated in the COPYING
+  file at the root of the liquidsoap distribution.
+
+  You should have received a copy of the GNU General Public License
+  along with this program; if not, write to the Free Software
+  Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
+
+*****************************************************************************)
+
+open Lang_builtins
+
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

**Comment by toots** [comment]
@davehayes, does the example script make sense to you? Could it work to implement your functionality?

**Comment by davehayes** [comment]
There's a lot of this that is opaque to me.

First of all, just as an FYI you have a redirect issue with `liquidsoap.info`. The url `www.liquidsoap.info` works, but the url  `liquidsoap.info` redirects infinitely. 

Next, is this scripting language supposed to be a derivative of OCaml? I would really like to see some BNF or more formal documentation on how it works.  

Also, my confusions:

- I assume `ref` is taking the reference of a variable? I'm also assuming `#{value}` dereferences a reference. If that's true, what does `value = ! value` do?
- You define `wait` as a variable, and then as a function. Wait is not defined as a recursive function, yet you call this inside of the `wait` function definition. So ... which wait is which?  
- You used a telnet server, which might be slightly different from a unix domain socket server. Do they both work exactly the same way? 

Finally, would you say something like this would work to print a message 5 seconds before a song ends to any connected clients?

```
def dontKnowWhatToDoWithThis 
    "blargh"
end

def song_ends(remainingTime, metaData) 
    server.partial_write(after=dontKnowWhatToDoWithThis, "Song about to end: $(remainingTime) $(metaData)\n")
end

on_end(sourceID, 5.0, song_ends, whatDoesThisArgDo)
```

**Comment by toots** [comment]
Yeah, my example is more complex than you need to. No, liquidsoap isn't a derivative of OCaml but it borrows heavily from it.

Let's try to sketch out the kind of scripting you'd need:
```
# Create a pair (wait,resume)
x = server.wait()
wait = fst(x)
resume = snd(x)

# Create a reference to the string to be broadcasted:
to_send = ref ""

# Create a function that updates the string to send and wakes up waiting clients:
def wake_up_and_send(data) =
  to_send := data
  resume()
end

# Client server command
def on_end_cmd (_) =
  # We need a recursive handler here
  def rec handler () =
    # After waiting for signal, we print data and wait again
    def after () =
       # Dereference to_send:
       to_send = !to_send

       # End if special value "exit" is given
       if value == "exit" then
        "All done!"
       else
         # Otherwise print to_send and return line:
         server.partial_write(after=handler,"#{to_send}\n")
      end
    end

    # Now wait for signal
    wait(after)
  end
  
   # Let's go!
   handler ()
end

# Register command:
server.register(namespace="songs",usage="on_end",description="Bw notified when a song ends","on_end",on_end_cmd)

# Later in your script:
def on_end () =
  wake_up_and_send("Song ended!")
end

s = operator(on_end=on_end,...)
```

PS: redirect fixed ;-)

toots:
I might try to write a higher level wrapper for this and add it to the standard library, just need a minute to think it out..

toots:
Ok, I've updated the syntax, it should be easier this way now.

davehayes:
Do you have an example of the better syntax? :D

toots:
Yeah, it's been updated above ;-)

PR #591, opened 2018-08-05
Remove useless data copy in ffmpeg_decoder

gndl:
Ocaml-ffmpeg no longer reuses output buffers by default. So it is not necessary to copy the output data anymore.


toots:
Thanks!

PR #626, opened 2018-09-16
Set up CI with Azure Pipelines

toots:


toots:
Don't think we'll do this. Closing

Issue #668, opened 2018-12-02
Change of semantics for the add() operator


toots:
I was working on the scripted fade operators the other day and ran into an issue with add using this code:
```
add(normalize=false,[s,blank(duration=5.)])
```
My intuitive idea was to add a blank source to `s` in order to make sure that the resulting source would run for at least 5s even if `s` ended earlier.

However, this didn't work as expected. `add` actually pumps as much data as it can and removes all track marks.

I also experienced a freeze in the streaming thread while experimenting with it, mostly likely because `add` does not currently ticks its child's sources while pumping more than once.

I suggest that we should change and simplify `add` behavior the following way:
- `add` gets data from each source only once per clock cycle. The furthest break is kept. If that is a partial fill (track mark) then let it be.
- `add` keeps metadata from all sources. Let the user decide if he/she wants to silence a source by dropping all metadata on it.

toots:
I think this is too late for the coming release.

smimram:
Is it improved by #1012?

toots:
Ok, just getting back to this.

First problem: `blank` is marked as `is_ready` even past its duration.. Fixed now..

toots:
I merged the other branch. Only thing that wasn't implemented is the changes in metadata. It's documented that only the metadata from the first source are passed and it is not a bug so I left it this way.

PR #671, opened 2018-12-05
Simplify add.

toots:
This is the implentation for #668.


toots:
I'm merging this. At the very least, `add` was never meant to control its underlying source's time, which is the current implementation and is wrong. Let's keep an eye on it to see if this changes causes issue.

PR #702, opened 2019-01-29
 Use aacgain to extract replay gain from mp3 files

mdurys:
`aacgain` is a drop in replacement for `mp3gain`. `aacgain` is easily available in recent Linux distributions, which is not necessarily the case with `mp3gain`.


toots:
Thanks!

PR #780, opened 2019-05-18
Convert documentation do markdown.

smimram:
The whole documentation has been converted to markdown.


smimram:
@toots can you check that this does not mess up too much the current website (I don't really know how to build it). Otherwise, we can merge I think!

Issue #781, opened 2019-05-18
`peak` and `rms` units


Yamakaky:
What are the units of `peak()` and `rms()`? From testing I assume it's the amplitude between 0 and 1, but I sometime get >1 values for peak (max I saw was 1.003).
I think I will take some time to improve the doc XD

smimram:
Samples are (usually) between -1 and 1 as usual in floating point representation. `peak` is the maximal absolute value and `rms` is [what it is](https://en.wikipedia.org/wiki/Root_mean_square). This is satisfied for input files, but some operators (e.g. a compressor, replaygain, etc.) might produce value greater than 1, and Liquidsoap does nothing to enforce this (unless you use the `clip` operator) but most outputs correctly handle this by clipping anyway.

PR #783, opened 2019-05-19
Improve doc for `rms()` and `peak()`

Yamakaky:
Fix https://github.com/savonet/liquidsoap/issues/781


toots:
Thanks! @smimram how do you want to coordinate this with your current work? Should we merge this directly in the markdown formatted doc?

Yamakaky:
Oh, OK, I can resubmit it after the migration!

toots:
@Yamakaky or you can submit directly there. I believe we won't change the doc here while this effort is underway. Thanks very much for the contribution tho!

smimram:
Yes, it great to have contributions :)

PR #787, opened 2019-05-20
Pretty-print getters.

### LINKED ISSUE #786: Improve printing of getter types
Right now the type of `gstreamer.hls` is
```
(?id:string,?on_error:((string)->float),'a)->source(audio=2,video=1,midi=0)
where 'a is either string or ()->string
```
It would be nice to hide the technical details of getters and display something like:
```
(?id:string,?on_error:((string)->float),{string})->source(audio=2,video=1,midi=0)
```
which would be much more readable, i.e. write `{t}` for a getter of type `t`.

What do you think?

smimram:
A getter on t is printed as {t}. Fixes #786.

### DIFF
--- a/src/lang/lang_types.ml
+++ b/src/lang/lang_types.ml
@@ -22,6 +22,9 @@
 
 let debug = Utils.getenv_opt "LIQUIDSOAP_DEBUG_LANG" <> None
 
+(** Pretty-print getters as {t}. *)
+let pretty_getters = ref true
+
 (* Type information comes attached to the AST from the parsing,
  * with appropriate sharing of the type variables. Then the type inference
  * performs in-place unification.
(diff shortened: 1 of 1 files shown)


dbaelde:
With this modification, you print EVar and UVars in the same way, which might be confusing.

Also, if I understand correctly, the second half of the commit (for printing constraints next to the type) is now dead code.

dbaelde:
Currently the following two types are both printed as `({float})->{float}`:
```
1. ('a)->'b where 'a and 'b are either float or ()->float
2. ('a)->'a where 'a and 'b are either float or ()->float
```
Also, the next two are both printed as '()->{float}':
```
3. ()->'a where 'a is float or ()->float
4. ()->?a where ?a is float or ()->float
```
What do we want for these examples:
```
1. ({float})->{float}
2. ('a)->'a where 'a is {float}
3. ()->{float}
4. ()->?{float}
```
I'm not completely comfortable with the notation {float} which does not insist on the disjunction, but this may be the price to pay to make things look simple for simple uses.

PR #790, opened 2019-05-21
Better error reporting and formatting.

toots:


smimram:
What does this change exactly? It only adds colors? (this is cool!)

toots:
Also uniformed error formatting and cleanup code to have a single place with all errors. Before that some warning were popped in from other places.

I've also cleaned up some error messages such as:
```
      | Term.Ignored tm when Term.is_fun (T.deref tm.Term.t) ->
          flush_all () ;
          Format.printf
            "@[<2>%s:@ This term@ would evaluate to@ a function@ \
               which@ would then be dropped.@ \
               This is usually@ the sign@ of@ an unintended@ \
               partial application:@ some arguments@ may \
               be@ missing.@]@."
            (T.print_pos (Utils.get_some tm.Term.t.T.pos)) ;
          raise Error
```
Becomes:
```
      | Term.No_label (f,lbl,first,x) ->
          let pos_f =
            T.print_pos (Utils.get_some f.Term.t.T.pos)
          in
          let pos_x = T.print_pos (Utils.get_some x.Term.t.T.pos) in
          flush_all () ;
          error_header pos_x ;  
          Format.printf "Cannot apply that parameter because the function %s@ has %s@ %s!@]@."
              pos_f
              (if first then "no" else "no more")
              (if lbl="" then "unlabeled argument" else
                 Format.sprintf "argument labeled %S" lbl) ;
          raise Error
```
The reasoning is that error messages should be short and concise and then the (new) doc should be the point of reference to understand them.

toots:
Also downgraded some errors into warnings, as exemplified above..

toots:
@smimram Let me know if you think we can merge/any feedback.

smimram:
Ah I did not see that (the diff makes it difficult to compare the two).

I kind of liked the old behavior: most of the time, the users
- will not read the warnings
- will not read the doc
so that I think that it made lot of sense to have such things handled as errors, with explicit messages...

toots:
Warnings are what they are meant to be, systematically failing on warnings makes them errors ;-)

As for users not reading the doc, this is what we have to work on! The best then would be to add a link to the relevant doc url or, better since URL are bound to be outdated, add a "for more information, please refer to the documentation for details explanations about each error"

PR #791, opened 2019-05-22
Update ocaml compiler version error string

tatokis:
The error string in configure.ac for unsupported versions of the ocaml compiler wasn't updated, resulting in confusing messages like this:

```
checking for ocamlc version... 4.05.0
configure: error: version 4.03.0 or more of the OCaml compiler is required to build liquidsoap
```

even though
https://github.com/savonet/liquidsoap/blob/31d9d1b2fe6e925a0e3fcc9fb09627b31f27355c/configure.ac#L13


smimram:
Thanks!

PR #793, opened 2019-05-23
Stop process on metadata only when metadata is there.

### LINKED ISSUE #789: external encoder: ffmpeg with video=true
I can't for the life of me get the input format correct for reading with ffmpeg.
The following example in the documentation does not work for me.
```output.file(
  %external(process="ffmpeg -i pipe:0 -f avi pipe:1",video=true),
  "/tmp/test.avi", s)
```

Can I get a hint about how I should read in with ffmpeg? i have tried various combinations of input formats, avi demuxer, rawvideo, pcm_16le...

Also, is it possible to `output.external`  with video, similarly to ho the external decoder works (just no pipe back in to liquidsoap) ?
@smimram i've noticed you've been responsible for a lot of the video output stuff. Any hints?

smimram:
Fixes #789.

### DIFF
--- a/src/encoder/external_encoder.ml
+++ b/src/encoder/external_encoder.ml
@@ -99,10 +99,15 @@ let encoder id ext =
                         ~on_stderr ~log ext.process 
   in
 
-  let insert_metadata = Tutils.mutexify mutex (fun _ -> 
-    if ext.restart = Metadata then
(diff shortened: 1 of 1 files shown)


smimram:
@toots Do you agree with that? I would stop on metadata, only when `restart_on_metadata` is specified.

PR #796, opened 2019-05-25
Designation for RMS window size

Yamakaky:
`normalize` uses `window` while `compress` uses `rms_window`. I think the second one is better and should be used everywhere.


smimram:
I would rather put `window` in `compress` if that's fine with you. I don't see how this could bring ambiguity and I rather like simple argument names...

What do you think?

Yamakaky:
As long as they use the same same, good enough ^^

PR #805, opened 2019-05-30
Error colors

toots:
Let's review error colors. 

Here's @smimram favorite:
Screenshot 2019-05-30 11 15 07

I'd rather have error line in bold white:
Screenshot 2019-05-30 11 14 35

(indentation difference is because the `Format` pretty printer is not aware of terminal control codes. Suggestions welcome)

Finally, warning are yellow for now:
Screenshot 2019-05-30 11 15 39


toots:
Indeed. How does it look with OCaml errors? I replicated their color scheme there..

smimram:
Seems to be bold black here:
ocaml

toots:
Ok, found the issue. It's just bold for code location. Also switched warnings to magenta just like OCaml. Looking good to merge now!

smimram:
Looks great now :)

PR #815, opened 2019-06-03
Space up types display.

smimram:


smimram:
For instance, the type of `output.file.hls` is changed from
```
(?id:string,?fallible:bool,?on_start:(()->unit),
 ?on_stop:(()->unit),?perm:int,?playlist:string,
 ?segment_duration:float,?segments:int,?start:bool,
 string,[(string*format('a))],source('a))->
active_source('a)
```
to
```
(?id : string, ?fallible : bool,
 ?on_start : (() -> unit), ?on_stop : (() -> unit),
 ?perm : int, ?playlist : string,
 ?segment_duration : float, ?segments : int,
 ?start : bool, string, [string * format('a)],
 source('a)) -> active_source('a)
```

Isn't it much more readable?

toots:
I think it looks good. These days, I'm deferring to OCaml a lot, I'll look at what they do tomorrow.. 😉

toots:
I think it's fine 😉

Issue #804, opened 2019-05-29
Fade Skip


jdbuys:
Would it be possible to create a function that when it is called will fade out the song for x amount of seconds and then when it is done it calls skip?

It just makes the transition between track sound better than just an abrupt stop when you call skip.

Thanks.

Yamakaky:
I could use that ^^

smimram:
Fading is already almost entirely coded in liq: https://github.com/savonet/liquidsoap/blob/master/lib/fades.liq

It should not be too hard to adapt that in order to achieve what you want. I'll give it a show some day soon if you don't do it before me.

PR #821, opened 2019-06-06
Fade song and skip when metadata is set

jdbuys:
Fixes: #804

### DIFF
--- a/lib/fades.liq
+++ b/lib/fades.liq
@@ -148,8 +148,7 @@ def fade.skip(~id="fade.skip",~duration=5.,
       remaining = source.remaining(s)
       duration = if remaining < !duration then remaining else !duration end  
       log("Skip fade executed for: #{duration}s")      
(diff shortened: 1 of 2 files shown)


lib/fades.liq
```diff
@@ -112,6 +112,63 @@ def fade.out(~id="fade.out",~duration=3.,
...
+    -1.
+  end
+
+  def update_fade(m) = 
+    if m[override_skip] == "true" then
+      remaining = source.remaining(s)
+      duration = if remaining < !duration then remaining else !duration end  
+      log("Skip fade executed for: #{duration}s")      
+      fn := mkfade(start=1.,stop=0.,type=!type,duration=duration,s)
+      ignore(list.add((override_skip,""),m))
```
toots: Is that needed? Doesn't seem like it will do anything?

jdbuys: It is not. I removed it.

toots:
Looks great, thanks!

PR #823, opened 2019-06-06
Remove the warning log level.

smimram:
The terminology "warning" looks like a warning is more important than an information whereas it is not. Moreover, we never needed more than 5 levels.


toots:
Makes sense to me. I've never used level `6` either.

PR #824, opened 2019-06-07
Asynchronous reading of AVI headers.

smimram:


src/sources/external_input.ml
```diff
@@ -156,6 +156,9 @@ object (self)
 
   method stype = Source.Fallible
 
+  (* Whether we should read headers. *)
+  val mutable has_headers = true
```
toots: `need_headers`?

toots: You also need a mutex to protect changing/reading its value since it's gonna be accessed from different threads:
```
val mutable need_headers = true
val need_headers_m

method get_need_headers =
  Tutils.mutexify (fun () -> need_headers) ()

method set_need_headers =
  Tutils.mutexify (fun v -> need_headers <- v)
```

toots: Might be able to find better methods names but you see the idea..

src/sources/external_input.ml
```diff
@@ -202,24 +207,31 @@ object (self)
...
-                           that you are producing the correct kind of data?" d v a
-                      )
-                  with
-                  | End_of_file -> raise (Finished ("Process exited.", restart))
-                end;
+                (* Do we need to read the headers? *)
+                if has_headers then
+                  (
+                    has_headers <- false;
+                    read_headers in_d
```
toots: How big are `AVI` headers on the average? The executing task is set as `Non_blocking` and waiting on a `read` signal. However, the headers read could potentially be blocking. I might chime in an adapt `Avi.Read.headers_simple` to work as a non-blocking duppy task as well.

smimram: Headers are very "small" I would say (= a couple of metadata).

The thing is that
1. I precisely wanted to avoid duppyfying all the `Avi` module
2. pulling frames is also potentially blocking so it's not worse than before.

src/sources/external_input.ml
smimram: From the point of view of the source `has_header` indicates whether the stream has headers.

smimram: I don't really see when there are multiple threads running here, there's only one duppy thread no?

toots: Nope, duppy has a model with multiple threads picking up pending tasks.

smimram:
@toots are you ok now?

toots:
Almost there! Change priority to `Maybe_blocking` and we're good to go!

src/sources/external_input.ml
toots: Gotcha.

PR #825, opened 2019-06-07
Fix invalid syntax in depexts

Yamakaky:
Error message:

[WARNING] Failed checks on liquidsoap package definition from source at git+https://github.com/savonet/liquidsoap.git#master:
           warning 54: External dependencies should not contain spaces nor empty string: "automake autoconf", "automake autoconf", "automake autoco
nf", "automake autoconf", "automake autoconf", "automake autoconf"


smimram:
Well spotted, thanks.

PR #826, opened 2019-06-07
Improve merge_tracks doc

Yamakaky:


src/operators/sequence.ml
```diff
@@ -152,7 +152,9 @@ let () =
   Lang.add_operator "merge_tracks"
     [ "", Lang.source_t k, None, None ]
     ~category:Lang.TrackProcessing
-    ~descr:"Merge consecutive tracks from the input source."
+    ~descr:"Merge consecutive tracks from the input source. They will be \
+            considered as one big track, so on_track() will not trigger \
```
smimram: Great, thanks!

smimram: One thing, please use backquotes for code such as
```suggestion
            considered as one big track, so `on_track()` will not trigger \
```

Yamakaky: Right, forgot this syntax ^^

PR #846, opened 2019-06-22
Add reopen telnet command to output.external.

smimram:


toots:
That looks good to me with only one question about possible concurrent calls between telnet and the callback. An access mutex would be nice to add.

toots:
You forgot my mutex!

smimram:
Done in 86329b9dd273.

toots:
Thanks!

Issue #419, opened 2017-03-29
Type checking inconsistency with ref and source/active_source


toots:
```
# l = ref [];;
l : ref([?A]) = ref([])
# list.iter(source.shutdown,!l);;
- : unit = ()
# l;;
- : ref([source(?A)]) = ref([])
# l := [output.dummy(blank())];;
At line 1, char 3-4:
  this value has type
    ref([source(_)]) (inferred at line 1, char 5-10)
  but it should be a subtype of (the type of the value at line 1, char 3-28)
    ref([active_source(_)])
```
```
# l = ref [];;
l : ref([?A]) = ref([])
# l' = !l;;
l' : [?A] = []
# l := [output.dummy(blank())];;
- : unit = ()
# l;;
- : ref([active_source(audio=2,video=0,midi=0)]) = ref([<source>])
# l';;
- : [active_source(audio=2,video=0,midi=0)] = []
# list.iter(source.shutdown,l');;
- : unit = ()
```

toots:
Is that really a bug not that I think about it? `active_source` is a sub-type of `source`. Should the final check on the first example restrict the type for the list from `source list` to `active_source list`?

dbaelde:
I don't see any bug here. It's just that type inference isn't always the smartest wrt subtyping, which is to some extend unavoidable.

In the first example, there should be two ways to fix the type inference failure: either a type annotation in one of the first two lines to tell the type system that we want a list of active sources, or a type annotation/coercion in the last line to help the type checker see that the active source list is also a source list and can be stored in the reference. I don't have a working liquidsoap at hand to test that, though.

toots:
I'm taking this one off the bucket list for `1.4.0`, will tackle in during the next major release cycle.

smimram:
I guess this is solved by #1012 .

toots:
Good question. Let me think about it.

PR #847, opened 2019-06-24
Make source and active_source the same for the type checker.

toots:
This one seems like a straight shot to me because of this comment:
```
 * Memo: A <: B means that any value of type A can be passed where a value
 * of type B can.
```
However, I'm no expert on this part so it'd be nice to have @dbaelde's eyes on it.

### DIFF
--- a/src/lang/lang_parser.mly
+++ b/src/lang/lang_parser.mly
@@ -131,7 +131,6 @@
       | "string" -> Lang_types.make (Lang_types.Ground Lang_types.String)
       | "source" ->
           (* TODO less confusion in hiding the stream_kind constructed type *)
-          (* TODO print position in error message *)
           let audio,video,midi =
             match args with
               | ["",a;"",v;"",m] -> a,v,m
--- a/src/lang/lang_parser.mly
+++ b/src/lang/lang_parser.mly
@@ -129,7 +129,7 @@
       | "int" -> Lang_types.make (Lang_types.Ground Lang_types.Int)
       | "float" -> Lang_types.make (Lang_types.Ground Lang_types.Float)
       | "string" -> Lang_types.make (Lang_types.Ground Lang_types.String)
-      | "source" | "active_source" ->
+      | "source" ->
           (* TODO less confusion in hiding the stream_kind constructed type *)
           (* TODO print position in error message *)
           let audio,video,midi =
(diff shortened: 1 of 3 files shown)


toots:
Also, this is a fix for #419

dbaelde:
The point of distinguishing sources and active sources as types is to leverage the type system to detect some errors. Concretely, this is (only?) used to warn users when they silently ignore sources (e.g. in `blank() ; blah`) without issuing false warnings on legitimate code (e.g. `output.dummy(blank()) ; blah`). The price to pay for this feature is slightly more verbose types and, sometimes, type inference failures because type inference isn't always smart enough to use the subtyping relation `active_source <: source`.

I cannot evaluate the cost/benefit balance of that feature, since I did not follow user reports recently: I leave this to you.

Regarding the commit, it should be improved. If the goal is to get rid of the distinction between active and passive sources, the `active_source` type should be completely removed. The current change of adding a subtyping relation is semantically unsound (because not all sources are active sources) and it will make the type system confusing: users will still get warnings about unused passive sources, but not if a type annotation is used to turn these passive sources into (incorrectly) active ones.

toots:
Thanks for your comment. If I'm correct, it seems that the confusion here comes from the fact that the distinction between `active_source` vs. `source` is a _runtime_ feature. Meanwhile, in the code, `active_source` and `source` can be swapped anywhere.

The only other distinction, indeed, is a top-level, where warnings will be issued for dangling `source`s but not `active_source`s. 

It seems to me that the change in this PR would juste make the `source` type equivalent to the `active_source` type or, to be more formal:
```
source(audio=<a>,video=<b>,midi=<c>) <: active_source(audio=<a>,video=<b>,midi=<c>)
active_source(audio=<a>,video=<b>,midi=<c>) <: source(audio=<a>,video=<b>,midi=<c>)
```
Considering that this will only impact warnings for top-level dangling sources, I'm not sure where this could trigger any inconsistency that the user should be aware of.

The only issue, as you said could be by using type annotation, which we could simply disable for `active_sources`. Actually, I'll push that change next.

Finally, I think it's nice to still keep the distinction to let users know which sources are actively pulled.

For all these reasons, I'm sorry to say but I still don't see where this patch would cause troubles. Perhaps you could show a concrete example?

Thanks for reviewing!

dbaelde:
There is a runtime distinction between active and passive sources but also a static distinction : types are static, and they are used for the warning that I mentioned (through `Lang_values.can_ignore`). Also, note that the warning is not only for toplevel stuff, but more generally for expressions in sequence, e.g. possibly inside functions.

Again, you can remove all static notion of active source, but then it seems odd to erase the distinction between active and passive sources using two type inclusions: instead, you should simply remove `active_source` from the type system!

toots:
Ok. I think long term we could simply rename `active_source` as `output` and make it an entirely different type. As far as I know, `active_source` are mostly outputs and it would seem much cleaner to make sure that outputs are never re-used as sources.

PR #855, opened 2019-07-02
Add support for bash completion.

smimram:


doc/content/install.md
```diff
@@ -21,7 +21,7 @@ how to use OPAM. In order to use it:
 - [you should have at least OPAM version 2.0](https://opam.ocaml.org/doc/Install.html),
 - you should have at least OCaml version 3.08.0, which can be achieved by typing
   ```
-  opam switch install 3.08.0
+  opam switch create 3.08.0
```
toots: `4.08.0`

toots: Above as well.

configure.ac
```diff
@@ -127,6 +127,9 @@ if test -z `which pkg-config 2>/dev/null` ; then
   AC_MSG_ERROR(Cannot find pkg-config)
 fi
 
+bashcompletiondir='${prefix}'`pkg-config --variable=completionsdir bash-completion`
+AC_SUBST(bashcompletiondir)
```
toots: What's the portability of this? How does it behave when there is no `bash-completion`? Here, I have:
```
% pkg-config --variable=completionsdir bash-completion
Package bash-completion was not found in the pkg-config search path.
Perhaps you should add the directory containing `bash-completion.pc'
to the PKG_CONFIG_PATH environment variable
No package 'bash-completion' found
zsh: exit 1     pkg-config --variable=completionsdir bash-completion
```

doc/content/install.md
smimram: Just corrected in master.

toots:
This looks good to me now.

configure.ac
smimram: The new version should behave correctly when the package is not found (a default path is given).

PR #860, opened 2019-07-04
Use preprocessor to handle uminus.

smimram:
This avoids the whole grammar duplication business....


smimram:
I like my idea, since it simplifies much the code, but it can wait for 1.5.0 (although I'd be glad to have reviews before that).

Issue #732, opened 2019-03-10
Audio just stops, logging output stops, process still running


rf89:
**Describe the bug**
Hi, We're using liquidsoap with Centova Cast. It was installed with their auto installer. Liquid soap forms part of their 'auto dj' system. I keep finding streams where liquid soap process is running, but there is no audio, no source connected and logging has simply stopped abruptly i.e there is no obvious error, and no pattern to where the logging is stopping. It can happen once or twice w day, or sometimes it may be days or weeks between restarting the process and it stopping again.

**To Reproduce**
I'm not sure whats causing it, but it is only affecting a small number of stations, its unclear if there is anything common between the affected streams. All but 1 is SC2, one other was Icecast2. The issue is resolved temporarily by restarting liquidsoap.

**Expected behavior**
No sudden stoppage of liquid soap.

**Version details**
 - OS; CentOS 7
 - Version 1.3.3

**Install method**
Centova Cast. I'm not sure how they build. Its not OPAM though - they say they are unable to do this due to licence issues?

Centova have said a few users have reported the issue to them and the fix is to rebuild using OPAM. They said they can't deploy the update via OPAM due to a licence issue and seem unwilling to help.

The main reason I've not attempted to install via OPAM my self is because I'm unsure if a 'standard' install of liquid soap will work or what extra packages/options need to be specified during install, and they aren't being so responsive at helping.

Is this an issue you are aware of? Is it resolved in 1.3.4? Any other suggestions?

Thanks

autonarcosis:
Install via Opam. I can't remember the last version that worked by compiling it outside of Opam.  Opam will tell you what modules is needed for it. They are all there within Opam. You'll also need to using Opam v4.06 unless LS has been updated for v4.07. This won't affect the system installed LS.  The executable will be in something like /home/user/.opam/4.06/bin/

toots:
Hi,

I remember helping another user with CentovaCast and it was installing via `opam`. We've had reports of similar issues in the past but never were able to track this bug down, I believe. 

I'd say the first order of action would be indeed to try to see if the bug happens with recent versions. `1.3.6` is the latest. However, there aren't much differences between the two so I wouldn't stress it too much.

If, as you describe, the process is still running then capturing a stacktrace could prove very useful. This can be achieved with `gdb`. Rough sequence of commands (from memory):
```
% gdb -p <process PID>
> thread apply all bt
```

dhannyz:
@toots ive just had this happen (on current master)
[...]

PR #871, opened 2019-07-07
Remove logging in finalise callback.

toots:
Fixes: #732


src/source.ml
```diff
@@ -139,6 +139,8 @@ and 'a link_t =
   | Unknown of 'a list * 'a var list (** the clock variable is unknown but depends on other variables *)
   | Same_as of 'a var (** the clock variable is subtituted by another *)
 
+let debug = Utils.getenv_opt "LIQUIDSOAP_DEBUG" <> None
```
smimram: Maybe also a parameter within Liquidsoap?

Issue #866, opened 2019-07-05
Allow more option on output.file.hls


mbugeia:
**Is your feature request related to a problem? Please describe.**
The output.file.hls lacks some parameter the could be useful in production environments.

**Describe the solution you'd like**
I would like the following parameter to be added in output.file.hls:
segments_list_size (of type int, which defaults to 10): Number of segments to keep in playlist.
segments_on_disk (of type int, which defaults to 15): Number of segments to keep on disk
segment_name (of type string, which defaults to $formatname_%06.ts): Pattern for the segment name.

I would like to construct segment whose name would be something like: myradio_$timestamp_$segment_duration.ts
Where $timestamp would be an unix timestamp. I like the ffmpeg approach on it https://ffmpeg.org/ffmpeg-all.html#Options-41 it allow pattern like %03 (for 000,001,002,...) or strftime format like %Y%m%d-%s

The following parameter would be remove (replaced by segments_list_size in my example):
segments (of type int, which defaults to 10): Number of segments to keep.

**Describe alternatives you've considered**
None, but I'm not fixed on parameter name ;)

PR #872, opened 2019-07-07
HLS improvements

toots:
Fixes: #866


src/outputs/hls_output.ml
```diff
@@ -36,10 +53,21 @@ let hls_proto kind =
...
 
+     "segment_name",
+     segment_name_t,
+     Some default_name,
+     Some "Segment name. \
+           Default: `fun (~duration:_,~position,~extname,stream_name) -> \"#{stream_name}_#{position}.#{extname}\"`";
+
      "segments",
      Lang.int_t,
+     Some (Lang.int 15),
```
smimram: I would do everything in terms of duration rather than segments.

toots: What's your rationale for that?

mbugeia: While using duration could make sense, I think it's less user friendly because it need to be a multiple of the segment duration.

smimram: My idea was that we want to specify how much time back we want to keep rather than a number of segments (for instance if you double the duration of segments you are likely to want to halve the number of segments to keep). But this is rather minor.

mbugeia:
LGTM :)

PR #886, opened 2019-07-22
Add on_frame.

smimram:


toots:
Let's hold on to that for a second if you don't mind, I'm not sure that the proposed use for it is warranted.

smimram:
Yes, no rush, it was only a proposal.

toots:
Closing this as we realized executing script stuff on a pre-frame basis is a very bad idea!

smimram:
With recent improvements in the language I'd like to reopen this one.

smimram:
Let's merge this, we'll see what can be done with it.

PR #897, opened 2019-07-24
Add buffer_length telnet command to input.harbor.

smimram:


src/sources/harbor_input.ml
```diff
@@ -44,12 +44,22 @@ struct
...
-              ~empty_on_abort:false ~replay_meta ~bufferize
+    inherit  Source.source
+      ~name:Harbor.source_name kind as super
+    inherit Generated.source abg
+      ~empty_on_abort:false ~replay_meta ~bufferize
+
+    initializer
+      self#register_command "buffer_length"
+        ~descr:"Show internal buffer length (in seconds)."
+        (fun _ -> Printf.sprintf "%.02f" (Frame.seconds_of_master (Generator.length abg)))
```
toots: Technically, this would need to be mutex-protected but I think the risk is low.

PR #898, opened 2019-07-26
SRT I/O

toots:


src/io/srt_io.ml
```diff
@@ -0,0 +1,439 @@
...
+  Lang.add_operator "input.srt"
+    ~kind:(Lang.Unconstrained kind)
+    ~category:Lang.Input
+    ~descr:"Receive a SRT stream."
+    [ "bind_address", Lang.string_t, Some (Lang.string "0.0.0.0"),
+      Some "Address to bind on the local machine.";
+
+      "port", Lang.int_t, Some (Lang.int 8000),
+      Some "Port to bind on the local machine (note: ports in SRT are \
+            not related to TCP ports.)";
```
mbugeia: This is confusing, maybe it should be pointed out that these are UDP ports ? Wouldn't it be simplier to have bind_address like 0.0.0.0:8000 ?

src/io/srt_io.ml
```diff
@@ -0,0 +1,439 @@
...
+      Some "Maximum duration of the buffered data." ;
+
+      "payload_size", Lang.int_t, Some (Lang.int 1316),
+      Some "Payload size." ;
+
+      "messageapi", Lang.bool_t, Some (Lang.bool true),
+      Some "Use message api" ;
+
+      "", Lang.string_t, None,
+        Some "Mime (Content-Type) used to find a decoder for the iput stream." ]
```
mbugeia:
```suggestion
        Some "Mime (Content-Type) used to find a decoder for the input stream." ]
```

src/io/srt_io.ml
```diff
@@ -0,0 +1,439 @@
...
+    self#close_socket;
+    feeding_thread <- None
+end
+  
+let () =
+  let kind = Lang.univ_t 1 in
+  Lang.add_operator "input.srt"
+    ~kind:(Lang.Unconstrained kind)
+    ~category:Lang.Input
+    ~descr:"Receive a SRT stream."
```
mbugeia: Maybe be more precise in what this input is doing. Something like "Start a SRT agent in listener mode to receive and decode stream." ?

src/io/srt_io.ml
toots: I'll clarify. Here's what the doc says, which isn't very clear either:
>The term "port" used in SRT is occasionally identical to the term "UDP port". However SRT offers more flexibility than UDP (or TCP, if we think about the more logical similarity) because it manages ports as its own resources. For example, one port may be shared between various services.

src/io/srt_io.ml
toots: 👍

PR #906, opened 2019-08-04
Add LV2 support.

smimram:


Yamakaky:
Nice!

toots:
This looks good to me. I haven't looked too much in details the source implementation but I'll have to trust you on this.

Couple of remarks:
* Would you mind merging this with `1.5.0`? I'd like to have some exposure to test users before shipping it
* It'd be nice if the underlying binding could be more documented. At least a cleaned up `.mli` 🙂

smimram:
Thanks for the comments. I added a mli, and I am fine for waiting next major release.

PR #917, opened 2019-08-17
Add for and while loops.

smimram:
Question: do we also want to add a more usual syntax for those?


toots:
Shouldn't we do this at the parser/lexer level instead?

PR #920, opened 2019-08-21
Native implementation of lists.

smimram:
Implement most list functions in Liquidsoap. Type generalization is not smart enough for now on recursive functions for this to be useable.

Do we want this to be integrated at some point? I tend to think that yes since more people are able to read Liquidsoap than OCaml in our community. However, there is a little slowdown on lists (but I guess this is pretty ok for the typical usage of lists).


smimram:
The generalization problem is explained in #921.

toots:
Looks great to me! Let's hold on for `1.5`, fix the generalization issue and merge!

Issue #895, opened 2019-07-24
Bad positions when reporting errors


smimram:
The positions of errors are quite off sometime. A simple example:
```
print(11 == 2.)
```
results in
```
At line 1, char 10-11:
Error 5: this value has type
  float
but it should be a subtype of int (inferred at line 1, char 6)
```
but characters 10 and 11 are `= `...

toots:
I can confirm that this is a regression from `1.3.7`..

smimram:
I have tried [restoring ocamllex in a branch](https://github.com/savonet/liquidsoap/tree/ocamllex) and still have the same problems. It might rather come from menhir?

smimram:
Ok, so consider the following file

```
print = df == Xaaaaaaaa * 123
```
Do
```
export LIQUIDSOAP_DEBUG_LANG=true
```
and then run Liquidsoap on the file. You will see the positions assigned to expressions:
```
at line 1, char 6 (df): assigned type var ?98
at line 1, char 11-13 (Xaaaaaaaa): assigned type var ?99
at line 1, char 24 (123): assigned type var ?100
at line 1, char 14-23 (*): assigned type var ?101
at line 1, char 11-25 (*(Xaaaaaaaa,123)): assigned type var ?102
at line 1, char 8-10 (==): assigned type var ?103
at line 1, char 6-25 (==(df,*(Xaaaaaaaa,123))): assigned type var ?104
at line 0 char 0 - line 1 char 25 (()): assigned type var ?105
at line 0 char 0 - line 1 char 25 (<?>): assigned type var ?106
```
As you can see there is some kind of mixing up postions: a token gets the position of the previous one...

smimram:
NB: this one has good error reporting:
```
322 = XXX + 2
```
at least for the 322:
```
At line 1, char 0-3:
Error 3: Syntax error!
at line 0 char 0 - line 1 char 0 (322): assigned type var ?98
```

PR #924, opened 2019-08-22
Remove sedlexing_compat.

smimram:
Trying to improve #895 without much success (but the code is cleaner
anyway).


toots:
If you've tested it, it looks good to me.

Issue #921, opened 2019-08-21
No generalization for recursive functions


smimram:
Recursive functions do not seem to have their type correctly generalized. Consider
```
def f(x) = x end
def rec g(x) = x end
```
we respectively get the types
```
f     : ('a) -> 'a
g     : (?A) -> ?A
```
which means that `g` is not polymorphic...

smimram:
The following fixes the bug but I am not sure about the implications:
```
diff --git a/src/lang/lang_values.ml b/src/lang/lang_values.ml
index 9c81d7ee..ee8eb101 100644
--- a/src/lang/lang_values.ml
+++ b/src/lang/lang_values.ml
@@ -546,6 +546,7 @@ let (>:) = T.(>:)
 let rec value_restriction t = match t.term with
   | Var _ -> true
   | Fun _ -> true
+  | RFun _ -> true
   | List l | Tuple l -> List.for_all value_restriction l
   | _ -> false
 
```
@dbaelde could you comment on it?

smimram:
With the above patch we seemingly get the right type but the following is not working
```
def rec recid(x)
  x
end
ignore(recid(3) == 3)
ignore(recid("a") == "a")
```
although `recid` has type
```
('a) -> 'a
```

smimram:
Narrowing it down. The following more minimal example is working
```
def id (x) = x end
_ = id(3)
_ = id("a")
```
but not
```
def rec recid (x) = x end
_ = recid(3)
_ = recid("a")
```
However, the following is working
```
def rec recid (x) = 5 end
_ = recid(3)
_ = recid("a")
```
Somehow the generalization of the type does not refresh the type of the variable `x` in the body of `recid`. This is validated with
```
export LIQUIDSOAP_DEBUG_LANG=true
```
with the following patch to show runtime instantiations:
```
diff --git a/src/lang/lang_values.ml b/src/lang/lang_values.ml
index ee8eb101..e6db148a 100644
--- a/src/lang/lang_values.ml
+++ b/src/lang/lang_values.ml
@@ -863,6 +863,9 @@ let instantiate ~generalized def =
 let lookup env var ty =
   let generalized,def = List.assoc var env in
   let v = instantiate ~generalized def in
+    if debug then
+      Printf.eprintf "Runtime instantiate %s[%d] : %s becomes %s\n"
+        var (T.deref def.V.t).T.level (T.print def.V.t) (T.print v.V.t);
     if debug then
       Printf.eprintf
[...]

PR #934, opened 2019-08-27
Implementation of recursive functions without delayed terms.

smimram:
Trying to implement recursive function with terms instead of unit -> term.

This should not be merged (for now), see #921.


smimram:
Looks mostly ok to me now. There is a hack at https://github.com/savonet/liquidsoap/pull/934/files#diff-84ccbca8728deab9d31a31fe3cbdc986R475 but it's better than the current situation anyway.

smimram:
I now have a clean implementation by using lazy values for environments. No more dirty hacks!

This can be merged once 1.4.0 is released.

PR #943, opened 2019-08-31
%ffmpeg encoder!

toots:


smimram:
Great work, thanks!

toots:
Thanks! Have to admit, @gndl's awesome work on `ocaml-ffmpeg` made this very nice to work on.

src/encoder/ffmpeg_encoder.ml
```diff
@@ -0,0 +1,127 @@
...
+  let dst_channels =
+    match ffmpeg.Ffmpeg_format.channels with
+      | 1 -> `Mono
+      | 2 -> `Stereo
+      | _ -> failwith "%ffmpeg encoder only supports mono or stereo audio for now!"
+  in
+  let buf = Buffer.create 1024 in
+  let make () =
+    let converter =
+      Resampler.create ~out_sample_format
```
smimram: Can't we use ffmpeg to do the resampling here?

toots: I think it's doing it no?

smimram: isn't `Resampler` using libsamplerate by default?

toots: In this context, `Resampler` is a module generated via `Swresample` above that converts to the native `frame` format for `FFmpeg`.

smimram: Ah ok then, sorry!

toots:
That branch is called `ffmpeg-decoder`.. Fail.. Re-opening..

PR #947, opened 2019-09-02
Resample using ffmpeg.

smimram:
Not tested yet, but should work pretty well :)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -46,6 +46,7 @@ New:
 - Added `{input,output}.srt` (#898)
 - Added `path.remove_extension`.
 - Added SSL read/write timeout options, use it for incoming socket connections (#932)
+- Added ffmpeg resampler (#947).
 
 Changed:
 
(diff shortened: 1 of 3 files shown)


smimram:
This can be merged in my opinion. You can test with
```
set("log.level", 5)
set("audio.converter.samplerate.preferred", "ffmpeg")

s = playlist("~/Music")
s = mksafe(s)
s = stretch(ratio=1.2, s)
output.pulseaudio(s)
```

toots:
Look good to me, thanks!

smimram:
Thanks for merging! BTW, it seems to be better than samplerate with which we are secretly loosing some samples sometimes...

toots:
Ho wow. Wanna make it the default before samplerate when found?

smimram:
I think it could be a good idea, yes. Where is the default chosen?

toots:
@smimram here: https://github.com/savonet/liquidsoap/blob/master/src/converters/audio_converter.ml#L55
You might want to switch this to an array like we do for decoders..

PR #950, opened 2019-09-03
Use bigarrays for audio buffers.

### LINKED ISSUE #946: Change audio buffer formats
We should change audio buffers from `float array` to bigarrays: this way, we could avoid copies when passing data to C libraries.

@toots do you think it's worth the work?

smimram:
You need https://github.com/savonet/ocaml-mm/pull/9 of ocaml-mm.

Should fix #946.

Compiling but roadwork ahead (we should fix the TODO, which are raising warnings at compilation time).


smimram:
Note for later:
```
./liquidsoap ../libs/pervasives.liq 'output.pulseaudio(mksafe(buffer(stretch(input.pulseaudio(clock_safe=false), ratio=2.))))'
```
is failing.

PR #955, opened 2019-09-10
List of converters.

smimram:
Have a list of samplerate converters to try.


dhannyz:
Have been wondering about ffmpeg for decoding + samplerate conversion, does this 'inline' the resampling in the decoding pass?

toots:
Resampling when using the ffmpeg decoder should already be done by ffmpeg, this PR adds support for using it with other decoders/encoders.

PR #979, opened 2019-10-09
Switch to Unix.read style API.

toots:


src/decoder/external_decoder.ml
```diff
@@ -31,10 +31,12 @@ let log = Log.make ["decoder";"external"]
...
   * the available output. *)
 let external_input process input =
+  let buflen = 1024 in
+  let buf = Bytes.create buflen in
   let on_stdin pusher =
-    let s,read = input.Decoder.read 1024 in
+    let read = input.Decoder.read buf 0 buflen in
     if read = 0 then `Stop else begin
-      Process_handler.write (Bytes.of_string s) pusher;
+      Process_handler.write (Bytes.sub buf 0 read) pusher;
```
smimram: I am not following the details, but can't we generalize `Process_handler.write` to take an offset and a length instead of having to use `Bytes.sub` here?

toots: This is a convenience API. Having it take offsets/length makes it moot.. Might just get rid of it.

PR #984, opened 2019-10-11
Have encoders return lists of strings

smimram:
Right now encoder have to make lots of concatenations which is unnecessary. This is mitigated a bit by the use of `Buffer`, but really we should avoid copies at all price. I thus propose to have encoders return lists of strings, and never really concatenate those. I have introduced a new module `Strings` for operations on these.


src/outputs/harbor_output.ml
```diff
@@ -504,9 +504,10 @@ module Make (T : T) = struct
                    on_connect ~protocol ~uri ~headers:h_headers ip )) )
 
       method send b =
-        let slen = String.length b in
+        let slen = Strings.length b in
+        failwith "TODO"; let b = assert false in
```
smimram: @toots I need some help for this last one, I am not really sure about the semantics of everything (burst in particular)... I guess that some `Buffer.t` can be replaced by `String.t`.

src/io/srt_io.ml
```diff
@@ -441,13 +441,12 @@ object (self)
...
-        Buffer.blit buffer 0 tmp 0 payload_size;
-        Utils.buffer_drop buffer payload_size) ();
+          Strings.blit !buffer 0 tmp 0 payload_size;
+          buffer := Strings.drop !buffer payload_size) ();
       let rec f = function
         | pos when pos < payload_size ->
-          let ret =
-            send (Bytes.sub tmp pos (payload_size-pos))
-          in
+          (* TODO: it would be better if Srt.send / Srt.sendmsg would take an offset... *)
```
smimram: @toots don't you think?

src/outputs/harbor_output.ml
```diff
@@ -204,59 +202,78 @@ module Make (T : T) = struct
...
-          | Some s ->
-              Duppy.Monad.Io.write ?timeout:(Some c.timeout)
-                ~priority:Tutils.Non_blocking c.handler (Bytes.of_string s) )
+            else
+              let rec write = function
+                | s::l ->
+                  Duppy.Monad.bind
+                    (Duppy.Monad.Io.write ?timeout:(Some c.timeout)
+                       ~priority:Tutils.Non_blocking c.handler
+                       (* TODO: can we use unsafe_of_string here? *)
```
smimram: @toots I would expect `Duppy.Monad.Io.write` to take a `string` not a `bytes`

src/io/srt_io.ml
toots: The C API does a copy of the string to run without the global lock. I think that I have tried registering the string as global root but it didn't work. I'll give it another try,

src/outputs/harbor_output.ml
toots: I think that this is mimicking the `Unix` API, which also takes a `bytes` argument now.

toots: And, yes, `unsafe_of_string` is totally acceptable I believe.

src/tools/strings.ml
```diff
@@ -0,0 +1,120 @@
+(** Operations on lists of strings. This is module is used in order to avoid
```
toots: Missing header.. 🙂

src/tools/strings.ml
```diff
@@ -0,0 +1,120 @@
...
+    concatenating (large) strings. Iterators are FIFO. *)
+
+(* List of "concatenated" strings, stored backwards. *)
+type t = string list
```
toots: We should make this API thread-safe and add mutex/mutexify behind the scene.

toots: Also, adding offset/length is indeed extra work but would be totally worth it.

smimram: > Also, adding offset/length is indeed extra work but would be totally worth it.

If you want to do this, I would go for a "reimplementation" of the `String` module (say `OString`) which takes care of that transparently. This would mean that the module `OString` would look like
```
type t = string * int * int

let sub (s,o,l) o' l' = s (o+o') l'
```
and so on.

src/outputs/harbor_output.ml
```diff
@@ -395,9 +406,8 @@ module Make (T : T) = struct
 
       val mutable chunk_len = 0
 
-      val mutable burst_data = []
-
-      val mutable burst_pos = 0
+      val burst_data =
+        Strings.Mutable.of_strings Strings.empty
```
toots:
```suggestion
        Strings.Mutable.empty ()
```

src/tools/process_handler.ml
```diff
@@ -332,13 +332,14 @@ let run ?priority ?env ?on_start ?on_stdin ?on_stdout ?on_stderr ?on_stop ?log c
     Duppy.Task.add Tutils.scheduler (get_task handler (on_start (pusher fd)));
     t
 
-let really_write data push =
-  let len = Bytes.length data in
+let really_write ?offset ?length data push =
```
toots:
```suggestion
let really_write ?(offset=) ?length data push =
```

src/tools/process_handler.ml
```diff
@@ -332,13 +332,14 @@ let run ?priority ?env ?on_start ?on_stdin ?on_stdout ?on_stderr ?on_stop ?log c
...
+  let offset = match offset with Some offset -> offset | None -> 0 in
```
toots:
```suggestion
```

src/tools/process_handler.ml
```diff
@@ -332,13 +332,14 @@ let run ?priority ?env ?on_start ?on_stdin ?on_stdout ?on_stderr ?on_stop ?log c
...
+  let length = match length with Some length -> length | None -> Bytes.length data in
```
toots:
```suggestion
  let length =
    match length with
        | Some length -> length
        | None -> Bytes.length data
  in
```

src/tools/utils.ml
```diff
@@ -191,18 +191,20 @@ let rec may_map f = function
...
-  let contents = ref Strings.empty in
+  let contents =
+    Strings.Mutable.empty ()
+  in
   let rec read () =
     let ret = input channel tmp 0 pagesize in
     if ret > 0 then
       begin
-        contents := Strings.add_subbytes !contents tmp 0 ret;
+        Strings.Mutable.unsafe_add_subbytes contents tmp 0 ret;
```
smimram: Is this unsafe really safe? I think that the buffer `tmp` is going to be reused isn't it?

src/encoder/avi_encoder.ml
```diff
@@ -76,24 +76,26 @@ let encoder avi =
...
+    Avi.header ~channels ~samplerate ()
+  in
   let need_header = ref true in
   let encode frame start len =
     let ans = encode_frame ~channels ~samplerate ~converter frame start len in
     if !need_header then
-      (
+     begin
         need_header := false;
-        Strings.append header ans
```
smimram: Actually `append` was a better name than `dda`... :)

src/encoder/external_encoder.ml
```diff
@@ -88,15 +88,15 @@ let encoder id ext =
...
     `Continue)
   in
 
-  let flush_buffer = Tutils.mutexify mutex (fun () ->
-      let ans = !buf in
-      buf := Strings.empty;
-      ans)
+  let flush_buffer () = 
+    let ans = Strings.Mutable.to_strings buf in
+    Strings.Mutable.flush buf;
```
smimram: I think that is would be more natural to have `flush` both empty the buffer *and* return the previous contents, so that this would be only one line.

src/outputs/harbor_output.ml
```diff
@@ -535,11 +523,8 @@ module Make (T : T) = struct
               true )
             else false
           in
-          burst_data <- (
-            let bd = Strings.append burst_data b in
-            let len = Strings.length bd in
-            Strings.sub bd (max 0 (len-burst)) (min len burst)
-          );
+          Strings.Mutable.append_strings burst_data b;
+          Strings.Mutable.keep burst_data burst;
```
smimram: My `sub` is more simple than `keep` here: since we now have string views, I would simply remove `keep` and use `sub`.

src/outputs/harbor_output.ml
```diff
@@ -551,13 +536,16 @@ module Make (T : T) = struct
                       (fun () ->
                         match c.state with
                         | Hello ->
-                            let bdlen = Strings.length burst_data in
-                            c.buffer <- Strings.append c.buffer (Strings.sub burst_data (max 0 (bdlen - burst)) (min bdlen burst));
+                            let bdlen = Strings.Mutable.length burst_data in
+                            let data =
+                              Strings.Mutable.sub burst_data (max 0 (bdlen - burst)) (min bdlen burst)
```
smimram: You don't need to `sub` here if you do the above change from `keep` to `sub`.

src/tools/bytes_buffer.mli
```diff
@@ -0,0 +1,106 @@
...
+  GNU General Public License for more details, fully stated in the COPYING
+  file at the root of the liquidsoap distribution.
+
+  You should have received a copy of the GNU General Public License
+  along with this program; if not, write to the Free Software
+  Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301  USA
+
+ *****************************************************************************)
+
+(** Bytes buffers where the main operation is to add bytes at the end.
```
smimram: This is crazy and we don't have the corresponding ml, please remove this mli file.

src/tools/strings.ml
```diff
@@ -1,144 +1,275 @@
...
+  (** Add a string at the end of a buffer. *)
+  val add : buffer -> string -> return
+
+  (** Add bytes at the end of a buffer.
+    * bytes will be copied. *)
+  val add_bytes : buffer -> bytes -> return
+
+  (** Add bytes at the end of a buffer.
+    * bytes will not be copied. *)
+  val unsafe_add_bytes : buffer -> bytes -> return
```
smimram: I would rather not have such functions in the API and let the user do the `Bytes.unsafe_to_string` by himself: we want to make it hard for everybody to do unsafe casts since they are potentially quite unsafe (e.g. see my comments on your other changes).

src/outputs/harbor_output.ml
smimram: Yes, why didn't you commit this change?

src/encoder/external_encoder.ml
```diff
@@ -88,15 +88,15 @@ let encoder id ext =
       let len = puller bytes 0 Utils.pagesize in
       match len with
         | 0 when !is_stop -> Condition.signal condition
-        | _ -> buf := Strings.add_subbytes !buf bytes 0 len
+        | _ -> Strings.Mutable.unsafe_add_subbytes buf bytes 0 len
```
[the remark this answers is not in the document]
toots: Good catch.
```suggestion
        | _ -> Strings.Mutable.add_subbytes buf bytes 0 len
```

src/encoder/external_encoder.ml
toots: Good idea.

src/outputs/harbor_output.ml
toots: I'm waiting for more to commit all of them in a batch..

src/outputs/harbor_output.ml
toots: They are very similar indeed. I added `keep` to simplify the semantics of the code, making it easier to read without having to do the math (and potentially get it wrong).

src/outputs/harbor_output.ml
toots: I'm not sure what you are referring to. It actually looks to me like that whole `sub` call is not needed now that I'm looking at it. The original code wasn't great (or else, wasn't abstracted like we just did) so its real intent was hidden. I believe the `add`+`keep` is all it was supposed to do, passing `burst_data` as a whole to the client upon connecting.

src/tools/utils.ml
toots: Good catch, thanks.

toots:
```suggestion
        Strings.Mutable.add_subbytes contents tmp 0 ret;
```

src/tools/strings.ml
toots: I think any pattern that leads to repeated nested call should be abstracted away in a API call to make the code more easily readable. Let's see what's left of this API after we commit the fixes you spotted.

src/outputs/harbor_output.ml
smimram: Yes, you already `sub` above, so you don't need it at all here. In my old code `keep` would keep *at least* the required amount of data (we did not have string views) so that I would need to `sub` again when using it.

src/tools/strings.ml
```diff
@@ -1,144 +1,275 @@
...
   Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301  USA
-
+  
  *****************************************************************************)
-
+  
 (** Operations on lists of strings. This is module is used in order to avoid
     concatenating (large) strings. Iterators are FIFO. *)
 
+module type Strings = sig
```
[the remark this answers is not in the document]
toots: This is the way a lot of modern OCaml is written. It helps leveraging the rich module typing tools to declare a consistent API contract for each sub-module implementing it. See for instance: https://github.com/ocamllabs/ocaml-integers/blob/master/src/signed.mli

I like it this way. We should split into 3 different files, though, one signature file and two implementation files.

src/tools/strings.ml
```diff
@@ -147,25 +278,77 @@ module Mutable = struct
...
+    blit m.strings b n) ()
+
+  let sub m ofs len = mutexify m.mutex (fun () ->
+    of_strings (sub m.strings ofs len)) () 
+
+  let substring m ofs len = mutexify m.mutex (fun () ->
+    substring m.strings ofs len) ()
 end
+
+include Immutable
```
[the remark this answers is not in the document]
toots: The main reasons for its existence is to make it possible to refer to `Immutable.t` inside `Mutable` with another `type t` there.. Can be solved by splitting them into two files.

src/tools/strings.ml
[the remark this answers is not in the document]
smimram: There is a major difference. In the example you give there are three implementations of the same module `Int`, `Int32` and `Int64` whereas in our case there will never be more than one implementation of `Imperative`...

src/tools/strings.ml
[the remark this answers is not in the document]
smimram: Yes or simply having an alias for `t`:
```
type buffer = t
```
which would be much lighter in terms of code.

src/encoder/external_encoder.ml
```diff
@@ -88,17 +88,11 @@ let encoder id ext =
       let len = puller bytes 0 Utils.pagesize in
       match len with
         | 0 when !is_stop -> Condition.signal condition
-        | _ -> Buffer.add_subbytes buf bytes 0 len
+        | _ -> Strings_mutable.unsafe_add_subbytes buf bytes 0 len
```
toots:
```suggestion
        | _ -> Strings_mutable.add_subbytes buf bytes 0 len
```

PR #986, opened 2019-10-12
Format codebase with ocamlformat.

toots:
Just an experiment for now. Would be nice to use in conjunction with a commit hook but these are hard to share..


smimram:
Please use ocp-indent. This is more standard and more widely supported by editors: emacs or vim can use that thus avoiding to have weird indentation when editing.

PR #989, opened 2019-10-13
Encode list, toots version

toots:


smimram:
Can we close this PR, since we mostly agree on #984?

toots:
Sure but since it's merge in your branch, it should also be closed automatically when that branch is merged ;-)

PR #1001, opened 2019-10-22
Honor DESTDIR.

### LINKED ISSUE #998: Bad handling of opam install --destdir
**Describe the bug**
When building liquidsoap from opam with the `--destdir` flag, the binary keeps reference to the local opam install.

**To Reproduce**
```
# opam install --destdir=/tmp/liquidsoap liquidsoap
# grep opam /tmp/liquidsoap/bin/liquidsoap
/home/opam/.opam/4.09/lib/ocaml/caml
/home/opam/.opam/4.09/.opam-switch/build/biniou.1.2.1/_build/default
/home/opam/.opam/4.09/.opam-switch/build/bjack.0.1.5/src
/home/opam/.opam/4.09/.opam-switch/build/...
/home/opam/.opam/4.09/lib
/home/opam/.opam/4.09/lib/liquidsoap/share/liquidsoap/1.4.0/bin
/home/opam/.opam/4.09/lib/liquidsoap/share/liquidsoap/1.4.0/libs
/home/opam/.opam/4.09/lib/liquidsoap/var/log/liquidsoap
/home/opam/.opam/4.09/lib/liquidsoap/var/run/liquidsoap
/home/opam/.opam/4.09/lib/ocaml
/home/opam/.opam/4.09/lib/ocaml/caml
/home/opam/.opam/4.09/lib/ogg
/home/opam/.opam/4.09/share/camomile
```

**Expected behavior**
The package in destdir should be self contained and not keep reference to the opam local install. At least that how I understand it. I'm not very familliar to opam so no idea how to fix it.

My usecase is that I use a opam docker image as first stage then copy the files from destdir in the final image. Currently, I have to also copy `.opam/4.09`, which should not be needed and adds a lot to the final image size.

**Version details**
 - OS: alpine
 - Version: opam 1.4.0

smimram:
This allows local installation and should fix #998.

I also added `--disable-daemon` option, otherwise local installation will not work as normal user.

### DIFF
--- a/Makefile.defs.in
+++ b/Makefile.defs.in
@@ -8,7 +8,7 @@ user=@install_user@
 group=@install_group@
 INSTALL_DAEMON=@INSTALL_DAEMON@
 
-prefix= @prefix@
+prefix= $(DESTDIR)@prefix@
 exec_prefix= @exec_prefix@
 bindir= @bindir@
 libdir= @libdir@
--- a/configure.ac
+++ b/configure.ac
@@ -110,23 +110,31 @@ fi
 AC_SUBST(install_user)
 AC_SUBST(install_group)
 
-AC_MSG_CHECKING(whether user $install_user exists)
-if ! id $install_user >/dev/null 2>&1 ; then
+AC_ARG_ENABLE([daemon],AS_HELP_STRING([--disable-daemon],[don't install daemon]))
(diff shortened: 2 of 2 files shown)


smimram:
This is already handled, see #998.

PR #1005, opened 2019-10-23
fix docs for http/https builtins

mcfiredrill:
The docs currently say `return (status,headers),data` but unless I'm
mistaken this should be `return (status,headers,data)`.


smimram:
Thanks for the PR!

Issue #949, opened 2019-09-02
Enqueue request from script instead of command


Nirei:
I'm trying to write my very first liquidsoap program. It goes something like this:

```
sounds_path = "../var/sounds"

# Log file
set("log.file.path","var/log/liquidsoap.log")
set("harbor.bind_addr", "127.0.0.1")
set("harbor.timeout", 5)
set("harbor.verbose", true)
set("harbor.reverse_dns", false)

silence = blank()
queue = request.queue()

def play(~protocol, ~data, ~headers, uri) =
    request.push("#{sounds_path}#{uri}")
    http_response(protocol=protocol, code=20000)
end

harbor.http.register(port=8080, method="POST", "^/(?!\0)+", play)

stream = fallback(track_sensitive=false, [queue, silence])
...output.whatever...
```

And I was wondering if there is any way to push to the queue from the harbor callback.

Else, how should I proceed about making requests originate from HTTP calls? I really want to avoid telnet. My final objective is having an endpoint that I can call to make my stream play a file on demand and be silent the rest of the time.

smimram:
Actually, this is quite a good question and I am not sure this is currently possible without telnet (whereas it really should and telnet should be registered from Liquidsoap...)

@toots ?

Nirei:
So I was thinking I could try to make myself useful.

After an interesting evening getting remotely aquainted with the codebase I tried putting this in the initializer of the ``Req_queue.queue`` class:

```ocaml
      Lang.add_builtin "request.queue.push"
        ~category:(Lang.string_of_category Lang.Liquidsoap)
        ~descr:"Add a request to a queue."
        [ "", Lang.request_t k, None, None ]
        Lang.unit_t
        (fun p ->
            let r = Lang.to_request (Lang.assoc "" 2 p) in
            self#push_request r;
            fun (_:Lang.t) -> Lang.unit)
```

But no use: my builtin is not recognized when I try to call it from my liquidsoap script.

Also, I didn't quite grasp why my ``fun p -> ...`` kept complaining about its return type having to be a function instead of pure ``Lang.unit``.

If you are interested and can give me some pointers I can try to write a PR for this?

Very cool software anyway, folks, and thanks for your time. : )

smimram:
#1010 should be quite helpful for this one, hold on.

smimram:
This should be solved in #1013 which I guess is going to be merged soon (feel free to try before that...)

Nirei:
[...]

PR #1010, opened 2019-10-28
Commands

smimram:
The idea of commands is to be able to registers effects on sources. Trying some design.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,6 +8,7 @@ New:
 - Added `string.nth` (#970).
 - Added `string.binary.to_int` (#970).
 - Added `file.ls`.
+- Added support for commands (#1010).
 
 Changed:
 
@@ -17,6 +18,8 @@ Changed:
 - Use system pagesize for buffer allocation (#915).
 - Use new Strings module in order to avoid concatenations (#984).
 - Native Liquidsoap implementation of list functions (#920).
+- Changed the type of `playlist.reloadable`: the skip function is now a command
+  instead of being returned (#1010).
 
 Fixed:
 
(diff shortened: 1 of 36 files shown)


smimram:
A typical example I am trying to get working is
```
reload = ?
s = playlist.reloadable(cmd_reload=reload, "pl")
add_timeout(2., fun () -> begin reload(); 2. end)
out(s)
```

smimram:
@toots you can now play 2 sec of your mp3 files with:
```
s = mksafe(playlist("~/Music"))

skip = ?
output.pulseaudio(cmd_skip=skip, s)

def skip () =
  print("Skipping!");
  skip ();
  2.
end
add_timeout(2., skip)
```
Note that we can access the skip command of each output, but the type is essentially unchanged (i.e. it should be backward compatible).

What do you think of that?

smimram:
Another example, we can push in queues directly from Liq:
```
push = ?
s = request.queue(cmd_push=push)
s = mksafe(s)

def push () =
  push("test.mp3")
  2.
end
add_timeout(2., push)

output.pulseaudio(s)
```
which solves #949.

smimram:
Just added an option to allow unsafe use of commands. It is actually quite useful. For instance, we can list the metadata of all files in a directory with
```
#!./liquidsoap --unsafe-commands ../libs/pervasives.liq

s = mksafe(playlist(mode="normal", "~/Music"))

skip = ?
n = ref 0
def f(m) =
  print("Metadata: #{m}")
  skip ()
  n := !n + 1
  if !n > 10 then exit(0) end
end
s = on_metadata(f, s)
output.dummy(cmd_skip=skip, s)
```

smimram:
Keeping this open for now, but it should be superseeded by #1197.

smimram:
Closing now that we have records as per #1197

PR #1012, opened 2019-10-30
Implement per-frame auto-sync mechanism

toots:
This PR implements a new per-frame synchronization mechanism. On each frame production, the clock checks if one of the source filling its frame is declared as `self_sync`.

`self_sync` means that the source has its own notion of time/synchronization. Typically alsa, ao sources or SRT I/O. However, it does not include sources that can be blocking at times but without any relation to a notion of time such as icecast output.

All the previous classes of clock have now been merged into a single class, with 3 possible behaviors:
* Synchronize frame production with the CPU clock, regardless of `self_sync` (`"cpu"` mode)
* No synchronization control at all (`"none"` mode)
* On each frame production, synchronize with the CPU clock only if none of the active sources are declared as `self_sync` (`"auto"` mode, new default)

This was done with an approximation in mind: if there are multiple sources with different synchronization mechanisms (for instance the end of an alsa input followed by a blank or the end of a SRT input followed by another) then we only account for `self_sync` as declared by one source only (I believe the first one).

The reasons for this choice are:
- There is simply no consistent notion of time in this case
- We assume that, in most cases, this only happen on track changes and for an amount of data (or stream time!) that is negligible.

Another thing to note: the old `clock_safe` mechanisms still apply. This means that, to have a case where two sources with two different sources of synchronization fill the same frame, you need to explicitly declare them as `self_sync=false`.

Furthermore, the old mechanism for `self-sync` clock is also preserved, with the added case that situations like this:
```
fallback([input.alsa(),blank()]
```
Are now guaranteed to never loop with no synchronization on the `blank()` source. This was possible before because `input.alsa` would tell the source to delegate its synchronization but might not be actually the source being pulled.

giphy


src/clock.ml
```diff
@@ -89,8 +89,49 @@ let leave (s : active_source) =
...
+   * Thread.delay uses Unix.select which can raise EINTR.
+   * A really good implementation would keep track of the elapsed time and then
+   * trigger another Thread.delay for the remaining time.
+   * This cheap thing does the job for now.. *)
+  try Thread.delay d with Unix.Unix_error (Unix.EINTR, _, _) -> ()
+
+(** In [`CPU] mode, synchronization is governed by the CPU clock.
+  * In [`None] mode, there is no synchronization control (except for
+  * active sources latencies)
+  * In [[Audo] mode, synchronization is governed by the CPU unless at
```
smimram:
```suggestion
  * In [`Auto] mode, synchronization is governed by the CPU unless at
```

src/clock.ml
```diff
@@ -89,8 +89,49 @@ let leave (s : active_source) =
...
+
+let usleep d =
+  (* In some implementations,
+   * Thread.delay uses Unix.select which can raise EINTR.
+   * A really good implementation would keep track of the elapsed time and then
+   * trigger another Thread.delay for the remaining time.
+   * This cheap thing does the job for now.. *)
+  try Thread.delay d with Unix.Unix_error (Unix.EINTR, _, _) -> ()
+
+(** In [`CPU] mode, synchronization is governed by the CPU clock.
```
smimram: I not sure about this comment: above, you say that cpu corresponds to "cpu burning mode" where you produce as fast as possible, but this comment sounds like you use the cpu clock to achieve something like realtime.

src/clock.ml
```diff
@@ -89,8 +89,49 @@ let leave (s : active_source) =
...
+  * In [`None] mode, there is no synchronization control (except for
+  * active sources latencies)
```
smimram: Please be more precise about these "active source latencies", do you mean that they can be blocking?

src/clock.mli
```diff
@@ -24,19 +24,19 @@
...
   * which prevents inconsistent uses of the source. Clocks are assigned to
   * sources at the end of the typing phase. *)
-class clock : string -> Source.clock
 
-(** Wallclocks are clocks attached to a thread that periodically makes the
-  * clock tick, which triggers the streaming of the active sources attached
-  * to the clock. *)
-class wallclock : ?sync:bool -> string -> clock
+(** In [`CPU] mode, synchronization is governed by the CPU clock.
+  * In [`Mone] mode, there is no synchronization control (except for
```
smimram:
```suggestion
  * In [`None] mode, there is no synchronization control (except for
```

src/clock.mli
```diff
@@ -24,19 +24,19 @@
...
+  * active sources latencies)
+  * In [[Audo] mode, synchronization is governed by the CPU unless at
```
smimram:
```suggestion
  * In [`Auto] mode, synchronization is governed by the CPU unless at
```

src/io/oss_io.ml
```diff
@@ -71,15 +73,9 @@ object (self)
           Unix.close x ;
           fd <- None
 
-  method output_start =
-    if clock_safe then
-      (get_clock ())#register_blocking_source ;
```
smimram: Why don't you have to register/unregister anymore?

src/source.ml
```diff
@@ -58,7 +58,7 @@ type source_t = Fallible | Infallible
   *
   * Most clocks are passive, i.e. they don't run anything
   * directly, but may only tick when something happens to a source.
-  * The wallclock (formerly root.ml) is the default, active clock:
+  * The clock (formerly root.ml) is the default, active clock:
```
smimram:
```suggestion
  * The clock is the default, active clock:
```
I don't think that many people remember about `root.ml` :)

src/clock.ml
toots: I'm trying to clarify the terminology. "cpu burning mode" is now gone, `CPU` means just that the CPU clock (I believe the one on the motherboard) is driving synchronization by keeping track of the time elapsed while producing each frame and sleeping or accelerating accordingly.

src/io/oss_io.ml
toots: That was the way for a source to declare that it had self-synchronization capabilities. It was done globally when waking up and did not account for wether or not that source was actually being pulled. This mechanism has been brought down to the frame producing level and accounts for the source that is actually filling the frame.

toots:
Okay, this is ready for merge. Gonna ask a review from @dbaelde just in case he wants to chime in. Also, I need to review the operators and outputs' `self_sync` definition to make sure they make sense. @smimram let me know if you have opinions on this, specially with the midi operators, I wasn't always 100% sure.

src/clock.ml
```diff
@@ -147,8 +192,91 @@ class clock id =
...
+               true
+      in
+      begin
+       match self_sync, new_val with
+         | None, false
+         | Some true, false ->
+           log#important "Delegating synchronisation to CPU clock"
+         | None, true
+         | Some false, true ->
+           log#important "Delegating synchronisation to active clocks"
```
toots:
```suggestion
           log#important "Delegating synchronisation to active sources"
```

doc/content/clocks.md
```diff
@@ -181,7 +181,7 @@ If you want to run an output as fast as your CPU allows,
 just attach it to a new clock without synchronization:
 
 ```liquidsoap
-clock.assign_new(sync=false,[output.file(%vorbis,"audio.ogg",source)])
+clock.assign_new(sync="none",[output.file(%vorbis,"audio.ogg",source)])
```
smimram: A question reading this again: do we want to used `clock.assign_new` with `sync` being something else than `"none"` (and example would be nice...)? At least shouldn't this be the default value for the argument?

doc/content/clocks.md
```diff
@@ -181,7 +181,7 @@ If you want to run an output as fast as your CPU allows,
...
 ```
```
smimram: Another remark is that you never give an example of `clock.assign_new` applied to a list of sources with more than one element: are there really usecases for this?

libs/utils.liq
```diff
@@ -420,9 +420,8 @@ end
 # and return the source. It is a conveniency wrapper around clock.assign_new(),
 # allowing more concise scripts in some cases.
 # @category Liquidsoap
-# @param ~sync When `false`, do not synchronize the clock on regular wallclock time, \
-#              but try to run as fast as possible (CPU burning mode).
-def clock(~sync=true,~id="",s)
+# @param ~sync Synchronization mode. One of: `"auto"`, `"cpu"`, or `"none"`.
```
smimram: Some doc please!

doc/content/video.md
```diff
@@ -230,7 +230,7 @@ set("frame.video.samplerate",12)
 s = single("big_buck_bunny_720p_stereo.ogg")
 # This hack is necessary (for now) in order
```
smimram: Is this part of doc still relevant? I don't understand why we would need that here.

src/io/portaudio_io.ml
```diff
@@ -21,7 +21,7 @@
  *****************************************************************************)
 
 (** Dedicated clock. *)
-let get_clock = Tutils.lazy_cell (fun () -> new Clock.self_sync "pa")
+let get_clock = Tutils.lazy_cell (fun () -> new Clock.clock "pa")
```
smimram:
```suggestion
let get_clock = Tutils.lazy_cell (fun () -> new Clock.clock "portaudio")
```

src/io/pulseaudio_io.ml
```diff
@@ -23,7 +23,7 @@
 open Pulseaudio
 
 (** Dedicated clock. *)
-let get_clock = Tutils.lazy_cell (fun () -> new Clock.self_sync "pulse")
+let get_clock = Tutils.lazy_cell (fun () -> new Clock.clock "pulse")
```
smimram:
```suggestion
let get_clock = Tutils.lazy_cell (fun () -> new Clock.clock "pulseaudio")
```

src/io/srt_io.ml
```diff
@@ -141,12 +141,14 @@ object(self)
...
 
+  method self_sync = true
+
   val mutable clock = None
   method private get_clock =
     match clock with
       | Some c -> c
       | None ->
-         let c = new Clock.self_sync self#id in
+         let c = new Clock.clock self#id in
```
smimram: I would have expected to have
```suggestion
         let c = new Clock.clock "srt" in
```
as everywhere else. Maybe the creation function should take both a string an (optional labeled) id?

src/operators/append.ml
```diff
@@ -125,6 +125,11 @@ object (self)
           source#seek n
       | `Append s -> s#seek n 
 
+  method self_sync =
```
smimram: You should add a comment for future generations stating that this is an approximation...

src/operators/comb.ml
```diff
@@ -34,6 +34,7 @@ object
 
   method remaining   = source#remaining
   method seek        = source#seek
+  method self_sync = source#self_sync
```
smimram:
```suggestion
  method self_sync   = source#self_sync
```

src/operators/cross.ml
```diff
@@ -41,6 +41,8 @@ object (self)
   (* This actually depends on [f], we have to trust the user here. *)
   method stype = s#stype
 
+  method self_sync = false
```
smimram: Are you sure for this one? All the data is pre-buffered?

src/operators/cuepoint.ml
```diff
@@ -47,6 +47,7 @@ object (self)
   method stype = source#stype
   method is_ready = source#is_ready
   method abort_track = source#abort_track
+  method self_sync = false
```
smimram: Really? Why not
```suggestion
  method self_sync = source#self_sync
```

src/operators/resample.ml
```diff
@@ -30,6 +30,8 @@ object (self)
   (* Hide our child: we'll treat it specially. *)
   inherit source ~name:"resample" kind as super
 
+  method self_sync = source#self_sync
```
smimram: Ouch this one is really tricky: we are changing the speed of the source so that if it is self-syncing then we will have slower time as output. But I guess this actually does not cause any problem? Slower time is a valid notion of time?

src/operators/sequence.ml
```diff
@@ -33,6 +33,11 @@ object (self)
 
   val mutable seq_sources = sources
 
+  method self_sync =
```
smimram: Comment that there is an approximation.

src/operators/soundtouch_op.ml
```diff
@@ -54,6 +54,8 @@ object (self)
 
   method stype       = source#stype
 
+  method self_sync = source#self_sync
```
smimram:
```suggestion
  method self_sync   = source#self_sync
```

src/operators/switch.ml
```diff
@@ -123,6 +123,11 @@ object (self)
 
   method is_ready = need_eot || selected <> None || self#cached_select <> None
 
+  method self_sync =
```
smimram: I'm not sure I understand this one...

src/operators/switch.ml
```diff
@@ -195,9 +200,6 @@ object (self)
             | None -> ()
           end
     in
-      (* #select is called only when selected=None, and the cache is cleared
-       * as soon as the new selection is set. *)
-      assert (selected=None || cached_selected=None) ;
```
smimram: Why is this going away?

src/outputs/pipe_output.ml
```diff
@@ -314,6 +314,8 @@ object (self)
   inherit piped_output p
   inherit chan_output p
 
+  method self_sync = false
```
smimram: Really for this one? I'd rather have a parameter of the function specifying this since we don't know what binary is going to be called.

src/operators/soundtouch_op.ml
```diff
@@ -54,6 +54,8 @@ object (self)
```
smimram: Did you check that all the business above with clocks is still right?

Incidentally this could be improved (I just stumbled upon this problem myself). When the rate and tempo are both 1., this mens that we are pitch-shifting only, and therefore we should have the same clock as the source (in other cases, we do not).

doc/content/clocks.md
toots: No, auto is what it should be as it covers most of the "simple" situations. I'll add some doc, though.

doc/content/clocks.md
toots: Good point. Yes, for instance a stream involving input and output on the same clock, e.g. jack input & output.

libs/utils.liq
toots:
```suggestion
# @param ~sync Synchronization mode. One of: `"auto"`, `"cpu"`, or `"none"`. Defaults to `"auto"`, which synchronizes with the CPU clock if none of the active sources are attached to their own clock (e.g. ALSA input, etc). `"cpu"` always synchronizes with the CPU clock. `"none"` removes all synchronization control.
```

doc/content/clocks.md
```diff
@@ -181,7 +181,7 @@ If you want to run an output as fast as your CPU allows,
...
 
 This will automatically attach the appropriate sources to that clock.
```
toots:
````suggestion
This will automatically attach the appropriate sources to that clock.

Another important use case of this operator is if your script involves multiple sources from the same external clock, typically multiple ALSA input or output from the same sound card or multiple jack input and output. By default (the so-called `clock_safe` mode), liquidsoap will assign a dedicated clock to each of those sources, leading either to an error or forcing the use of an unnecessary `buffer` (see below). Instead, you can allocate each source with `clock_safe=false` and assign them a single clock:

```
s1 = input.jack(clock_safe=false, ...)
s2 = input.jack(clock_safe=false, ...)

clock.assign_new([s1,s2])
```
````

File: src/operators/switch.ml (hunk 24 above)
Comment (toots) [reply src/operators/switch.ml]: Good catch thanks for asking. I had an `assert` error when testing the code. I couldn't find the source of the error, unfortunately. but I was also able to confirm that this assertion is not enforced in the code. I'll revisit this before merging.

File: src/outputs/pipe_output.ml (hunk 25 above)
Comment (toots) [reply src/outputs/pipe_output.ml]: The process I/O are buffered internally by OCaml when calling the process creation API.

File: doc/content/video.md (hunk 12 above)
Comment (toots) [reply doc/content/video.md]: 👍

File: src/operators/cross.ml (hunk 18 above)
Comment (toots) [reply src/operators/cross.ml]: crossfades need to not be self_sync to function, the source they control needs to be accelerated to compute transitions.

File: src/operators/switch.ml (hunk 23 above)
Comment (toots) [reply src/operators/switch.ml]: If there's a selected source return its `self_sync` otherwise say `false`? In the last case, I believe the source will not be `is_ready` so it should never be pulled anyway.

File: src/operators/resample.ml (hunk 20 above)
Comment (toots) [reply src/operators/resample.ml]: Good remark, thanks. The code below:
```ocaml
  method private set_clock =
    let c = Clock.create_known (new Clock.clock self#id) in
      Clock.unify
        self#clock (Clock.create_unknown ~sources:[] ~sub_clocks:[c]) ;
      Clock.unify source#clock c ;
      (* Make sure the slave clock can be garbage collected, cf. cue_cut(). *)
      Gc.finalise (fun self -> Clock.forget self#clock c) self
```
Assures that the clock is the same as `source#clock`. Thence, it makes sense to also have the same `self_sync`.

If `source` has its own definition of time, the operator will still work but not as expected. It would be nice to catch this statically and fail if the underlying source cannot be accelerated or slowed down but I don't think that this is possible at the moment (in particular because this is a dynamic feature). The same limitation applies to sources used in crossfade transitions.

src/operators/append.ml
toots: I'll add it ad the `source.mli` level.

src/operators/cross.ml
toots: Now that I'm thinking about it.. 😅

src/operators/soundtouch_op.ml
toots: I just pushed some changes, thanks for noticing this. Could confirm that they implement the right thing?

src/outputs/pipe_output.ml
smimram: But really that much buffered? I guess we only have pipes here (= 65 kB), and read / write should be blocking when they are full, which should be enough to sync, isn't it?

src/operators/soundtouch_op.ml
```diff
@@ -54,6 +59,9 @@ object (self)
 
   method stype       = source#stype
 
+  method self_sync   =
+    if rate = 1. then source#self_sync else false
```
smimram: This does not even compile, because `rate` is a float getter...

src/operators/soundtouch_op.ml
```diff
@@ -87,7 +95,8 @@ object (self)
     List.iter
       (fun (_,m) -> Generator.add_metadata abg m)
       (AFrame.get_all_metadata databuf);
-    self#slave_tick
+    if rate <> 1. then
```
smimram: Same here...

src/operators/soundtouch_op.ml
toots: Ha. Then we can't do this, as the `rate` might change after the clock is set..

smimram: Yes, let's just return `false` after all :)

src/outputs/pipe_output.ml
toots: I added the option. We should do it for `input.external` ideally one day but this is too much change for now.

src/operators/soundtouch_op.ml
```diff
@@ -28,21 +28,21 @@ class soundtouch ~kind (source:source) rate tempo pitch =
   let abg = Generator.create `Audio in
   let channels = (Frame.type_of_kind kind).Frame.audio in
 object (self)
-  inherit operator ~name:"soundtouch" kind [source]
+  inherit operator ~name:"soundtouch" kind [source] as super
```
toots:
```suggestion
  inherit operator ~name:"soundtouch" kind [source]
```

PR #1013, opened 2019-10-30
Native implementation of request queues.

smimram:


libs/utils.liq
```diff
@@ -513,6 +502,16 @@ def stereo.right(s)
   mean(stereo.pan(pan=1., s))
 end
 
+# Append speech-synthesized tracks reading the metadata.
+# @category Source / Track Processing
+# @param ~pattern Pattern to use
+# @param s The source to use
+def say_metadata
+  p = 'say:$(if $(artist),"It was $(artist)$(if $(title),\\", $(title)\\").")'
```
toots: We need to remove those interpolation patterns one day..

src/sources/request_simple.ml
```diff
@@ -88,17 +88,25 @@ let () =
        let r = Lang.to_request (List.assoc "" p) in
        ((new unqueued ~kind r):>source))
 
-class dynamic ~kind (f:Lang.value) length default_duration timeout conservative = object (self)
+class dynamic ~kind ~available (f:Lang.value) length default_duration timeout conservative = object (self)
```
toots: 👍

src/sources/request_simple.ml
```diff
@@ -88,17 +88,25 @@ let () =
...
   inherit
     Request_source.queued ~kind ~name:"request.dynamic"
-      ~length ~default_duration ~timeout ~conservative ()
+      ~length ~default_duration ~timeout ~conservative () as super
+
+  method is_ready =
+    let ready = super#is_ready in
+    if not ready && available () then super#notify_new_request;
```
toots: Why is `notify_new_request` called unconditionally? I'm not too familiar with the code but it looks like this should only be called at specific times. The comment about not calling it during shutdown points to an important aspect: `notify_new_request` wakes up a task, calling it at the wrong time can lead to deadlock during the shutdown phases. What condition are you trying to maintain here?

libs/utils.liq
smimram: Yes...

src/sources/request_simple.ml
smimram: Suppose that we set `available` to `false` at some point (we currently don't have any request). Then the next request will be `None`, which makes the queue stop looking for new requests. When `available` becomes true again, we notify the queue that it should do this again. So, it is not really continuously notifying, but only when it is `not ready` (i.e. we had stopped feeding the queue and it got empty), and it is `available` (there are requests again), which should not be too often. I don't see an easy way out while keeping the logic simple.

toots: Makes sense.

PR #1014, opened 2019-10-31
Add a synth protocol.

smimram:
This should be much better for tests than threesines for instance.

You can test with
```
out(single("synth:shape=sine,duration=10."))
```
It's not working at the moment.


smimram:
It now working!

PR #1015, opened 2019-11-01
Native implementation of playlists.

smimram:
It is only missing some testing.

Also, do we really need to keep `playlist.safe` (which is not so easy to do in plain Liq).


toots:
> Also, do we really need to keep `playlist.safe` (which is not so easy to do in plain Liq).

I'm in favor of removing it.

libs/deprecations.liq
```diff
@@ -6,7 +6,8 @@
 # @param old Old function name.
 # @param new New function name.
 def deprecated(old, new)
-  log.severe(label="lang.deprecated", "WARNING: \"#{old}\" is deprecated and will be removed in future version. Please use \"#{new}\" instead.")
+  new = if new == "" then "" else " Please use \"#{new}\" instead." end
```
toots: nit: `string.cat` is nice way for doing this.. 🙂

libs/deprecations.liq
```diff
@@ -35,4 +36,11 @@ end
...
   deprecated("quote", "string.quote")
   string.quote(s)
-end
\ No newline at end of file
+end
+
+# Deprecated: use mksafe and playlist instead.
+# @flag hidden
+def playlist.safe(~id, ~mime_type, ~mode, ~on_track, ~prefix, ~reload, ~reload_mode, uri)
+  deprecated("playlist.safe", "")
```
toots: The deprecation message is misleading here. *de facto* `playlist.safe` has already been removed..

CHANGES.md
```diff
@@ -25,6 +25,12 @@ Changed:
 - Changed `request.queue` into a Liquidsoap implementation (#1013).
 - Removed `request.equeue`, such a feature could be re-implemented in
   Liquidsoap, see `request.queue`.
+- The `playlist` operator is now fully implemented in Liquidsoap (#1015).
+- Removed `playlist.once`, it can be achieved by passing `"once"` to the
+  `reload_mode` of `playlist.once` (#1015).
```
toots: Some is off in this sentence.. 😅

libs/playlist.liq
```diff
@@ -100,45 +109,47 @@ def playlist.reloadable(~id="",~random=false,~on_done={()},
...
 # @param uri Playlist URI.
-def playlist.merge(~id="",~random=false,
-                   ~default_duration=30.,~length=10.,~conservative=false,uri) =
-  pl = playlist.reloadable(id=id,random=random,
-                           conservative=conservative,default_duration=default_duration,
-                           uri)
-  let (reload, s) = pl
-  s = merge_tracks(s)
-  on_end(delay=0.,fun(_,_)->reload(),s)
+def playlist(~id="", ~check_next=fun(_)->true, ~conservative=false, ~default_duration=30., ~length=10., ~mime_type="", ~mode="randomize", ~prefix="", ~reload=0, ~reload_mode="seconds", ~timeout=20., uri)
```
toots: Let's merge `playlist.reloadable` and `playlist` when we add the new commands!

libs/deprecations.liq
smimram: Not sure I understand this one...

libs/deprecations.liq
smimram: I don't understand this either: _deprecated_ means that the function should not be used anymore, but we are still defining it (precisely with this definition).

libs/playlist.liq
smimram: Yeah, that was my idea, but let's take it slowly.

libs/deprecations.liq
toots:
```ruby
tail = if new == "" then [] else ["Please use \"#{new}\" instead."]

msg = list.add(
  "WARNING: \"#{old}\" is deprecated and will be removed in future version.",
  tail
)

log.severe(label="lang.deprecated", string.concat(separator=" ", msg))
```

libs/deprecations.liq
toots: Usually, deprecation messages means that the function is still here but may be removed. But in this case, it's here but not doing what the old function was doing so this will be confusing and scripts using it will break or behave differently.

libs/playlist.liq
toots: 👍

libs/deprecations.liq
smimram: Is this really more clear?? :)

libs/deprecations.liq
smimram: Well it's more or less an equivalent implementation...

PR #1017, opened 2019-11-01
Use a pager to display long help results.

smimram:
This means that `less` is automatically invoked when typing
```
liquidsoap --list-plugins
```


toots:
I think that it's a great feature but I'm a little concerned about introducing the whole thing with an external binary. Are there any built-in alternatives?

smimram:
I don't know of any builtin alternative, and this is the way most programs (such as opam) are doing, [see  here for instance](https://github.com/dbuenzli/cmdliner/blob/master/src/cmdliner_manpage.ml#L443).

src/tools/utils.ml
```diff
@@ -730,3 +730,51 @@ let string_of_size n =
     Printf.sprintf "%.02f MiB" (float_of_int n /. float_of_int (1 lsl 20))
   else
     Printf.sprintf "%.02f GiB" (float_of_int n /. float_of_int (1 lsl 30))
+
+(* From opam. *)
```
toots:
```suggestion
(* From dbuenzli/cmdliner *)
```

toots:
Ok, I've factored out the print logic, added support for both `more` and `less` and added it to `--help` and `-h <operator>`. Merging, thanks for working on this!

PR #1019, opened 2019-11-01
Cleanup thread functions.

smimram:
Better naming (I could never remember `add_timeout`...).


src/lang/builtins_thread.ml
```diff
@@ -0,0 +1,87 @@
...
+            considered to be fast. When set to `false` its priority will be \
+            lowered below that of request resolutions and fast timeouts. This is \
+            only effective if you set a dedicated queue for fast tasks, see the \
+            \"scheduler\" settings for more details." ;
+      "delay", Lang.float_t, Some (Lang.float 0.),
+      Some "Delay (in sec.) after which the thread should be lauched.";
+      "", Lang.fun_t [] Lang.unit_t, None, Some "Function to execute."
+    ]
+    Lang.unit_t
+    ~descr:"Run a function in a separate thread."
```
toots: This is not true. `duppy` runs a pool of threads who pick and execute tasks similarly to the way `node` is implemented. There isn't a single thread per tasks, which was the case before and the reason we introduced duppy. I believe that using `tasks` (or `task`) instead of `thread` would be more appropriate.

libs/thread.liq
```diff
@@ -0,0 +1,36 @@
...
+  def f ()
+    delay = f ()
+    if delay >= 0. then thread.run.recurrent(fast=fast, delay=delay, f) end
+  end
+  thread.run(fast=fast, delay=delay, f)
+end
+
+# Run a recurrent function in a separate thread.
+# @category Control
+# @param ~fast Whether the thread is supposed to quickly or not. Typically, blocking tasks (e.g. fetching data over the internet) should not be considered to be fast. When set to `false` its priority will be lowered below that of request resolutions and fast timeouts. This is only effective if you set a dedicated queue for fast tasks, see the "scheduler" settings for more details.
```
toots:
```suggestion
# @param ~fast Whether the task is supposed to return quickly or not. Typically, blocking tasks (e.g. fetching data over the internet) should not be considered to be fast. When set to `false` its priority will be lowered below that of request resolutions and fast timeouts. This is only effective if you set a dedicated queue for fast tasks, see the "scheduler" settings for more details.
```

libs/thread.liq
```diff
@@ -0,0 +1,36 @@
...
+# now, see #1018.
+def rec thread.run.recurrent(~fast, ~delay, f)
+  def f ()
+    delay = f ()
+    if delay >= 0. then thread.run.recurrent(fast=fast, delay=delay, f) end
+  end
+  thread.run(fast=fast, delay=delay, f)
+end
+
+# Run a recurrent function in a separate thread.
```
toots:
```suggestion
# Run a recurrent task in a separate thread.
```

libs/thread.liq
```diff
@@ -0,0 +1,36 @@
...
+# @category Control
+# @param ~fast Whether the thread is supposed to quickly or not. Typically, blocking tasks (e.g. fetching data over the internet) should not be considered to be fast. When set to `false` its priority will be lowered below that of request resolutions and fast timeouts. This is only effective if you set a dedicated queue for fast tasks, see the "scheduler" settings for more details.
+# @param ~delay Delay (in sec.) after which the thread should be lauched.
+# @param f Function to execute.
+def thread.run.recurrent(~fast=true, ~delay=0., f)
```
toots: Is there a semantics difference between `run` and `run.recurrent`? This doesn't seem clear to me.

src/lang/builtins_thread.ml
```diff
@@ -0,0 +1,87 @@
...
+           events   = [`Delay delay] ;
+           handler  = fun _ -> Lang.to_unit (Lang.apply ~t:Lang.unit_t f []); []
+         }
+       in
+       Duppy.Task.add Tutils.scheduler task;
+       Lang.unit)
+
+let () =
+  let t = Lang.univ_t 1 in
+  add_builtin "mutexify" ~cat:Liq
```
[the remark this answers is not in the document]
smimram: Good idea, done.

src/lang/builtins_thread.ml
smimram: I do agree with you, but to me this is only an implementation detail: the average user wants to know that there are something like threads in Liq, and "task" is not very clear nor common...

libs/thread.liq
smimram: Not at all, it's just that the function, instead of returning unit, returns a float indicating when it should be run again (this was the behavior of the old `add_timeout` and is sometimes useful. What is not clear in the doc?

smimram:
> I like the idea, you're right those names weren't clear. I don't like the use of `thread` as it is not technically accurate. I am aware that `lwt` also uses the thread terminology to describe theirs (which are not system thread) but, considering our user base, I suggest that we use `task(s)` instead.

The terminology "task" is not really clear nor standard. I'd rather have our users believe that these are threads, which is a good approximation of reality.

> Also, the original semantics was of a function returning a `float` to decide wether or not to re-queue a task. I understand that this was annoying to have to add to each call but it should at least be allowed in some kind of way. Typical use for it are exponential back-off and I can definitely see this being useful.

This is precisely what `thread.run.recurrent` does!

libs/thread.liq
toots: 👍

Issue #1026, opened 2019-11-05
Performances issues with scripted lists


toots:
Scripted lists have terrible performances. Doing a regular `list.append` on a `100` long list brings the CPU on its knees. A quick look at the implementation show that it a using non-tail recursive implementation:
```
def rec list.append(l, m)
  list.case(l, m, fun (x, l) -> list.cons(x, list.append(l, m)))
end
```
Which leads to a gigantic call stack.

Lot of similar functions in `list.liq` have the same problem.

toots:
This is blocking #1000 from being merged as the performances of the implementation are severely impacted.

smimram:
Here is my test function:
```
height = 100
width  = 100

print("Started")
ll = list.init(height, fun (_) -> list.init(width, fun(i) -> i))
print("Made list")

i = ref(0)
def f(l)
  l = list.append(l, l)
  print("#{!i}: #{list.length(l)}")
  i := !i + 1
end
ll = list.map(f, ll)
print("Appended")
exit(0)

ignore(ll)
output.dummy(blank())
```
I just added `list.ind` which should be tail-rec. With the following implementation of append
```
def list.append(l, m)
  list.ind(l, m, fun (x, l, r) -> list.cons(x, r))
end
```
we go down from 1.5s to 1.2s. I guess we don't have any other choice than hardcoding back the most used functions...

smimram:
Ok so for 1000 times a list of length 100 appended with itself
- native OCaml implementation : 7.26s
- `list.ind` implementation : 7.59s
- recursive implementation : 11.5s

smimram:
For reference, the code for `list.append`:
```
let () =
  let a = Lang.univ_t 1 in
  Lang.add_builtin "list.append"
    ~category:(string_of_category List)
    ~descr:"Concatenate two lists."
    [
      "",Lang.list_t a,None,None;
      "",Lang.list_t a,None,None;
    ]
    (Lang.list_t a)
    (fun p t ->
       let t = Lang.of_list_t t in
       let l1 = Lang.to_list (Lang.assoc "" 1 p) in
       let l2 = Lang.to_list (Lang.assoc "" 2 p) in
       Lang.list ~t (l1@l2)
    )
```

smimram:
We now down to 2.77 on the same example, it turns out that `list.length` was quite slow too (thanks profiler).

smimram:
@toots I am waiting for you to test #1000, seems ok to me.

[...]

PR #1027, opened 2019-11-05
Profiler for scripts

smimram:


toots:
Why do I see negative values in https://github.com/savonet/liquidsoap/issues/1026#issuecomment-550377510 ? Also, it would be nice if the profiling stats could be printed separately from the main logs, right now they are dumped in the middle of them..

smimram:
> Why do I see negative values in [#1026 (comment)](https://github.com/savonet/liquidsoap/issues/1026#issuecomment-550377510) ? Also, it would be nice if the profiling stats could be printed separately from the main logs, right now they are dumped in the middle of them..

The negative values are likely due to strange callbacks (its supposed to work correctly only for quite sequential programs). You can print the stats from liq whenever you want.

toots:
It looks great! Could you add some doc before we merge it?

PR #1036, opened 2019-11-13
Add a protocol to convert a file to stereo.

smimram:
@toots: could you help me make this work?


toots:
Is it related to our discussion about changing the default behavior for stereo/mono conversion? If so, keep in mind that it'll be better done at the source level -- most of the issues we have are from DJs connecting to a live source in mono..

smimram:
It's to get rid of sox in a first time...

smimram:
and I have other uses for such programming approaches

toots:
Ok.

Protocols are designed to be chained processing steps applied during request preparation. They usually involve a external program of some sort. I'm not sure that what you are trying to do here fits this use-case..

smimram:
I managed to make this work! Isn't it nice?

toots:
I'm glad it works but please hold off on merging this one until we've talked about it more. Pretty sure the file is being generated while at the same time being sent back for ingestion into the system, that may works but is not a desirable behavior.

toots:
If you want to use a source for it, it needs to be synchronous. I would use a `sync="none"` clock and wait for the end of track before returning the file and I'm afraid this is not currently possible since all the event handlers are asynchronous callbacks.

smimram:
Ok let's start a fight :)

toots:
> Ok let's start a fight :)

It's not a fight, it's the difference between `"it works"` and `"it works and I'm confident that it'll behave the right way"`

I like this article on this topic: https://medium.com/better-programming/the-differences-between-a-junior-mid-level-and-senior-developer-bb2cb2eb000d
> The mindset of a junior developer often is to just make the code work. Working software and good software are considered the same.

smimram:
I do agree with you, but here the idea is to replace the use of `sox` which might not be installed or without the right packages, by Liquidsoap itself which we know is installed because we are currently running it. It's a strict improvement over the current situation and is improving the protocols which are quite hacky already.

That being said, I agree that we really don't want `source.dump` being used in wide public...

Are you really sure that it is not going to work correctly with the required adjusting of clocks and everything? This is quite useful as a paradigm sometimes... For instance, it is also used in #1014.

toots:
I'm glad you mentioned this, I did not realize it was already used as a pattern. Yes, it is a bad pattern. Remember that request based sources estimate a file's duration and start playing them right away. Add to that for instance a crossfade or any operator that consumes faster than real time and you can see what can happen: the source reaching the end of the file before it has been fully generated.

toots:
I disagree that simple improvement over existing patterns are enough. If that introduces new unexpected side effects and bugs, the quicker and the more we rush to merge these changes, the harder it becomes to understand the new bugs when the come in (combinational complexity) and overall the quality of the software is degraded...

smimram:
Do you see a (simple) way of fixing these potential bugs? What do you suggest? We can easily remove those features and move them to a PR if you insist on it.

smimram:
I've done this, all the `source.dump` code is now in this branch, we can discuss calmly now.

smimram:
@toots can you reconsider, now that you know that this is not asynchronous? :)

toots:
I'll look into it as soon as possible!

libs/protocols.liq
```diff
@@ -516,19 +526,10 @@ add_protocol(static=true,"gtts",gtts_protocol,
...
 # @flag hidden
 def say_protocol(~rlog,~maxtime,arg) =
-  sox = get(default="sox", "protocol.say.sox_path")
-  if file.which(get(default="", "protocol.text2wave.path")) != "" then
-    [process_uri(extname="wav",uri="text2wave:#{arg}","#{sox} $(input) -c 2 $(output)")]
-  else
-    log.info(label="say_protocol", "text2wave binary not found, trying gtts.")
-    [process_uri(extname="wav",uri="gtts:#{arg}","#{sox} $(input) -c 2 $(output)")]
-  end
+  ["stereo:gtts:#{arg}", "stereo:text2wave:#{arg}"]
```
toots: Why two files?

src/lang/builtins_source.ml
```diff
@@ -96,3 +96,43 @@ let () =
...
+       in
+       let kind = Encoder.kind_of_format format in
+       let file = Lang.to_string (Lang.assoc "" 2 p) in
+       let s = Lang.to_source (Lang.assoc "" 3 p) in
+       let encoder = encoder "source.dump" Meta_format.empty_metadata in
+       let frame = Frame.create kind in
+       let oc = open_out file in
+       Strings.iter (output_substring oc) encoder.Encoder.header;
+       s#get_ready [];
+       while s#is_ready do
```
toots: Have you thought about using track markers instead? It might be a more natural way of doing it, avoiding the use of `once` every time?

src/lang/builtins_source.ml
```diff
@@ -96,3 +96,43 @@ let () =
...
+         Frame.clear frame;
+         let start = Frame.position frame in
+         s#get frame;
+         let stop = Frame.position frame in
+         let data = encoder.Encoder.encode frame start (stop-start) in
+         Strings.iter (output_substring oc) data
+       done;
+       Strings.iter (output_substring oc) (encoder.Encoder.stop ());
```
toots: You need to cleanup the source as well. I'd suggest the use of `output.file`, which has more advanced features (in particular in the coming `ffmpeg-video` branch, will be able to use encoder's file output capabilities to encode formats with a header such as `mp4`). 

You should build the whole `output.file(%format,"file",source)` in the code then follow the expected invokations:
- `file_output#get_ready [(file_output :> Source.source)]`
- `file_output#output_get_ready`
- `file_output#output` (many times)
- `file_output#leave (file_output :> Source.source)`

You can look at `clock.ml` for details about those calls.

I'm still unclear as to wether we will need to handle the clock situation in this case.

libs/protocols.liq
smimram: Because the first one might fail (if `gtts` is not installed) in which case it should default to the second.

src/lang/builtins_source.ml
smimram: Not really sure about how to get track markers, any hint? :)

src/lang/builtins_source.ml
```diff
@@ -109,3 +109,42 @@ let () =
...
+        try Encoder.get_factory format
+        with Not_found ->
+          raise (Lang_errors.Invalid_value (format_val, "Unsupported format"))
+      in
+      let kind = Encoder.kind_of_format format in
+      let file = Lang.to_string (Lang.assoc "" 2 p) in
+      let s = Lang.to_source (Lang.assoc "" 3 p) in
+      let encoder = encoder "source.dump" Meta_format.empty_metadata in
+      let frame = Frame.create kind in
+      let oc = open_out file in
```
toots, line 137: Some encoders (currently only `%ffmpeg`) can handle the file outputs themselves. It's important for formats such as `mp4` that need a final header after encoding is finished.

See here: https://github.com/savonet/liquidsoap/blob/master/src/outputs/pipe_output.ml#L384 and here: https://github.com/savonet/liquidsoap/blob/master/src/outputs/pipe_output.ml#L332

smimram: I am using the same code as `output.file` now. Are you alright with it?

src/lang/builtins_source.ml
```diff
@@ -109,3 +109,35 @@ let () =
...
+      let fo = Pipe_output.new_file_output p in
+      fo#get_ready [s];
+      fo#output_get_ready;
+      log#info "Start dumping source.";
+      while s#is_ready do
+        fo#output;
+        fo#after_output
+      done;
+      log#info "Source dumped.";
+      fo#leave s;
```
toots, line 142: Would you mind factoring this out to a `Utils`? I think this should prove pretty useful in other situations and also would make it possible to fix all cases using this trick if we later realize something needs to be changed with this flow. Thks!

toots: Actually, now that I'm thinking about it, how about surfacing this as a proper operator and using it instead?

`source.synchronous` -> process the whole source synchronously as in this code, and then, instead of `source.dum`, do: `source.synchronous(output.file(...))`..

smimram: The thing is that the input source has to be an output. Can we enforce this in the type?

toots: Ha, good point. No, I don't think that we have that in the type. `active_source` is used to denote sources that get actively pulled but that is not specific to outputs.

I guess we should merge as is with the workflow factored out as a `Utils` and revisit later if we see a more general need for this in the future. Now that I think about it, some (most?) outputs would not make sense with `source.synchronous`, like `output.icecast` etc.

smimram:
I'm merging, we'll do the factor out when needed.

Issue #1035, opened 2019-11-13
Type inference issue.


toots:
I ran into a very odd issue working on the reporting tools. The function that causes it has this type:
```
Monitor a source's internals on Prometheus

Type: (?window : float, labels : [string]) -> (label_values : [string], source('a)) -> unit

Category: Liquidsoap

Parameters:

 * window : float (default: 5.0)
     Window over which mean and peak metrics are reported.

 * labels : [string] (default: None)
     labels for the metric
```
Due to the prometheus API, the first call returns a the master metric factory, which is used to instantiate source-specific metrics, with the set of required labels attached, hence this type.

Now, here's the script that triggers the error:
```
s = blank()

monitor_metric = prometheus.monitor(labels=["foo"])

monitor_metric(label_values=["bar"],s)

print(source.is_ready(s))

output.dummy(fallible=true,s)
```
If you enable lang debugging (for instance set `LIQUIDSOAP_DEBUG_LANG=1` in the environment), you will see this:
```
Runtime instantiation of source.is_ready: (source(audio=2, video=0, midi=0)) -> bool becomes (source(audio=2, video=0, midi=0)) -> bool.
Runtime instantiation of print: (?newline : bool, audio=2, video=0, midi=0) -> unit targets (?newline : bool, bool) -> unit.
(?newline : bool, audio=2, video=0, midi=0) -> unit <: (?newline : bool, bool) -> unit
bool <: bool
bool <: audio=2, video=0, midi=0
At unknown position:
Error 5: this value has type
  (?newline : _, ,..., ...) -> _
but it should be a subtype of the type of the value at line 12, char 0-6
  (?newline : _, bool, ...) -> _
```
(full [log](https://github.com/savonet/liquidsoap/files/3838781/log.txt))

It seems to me that one universal quantifier is escaping its scope and is being unified with inconsistent types.. Here's the patch that I have applied to make it work:
```
diff --git a/src/lang/builtins_prometheus.ml b/src/lang/builtins_prometheus.ml
index 7deeca36..63a46d06 100644
--- a/src/lang/builtins_prometheus.ml
+++ b/src/lang/builtins_prometheus.ml
@@ -221,11 +221,10 @@ let source_monitor ~label_names ~labels ~window s =
   s#add_watcher watcher

 let () =
-  let kind_t = Lang.univ_t 1 in
   let source_monitor_register_t =
     Lang.fun_t [
       false,"label_values",Lang.list_t Lang.string_t;
[...]

PR #1037, opened 2019-11-14
Get rid of universal variable indices.

smimram:
It might help clarifying #1035.


toots:
Oh my. Gonna have to catch up on the typing stuff...

Question: how do you account for things like: `list.hd(default='a, ['a]) -> 'a` where `'a` is the same on all 3 spots?

smimram:
Like this: https://github.com/savonet/liquidsoap/pull/1037/files#diff-0c0cdd98d034cac51af8683f71f93cafL96

You declare your type variable before with a `let` and use it three times.

toots:
> Like this: https://github.com/savonet/liquidsoap/pull/1037/files#diff-0c0cdd98d034cac51af8683f71f93cafL96
> 
> You declare your type variable before with a `let` and use it three times.

Awesome. This looks great, give me a second to review it and let's merge soon. One question we should answer before though: why were those made to be cached in the first place?

smimram:
My guess is that they were really numbered first and then caching was introduced in order not to change the API. At least, I understand the old comment in this way.

src/conversions/audio_to_stereo.ml
```diff
@@ -68,9 +68,7 @@ object
 end
 
 let () =
-  let input_kind =
-    Lang.kind_type_of_kind_format ~fresh:1 Lang.audio_variable
-  in
+  let input_kind = Lang.kind_type_of_kind_format Lang.audio_variable in
```
toots:
```suggestion
  let input_kind =
    Lang.kind_type_of_kind_format Lang.audio_variable
  in
```

src/lang/builtins_list.ml
```diff
@@ -123,12 +124,13 @@ let () =
...
 let () =
+  let a = Lang.univ_t () in
   add_builtin "list.sort" ~cat:List
     ~descr:"Sort a list according to a comparison function."
     ["",
-     Lang.fun_t [false,"",Lang.univ_t 1;false,"",Lang.univ_t 1] Lang.int_t,
+     Lang.fun_t [false,"",a;false,"",a] Lang.int_t,
      None, None ;
-     "",Lang.list_t (Lang.univ_t 1),None,None] (Lang.list_t (Lang.univ_t 1))
+     "",Lang.list_t a,None,None] (Lang.list_t a)
```
toots: 🎉

src/lang/builtins_time.ml
```diff
@@ -48,9 +48,10 @@ let () =
 let () =
   add_builtin ~cat:Sys "time" ~descr:"Return the current time since 00:00:00 GMT, Jan. 1, 1970, in seconds."
     [] Lang.float_t (fun _ ->
-      Lang.float (Unix.gettimeofday ()));
+        Lang.float (Unix.gettimeofday ()));
+  let univ = Lang.univ_t () in
```
toots:
```suggestion
  let t = Lang.univ_t () in
```

src/lang/builtins_time.ml
```diff
@@ -48,9 +48,10 @@ let () =
...
   let execute cb tm =
-    Lang.apply cb ~t:(Lang.univ_t 1)
+    Lang.apply cb ~t:univ
```
toots:
```suggestion
    Lang.apply cb ~t
```

src/lang/lang.ml
```diff
@@ -63,30 +63,11 @@ let variable_t = Term.variable_t
...
-          v
-let string_getter_t n = univ_t ~constraints:[T.Getter T.String] n
-let float_getter_t n = univ_t ~constraints:[T.Getter T.Float] n
-let int_getter_t n = univ_t ~constraints:[T.Getter T.Int] n
-let bool_getter_t n = univ_t ~constraints:[T.Getter T.Bool] n
+let univ_t ?(constraints=[]) () = T.fresh ~level:0 ~constraints ~pos:None
+let string_getter_t () = univ_t ~constraints:[T.Getter T.String] ()
+let float_getter_t () = univ_t ~constraints:[T.Getter T.Float] ()
+let int_getter_t () = univ_t ~constraints:[T.Getter T.Int] ()
+let bool_getter_t () = univ_t ~constraints:[T.Getter T.Bool] ()
```
toots: giphy

PR #1044, opened 2019-11-30
Compute kinds by unification in order to get rid of types during evaluation.

smimram:


smimram:
This has been rebased (manually) to master in #1202.

PR #1046, opened 2019-12-02
Validate setting before changing its value

CyberDomovoy:
Add a new parameter to register: validate (fun(v) -> bool)
It register a function to be called before changing the value of a setting (when using set), if the function returns false, an error is raised.

In order for this to work, ocaml-dtools need to be changed, i'm sending a pull request for it right after this one.

Please tell me if there is anything i could have done better, i'm new to ocaml.

### DIFF
--- a/src/lang/builtins_dtools.ml
+++ b/src/lang/builtins_dtools.ml
@@ -96,11 +96,7 @@ let () =
 
 let () =
   let set cast path v =
-    try
-      (cast (Configure.conf#path (Dtools.Conf.path_of_string path)))#set v
-    with
-      | Dtools.Conf.Unbound (_, _) ->
-          log#severe "WARNING: there is no configuration key named %S!" path
+    (cast (Configure.conf#path (Dtools.Conf.path_of_string path)))#set v
   in
     add_builtin ~cat:Liq "set"
       ~descr:"Change some setting. \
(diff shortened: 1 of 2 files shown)


CyberDomovoy:
Use case:
```ocaml
register(
  name="Test",
  validate=fun(v) -> v==2,
  "rlk.test",
  1
)

set("rlk.test", 2)
set("rlk.test", 3)
```

toots:
Hi! Looks like a nice idea! Since we don't have a support for exceptions yet, I suppose the best option is to not set the new value when validation fails. The validation function could also be used as a callback when a set operation has failed b/c of validation errors.

CyberDomovoy:
> Since we don't have a support for exceptions yet, I suppose the best option is to not set the new value when validation fails.

Well, i guess that depends, this PR makes liquidsoap fail by telling you the value is wrong. For me it is what i want: if my script sets a wrong value for a setting, then i should correct it, instead of "hoping for the best" by letting it use a value i don't expect.

But sure, if people prefer, it could just print 'log(level=1)' and keep going.  In that case we should have a way of knowing if 'set' did change the value.
Changing 'set' so that it returns bool or the value (old one if failed) would break backward compatibility, so it's a no-go.
I could 'get' to compare it to the value i wanted, but it seems redundant to me.
Using 'validate' itself could work, but what of settings registered without one? How could we know if the value was set correctly?

> The validation function could also be used as a callback when a set operation has failed b/c of validation errors.

What do you mean by validation errors? Isn't that already happening by Dtools raising Dtools.Conf.Mismatch?

toots:
> What do you mean by validation errors? Isn't that already happening by Dtools raising Dtools.Conf.Mismatch?

Yes, it is but since liquidsoap script for now does not have any way to catch errors, it's usually a good practice to avoid raising exceptions inside the script, except during the initial evaluation phase before streaming starts. In theory, settings can be changed after that phase.

I suppose we should raise an exception before streaming has started and log an error afterward for now and maybe chance that when/once we have a proper support for exceptions.

CyberDomovoy:
I see, it makes sense, when streaming started, everything should be done so that it doesn't stop.
I'll take a look at that. Is there an easy way to know if streaming started?

toots:
> I see, it makes sense, when streaming started, everything should be done so that it doesn't stop.
> I'll take a look at that. Is there an easy way to know if streaming started?

I think that `Tutils.has_started ()` would work fine for this.

CyberDomovoy:
Something like this?

src/lang/builtins_dtools.ml
```diff
@@ -104,42 +108,71 @@ let () =
...
+              in
+              let kind =
+                match t#kind with
+                  | Some "unit" -> "of type unit"
+                  | Some "int" -> "of type int"
+                  | Some "float" -> "of type float"
+                  | Some "bool" -> "of type bool"
+                  | Some "string" -> "of type string"
+                  | Some "list" -> "of type [string]"
+                  | _ -> "untyped"
```
toots: I think that you can put a `assert false` here:
```suggestion
                  | _ -> assert false
```

CyberDomovoy:
Here we are.
Just don't forget: ocaml-dtools needs to be merged as well.

toots:
Looks great, thanks! Please make sure to disable code formatting in the future to focus the actual changes. Merging!

CyberDomovoy:
Yeah i'm fighting against my editor (geany) for code formatting.
What editor would you recommend? (i'm running debian linux)
If possible, i'd like to avoid big machines like eclipse.

smimram:
Old people like us use emacs or vim, and preferably the first :) The support for ocaml brought by tuareg and merlin is really wonderful there.

PR #1049, opened 2019-12-11
Everything is a URI, isn't it?

CyberDomovoy:
I created this PR following a discussion with @toots on slack.

Goal of this PR:
Rethink how requests and decoders get to/from a resource.
In the end, any resource can be be identified by a URI (that's what URIs are for).

A request resolves successfully if: either it finds a decoder, or the request kind is 'None' and we resolve to a uri of type 'file:<path>', <path> being a valid, readable local file.

This introduce a new decoder: uri_decoder, its goal would be to to "unify" every other decoder. It takes a uri, and decode it in any way it wants (buffer, stream, file... whatever)
It would avoid the need to create sources like 'input.http', and just use the "classic" sources with a uri.


CyberDomovoy:
How i tested this:

```
set("init.force_start", true)

uri = "/srv/liquidsoap/stations/ondecourte/media/jingles/16-jingle_tunning.ogg"
r = request.create(uri)
if request.resolve(r) then
  resolved = request.filename(r)
  log(level=1, label="request-uri", "Resolved to: #{resolved}")
end
request.destroy(r)
r = request.create.raw(uri)
if request.resolve(r) then
  resolved = request.filename(r)
  log(level=1, label="request-uri", "Resolved to: #{resolved}")
end
request.destroy(r)

uri = "https://john.doe@www.example.com:123/forum/questions/?tag=networking&order=newest#top"
r = request.create.raw(uri)
if request.resolve(r) then
  resolved = request.filename(r)
  log(level=1, label="request-uri", "Resolved to: #{resolved}")
end
request.destroy(r)

uri = "ldap://[2001:db8::7]/c=GB?objectClass?one"
r = request.create.raw(uri)
if request.resolve(r) then
  resolved = request.filename(r)
  log(level=1, label="request-uri", "Resolved to: #{resolved}")
end
request.destroy(r)

uri = "mailto:John.Doe@example.com"
r = request.create.raw(uri)
if request.resolve(r) then
  resolved = request.filename(r)
  log(level=1, label="request-uri", "Resolved to: #{resolved}")
end
request.destroy(r)

uri = "news:comp.infosystems.www.servers.unix"
r = request.create.raw(uri)
if request.resolve(r) then
  resolved = request.filename(r)
  log(level=1, label="request-uri", "Resolved to: #{resolved}")
end
request.destroy(r)

uri = "tel:+1-816-555-1212"
r = request.create.raw(uri)
if request.resolve(r) then
  resolved = request.filename(r)
  log(level=1, label="request-uri", "Resolved to: #{resolved}")
end
request.destroy(r)

uri = "telnet://192.0.2.16:80/"
r = request.create.raw(uri)
if request.resolve(r) then
  resolved = request.filename(r)
  log(level=1, label="request-uri", "Resolved to: #{resolved}")
end
request.destroy(r)

uri = "urn:oasis:names:specification:docbook:dtd:xml:4.1.2"
r = request.create.raw(uri)
if request.resolve(r) then
  resolved = request.filename(r)
  log(level=1, label="request-uri", "Resolved to: #{resolved}")
end
request.destroy(r)

shutdown()
```

CyberDomovoy:
About new uri_decoder:

Where should their modules reside? src/decoders, src/protocols ?

(the first one would be to decode 'file:' uris)

src/Makefile
```diff
@@ -224,7 +224,7 @@ builtins = lang/lang_builtins.ml \
 
 liquidsoap_sources= \
 	tools/log.ml $(tools) converters/audio_converter.ml $(stream) \
-	decoder/decoder.ml request.ml tools/server.ml \
+	URI.ml decoder/decoder.ml request.ml tools/server.ml \
```
toots: Have you considered using https://github.com/mirage/ocaml-uri ? Uri parsing can be tricky, using an existing, popular module makes sure some of the quirks have already been considered and fixes..

CyberDomovoy:
TODO:
encode/decode URIs

We can see the problem with ```uri="https://john.doe@www.example.com:123/forum/questions/?tag=networking&order=newest#top"```:

```Malformed URI: "process:osb,curl -sL \"https$(colon)//john.doe@www.example.com$(colon)123/forum/questions/?tag=networking&order=newest#top\" -o $(output)"```

CyberDomovoy:
> This feature should be preserved by your changes, making sure that `file:/path/to/file.mp3'` and `/path/to/file.mp3` both work to push file URIs.

It kind of does, when creating a URI without a scheme, it default to "file" (i believe this is sane, we _must_ have a default one anyway), and if the scheme is "file" the path is expanded using home_unrelate.

src/Makefile
CyberDomovoy: Nice, i'll take a look at that

CyberDomovoy: Done with 6085d8c

CyberDomovoy:
About metadatas:

As i understand it, they are set by:
Request.create, on the request itself
Request.mresolvers, on indicators

Do we need to differentiate metadata set on the request than the ones set on indicators?
Can't mresolvers just replace the request metadata?
In the end, when sending/querying metadata for a request, the result is the same, no?

CyberDomovoy:
About indicators:

Why a list of lists?
Wouldn't it work just as well if when pushing a list of indicators, each of those are just added on top of the list?

From what i can see, they are always pushed, popped, or peeked, so we either:
_ add a list on top of the list
_ OR access the first element of the first list in the list
A simple list of elements would work just fine for this, and be easier to manage.

CyberDomovoy:
persistent requests? why?

IMHO, when we ask to destroy a request, it should be destroyed. If something was still using it, then something is wrong with our code.
I think persistent requests are bad for system resources: it introduce the possibility to have a request in memory that is not used, and won't be used.

Is there any case where it makes sense to destroy a request and use it again?
Wouldn't it be better to have a way of "locking" a request? (telling it we need it, and hence nothing should be allowed to destroy it)

CyberDomovoy:
Protocols: argument static

Is it really needed? Can we even consider anything that doesn't resolve to a local file to be static?
From my experience, even this last one isn't always true (thanks guys for renaming/moving stuff on the ftp while they are part of the current scripted schedule). But we need some form of infallible source.

Or do we? What if we get rid of this need, and just have _one_ internal infallible source we use when everything else fails? (maybe simply some blank ?) Of course this would be severely logged.

toots:
Hi!

> Make a 'annotate' decoder which adds the metadata to a sub-request? Seems an overkill, but also the "right" way to do it with how i see requests/decoders...

I think that it's worth exploring for sure. We might even want to export metadata resolvers to the script side like we do for requests.

As for the indicators thing, I'm not sure exactly how it works. I think what's most important is to keep track of each resolved things along the resolution path, that makes it easier to debug when a request resolution does not return the expected result, being a file or metadata, or takes too long in one resolution stage etc.

>About indicators:
>
>Why a list of lists?
>Wouldn't it work just as well if when pushing a list of indicators, each of those are just added on top of the list?
>
>From what i can see, they are always pushed, popped, or peeked, so we either:
>_ add a list on top of the list
>_ OR access the first element of the first list in the list
>A simple list of elements would work just fine for this, and be easier to manage.
>
>Actually, why not use Stack?

Same thing, I'm not sure and, yes, the simpler the data structure the better. I'd argue against the use of `Stack` because it is opaque and will make it hard to inspect for debugging purposes. 

>persistent requests? why?
>
>IMHO, when we ask to destroy a request, it should be destroyed. If something was still using it, then something is wrong with our code.
>I think persistent requests are bad for system resources: it introduce the possibility to have a request in memory that is not used, and won't be used.
>
>Is there any case where it makes sense to destroy a request and use it again?
>Wouldn't it be better to have a way of "locking" a request? (telling it we need it, and hence nothing should be allowed to destroy it)

Yes, there is a case: liquidsoap is able to detect when a file-based source is static, for instance a top-level `single("/path/to/jingle.mp3")`. In this case, the request is resolved only once.

It matters because of resolution time. If, for instance, you have a `single("http://long-file.mp3")`, request resolution requires a whole download, which will stale the source for a long time.

Feel free to change the implementation details around this but this feature should be kept, for sure.

>Protocols: argument static
>
>Is it really needed? Can we even consider anything that doesn't resolve to a local file to be static?

There are two cases:
- A protocol which resolves to an existing file, that shouldn't be removed. Typically, a file from a playlist
- A protocol which resolves to a temporary file, which should be deleted when the request is cleaned up.

I believe that's what `static` is for. Again, feel free to rename/change the implementation details but the feature should certainly be kept.

>From my experience, even this last one isn't always true (thanks guys for renaming/moving stuff on the ftp while they are part of the current scripted schedule). But we need some form of infallible source.
>
>Or do we? What if we get rid of this need, and just have one internal infallible source we use when >everything else fails? (maybe simply some blank ?) Of course this would be severely logged.

I'm not sure what you are referring to here but the plan is to keep leaving the choice of how to make a source infallible up to the user.. 🙂

Hope this helps!

I've just pushed a big reformat of the code using `ocamlformat`. We decided to give it a try. You should make sure that your branch is merged or rebased against the latest commit before the reformat (`3cc0f682137db959101c964051afed1f5d3c7ab2`). After that, if you have trouble merging, I should be able to help.

Also, I'd suggest to cleanup the diff on changes unrelated to the things you're working on.

Bon courage, thanks for working on this!

CyberDomovoy:
> I'd argue against the use of `Stack` because it is opaque and will make it hard to inspect for debugging purposes.
My thinking was that maybe Stack has some optimizations that would make it more efficient than a List, but its behavior is easily reproducible using a List.

> Yes, there is a case: liquidsoap is able to detect when a file-based source is static, for instance a top-level `single("/path/to/jingle.mp3")`. In this case, the request is resolved only once.
> 
> It matters because of resolution time. If, for instance, you have a `single("http://long-file.mp3")`, request resolution requires a whole download, which will stale the source for a long time.
> 
> Feel free to change the implementation details around this but this feature should be kept, for sure.

I see, if i get you right, the idea is: a resolver tells us that it will _always_ resolve to the same result, so we only need to resolve once. If not, setup the source so that each time it needs the resource, it resolves it again.

> There are two cases:
> 
>     * A protocol which resolves to an existing file, that shouldn't be removed. Typically, a file from a playlist
> 
>     * A protocol which resolves to a temporary file, which should be deleted when the request is cleaned up.
> 
> 
> I believe that's what `static` is for. Again, feel free to rename/change the implementation details but the feature should certainly be kept.

Right, forgot about that.

> I'm not sure what you are referring to here but the plan is to keep leaving the choice of how to make a source infallible up to the user.. slightly_smiling_face

Simple case: my script has a `single("/some/local/file.mp3")`, LS considers it static, resolves only once, everything we just said. The thing is, people screw up, my script has been running for a week, but some guys goes to the ftp and think "hey, this should be filed under 'dub', not 'reggae'! Let's move it", bam! script fails because the (maybe only) "static" source is no longer available, we end up with an error where we initially thought there couldn't be one. What should we do? Well, if we don't consider _anything_ to be infallible to begin with, then we _do_ have a way to handle the case: we find another source, and if we don't have one we play some internally generated data (blank, beep, some raw audio for which the data is _compiled with the program data_...)

> Hope this helps!

A lot, thanks.

> I've just pushed a big reformat of the code using `ocamlformat`. We decided to give it a try. You should make sure that your branch is merged or rebased against the latest commit before the reformat (`3cc0f682137db959101c964051afed1f5d3c7ab2`). After that, if you have trouble merging, I should be able to help.

Will do that
 
> Also, I'd suggest to cleanup the diff on changes unrelated to the things you're working on.

Actually, i'm just keeping this PR for the sake of the discussion, i've started with a blank file for request.ml(i). And i've also started some other branches for changes unrelated to its rewrite.

CyberDomovoy:
> There are two cases:
> 
>     * A protocol which resolves to an existing file, that shouldn't be removed. Typically, a file from a playlist
> 
>     * A protocol which resolves to a temporary file, which should be deleted when the request is cleaned up.
> 
> 
> I believe that's what `static` is for. Again, feel free to rename/change the implementation details but the feature should certainly be kept.

Something is bugging me about this:
When i register a protocol, it will basically return a list of URIs (the case is a bit different when done from a script or from ocaml code, but lets stick to the simpler form: from script). This list will in turn be resolved using other protocols.
The thing is: the protocol i registered have no way of knowing _how_ these URIs will be resolved (resolver set from ocaml, maybe overwritten by the user registering its own protocol...).
So, how can i say "my protocol is static"? It might even end up _not_ being a file.
Worse: my protocol might create some temporary files, to be used by the protocols responsible of resolving the URIs i will return. Then, LS is not even _aware_ i have some files to cleanup.

Worse Example:
_ protocol "bar" exists, and resolves using the content of a file ("bar:some_file")
_ register protocol "foo"
_ resolve "foo:some_arg"
_ protocol "foo" creates "/tmp/infos_from_foo", and returns `["bar:/tmp/infos_from_foo"]`
_ protocol "bar" resolves to `["file:/tmp/file_from_bar"]`
_ protocol "file" ends the resolution, i find a decoder that can handle "file:/tmp/file_from_bar", i start decoding
_ The request gets destroyed
_ protocol "bar" is _not_ static, LS removes "/tmp/file_from_bar"
_ protocol "foo" is _not_ static, i'm not going to remove "bar:/tmp/infos_from_foo", it is not a local file, but "/tmp/infos_from_foo" is still there, and LS have no idea it should be removed since it is not returned by my protocol.

What i can think of:
registering a protocol is not a simple matter of giving it a uri and getting a list of other ones in return. The resolver may very well do much more than simply creating a file, and we have no way of predicting what.
It should be a set of:
_ a resolver (mandatory, same behavior as current one)
_ a cleanup function (optional, used to remove any temporary files it may have created, terminate connection to a remote service, whatever...)

Problem is: how to implement that, while keeping backward compatibility?

PR #1050, opened 2019-12-12
Add support for native timeval-based high-resolution time/latency.

toots:
This PR adds support for high-resolution timing API to be used to control latency during streaming.

The default implementation currently uses `Unix.gettimeofday`, which returns a `float`. Since a timestamp is a rather high number, this is guarantee to have bad resolution in the lower digits, which in turns will give approximative results to control latency.

In other word, the higher a `float` is, the less precise its lower digits are so a time interval `time2 - time1` will only be approximative.

Instead, by switching the the POSIX `tv_sec`/`tv_usec`, we can make exact computations of time intervals, giving us a much better latency control.


smimram:
Can you explain what's the use of this? Is it related to crossfade bugs?

toots:
> Can you explain what's the use of this? Is it related to crossfade bugs?

Just added a description!

smimram:
Very good idea, thanks for the explanation!

smimram:
I'm ok to review, but where can I find this sys-time library?

smimram:
Also, in parallel with the sys-time lilbrary I would propose a PR to OCaml so that it could get integrated in the Unix module on the long run.

toots:
Here: https://github.com/toots/ocaml-sys-time/ !

I'll think about the PR to OCaml, it's not a bad idea indeed. For now, the above is posix-only since the bound APIs do not exist as it on windows.

src/clock.ml
```diff
@@ -35,6 +35,22 @@ let log = Log.make ["clock"]
 let conf_clock =
   Dtools.Conf.void ~p:(Configure.conf#plug "clock") "Clock settings"
 
+module Time : Liq_time.T = (val !Liq_time.implementation)
```
smimram: Can't we simply use the configure / Makefile in order to choose the implementation of the module at compilation time, instead of having a this reference?

toots: I can see it either way. Let me think about it for a sec.. 🙂

toots:
> Apart from the useless module reference, it's all good to me!

Cool, thanks! The latest `posix-time` needs this PR merged: https://github.com/ocamllabs/ocaml-ctypes/pull/620. I'm a little concerned since the other last PR there is my 6 months ago with no comment at all..

toots:
Actually, `posix-time2` has just been published so it's time to revive this one!

src/clock.ml
toots: I think that this is fine this way. It's more explicit but also it should make it possible to later convert it to a settings if we feel like it.

src/clock.ml
```diff
@@ -213,25 +221,25 @@ class clock ?(sync = `Auto) id =
...
       let rec loop () =
         (* Stop running if there is no output. *)
         if outputs = [] then ()
         else (
           let self_sync = self#self_sync in
-          let rem = if self_sync then 0. else delay () in
+          let rem = if self_sync then time_zero else delay () in
           (* Sleep a while or worry about the latency *)
-          if self_sync || rem > 0. then (
+          if self_sync || time_zero |<| rem then (
```
toots, line 240: 😅

src/clock.ml
```diff
@@ -213,25 +221,25 @@ class clock ?(sync = `Auto) id =
...
             acc := 0;
-            if rem > 0. then usleep rem )
+            if time_zero |<| rem then sleep rem )
           else (
             incr acc;
             if rem < max_latency then (
```
toots, line 245: I think this is the guy that's causing #1251

PR #1058, opened 2019-12-15
Add support for vorbis covers.

smimram:


toots:
I think you might to have a look at: https://github.com/savonet/ocaml-ogg/blob/master/src/ogg_demuxer.mli

toots:
Also, we have a generic path to do that, via `request.resolve`/`request.metadata`. Is there a need for a vorbis/ogg specific tool?

smimram:
I'll have a look at the ogg demuxer. The reason why I want something custom is that usual tag readers such as taglib don't export cover art, which is what I am interested in right now.

smimram:
You were right, I can actually use the standard metadata for ogg!

smimram:
Are you ok now?

CHANGES.md
```diff
@@ -59,6 +58,8 @@ Changed:
 - Renamed `which` to `file.which`.
 - Change `blank()` duration semantics to mean forever only on negative values.
 - Get rid of numbering of universal variables (#1037).
+- Renamed `base64.decode`/`base64.encode` to
+  `string.base64.decode`/`string.base64.encode`.
```
toots: 👍

libs/metadata.liq
```diff
@@ -33,17 +33,50 @@ def metadata.float_getter(init, metadata, s)
...
 def file.cover(file)
-  m = file.mp3.tags(file)
-  m = list.filter(fun (lv) -> fst(lv) == "APIC", m)
-  m = list.map(snd, m)
-  if m == [] then "" else
-    # TODO: we could use file type in order to select cover if there are many
-    let (_,_,_,data) = file.mp3.parse_apic(list.hd(default="", m))
-    data
+  m = file.metadata(file)
+  print("**meta: #{m}")
```
toots: Remove?
```suggestion
```

libs/metadata.liq
```diff
@@ -33,17 +33,50 @@ def metadata.float_getter(init, metadata, s)
...
+  m = list.map (fun (lv) -> (string.case(lower=true, fst(lv)), snd(lv)), m)
+  if list.exists(fun (lv) -> fst(lv) == "coverart", m) then
```
toots:
```suggestion
  if list.mem_assoc("coverart", m) then
```

libs/metadata.liq
```diff
@@ -33,17 +33,50 @@ def metadata.float_getter(init, metadata, s)
...
+    # We are likely to have an ogg/vorbis file
+    m = list.filter(fun (lv) -> fst(lv) == "coverart", m)
+    m = list.map(snd, m)
+    m = list.hd(default="", m)
```
toots:
```suggestion
    m = list.assoc(default="", "coverart", m)
```
And remove the two lines above.

libs/metadata.liq
```diff
@@ -33,17 +33,50 @@ def metadata.float_getter(init, metadata, s)
...
+    if m == "" then "" else string.base64.decode(m) end
+  else
+    # Assume we have an mp3 file
+    m =
+      if list.exists(fun (lv) -> fst(lv) == "apic", m) then
```
toots:
```suggestion
      if list.mem_assoc("apic", m) then
```

libs/metadata.liq
```diff
@@ -33,17 +33,50 @@ def metadata.float_getter(init, metadata, s)
...
+      else
+        # Try the builting tag reader because APIC tags are not read by default,
+        # e.g. with taglib
+        file.mp3.metadata(file)
+      end
+    m = list.filter(fun (lv) -> fst(lv) == "apic", m)
+    m = list.map(snd, m)
+    if m == [] then
+      ""
+    else
```
toots:
```ruby
m = list.assoc(default="", "apic", m)
if m == "" then
```
And update below..

PR #1059, opened 2019-12-15
Add abstract type.

toots:
This PR adds support for opaque abstract types. I'm planning on using it to define the ffmpeg filter graph composition API, passing the config object as an abstract value.


smimram:
Could we have some concrete code / demo ?

toots:
> Could we have some concrete code / demo ?

For sure!

```ocaml
module Foo = Lang.MkAbstract (struct
  type content = string

  let name = "foo"
end)

module Bar = Lang.MkAbstract (struct
  type content = string

  let name = "bar"
end)

let () =
  Lang.add_builtin_base ~category:(string_of_category Sys) ~descr:"foo" "foo"
    (Foo.to_value "foo").Lang.value Foo.t;
  Lang.add_builtin_base ~category:(string_of_category Sys) ~descr:"bar" "bar"
    (Bar.to_value "bar").Lang.value Bar.t
```

```ruby
# foo;;
- : foo = <opaque>
# bar;;
- : bar = <opaque>
# def f() = foo end;;
f : () -> foo = <fun>
# def f(_=foo) = () end;;
f : (?foo) -> unit = <fun>
# f(bar);;
At line 5, char 2-5:
Error 5: this value has type
  bar
but it should be a subtype of the type of the value at line 4, char 6
  foo (inferred at line 4, char 8-11)
```

src/lang/lang_values.ml
```diff
@@ -639,6 +644,7 @@ let rec check ?(print_toplevel = false) ~level ~env e =
     e.t >: mk (T.Arrow (proto_t, body.t))
   in
   match e.term with
+    | Abstract (n, _) -> e.t >: mkg (T.Abstract n)
```
toots: This is the only part of these changes that I don't like. Ideally, abstract types should be decoupled from opaque values. Here are a couple of thoughts in this topic:
* Being a scripted language, I don't think that the language will ever support the same paradigm as in OCaml when you can mask a type's concrete value, making it abstract
* Thus, abstract types for us will most likely only ever be used for opaque values, like a binding's configuration element and etc.

toots: If these are correct assumptions then the implementation is also probably correct as well.

src/lang/lang.ml
```diff
@@ -715,3 +717,27 @@ let interactive () =
...
+
+module type AbstractDef = sig
+  type content
+
+  val name : string
+end
+
+module MkAbstract (Def : AbstractDef) = struct
+  let t = ground_t (T.Abstract Def.name)
+  let to_value c = mk ~t:string_t (Abstract (Def.name, Obj.magic c))
```
smimram: Can't we avoid `Obj.magic` if we have a static list of abstract types, e.g. with a GADT?

toots:
@smimram Check it out now!

Issue #1052, opened 2019-12-12
Question about redudant sources without cascading switching


kleinadam:
Hello liquidsoap users,

I would be happy if you could help finding a solution for my problem. 
```
radio = fallback(track_sensitive=false,[ingest1,ingest2,emergency])
```
ingest1 und ingest2 are two equivalent and identical content sources (= live mountpoints on icecasts servers in different data centers, both sources are “fallible”. 

emergency is a local file. 

When ingest1 fails the switch to ingest2 is done well, but if ingest1 comes up again, is there a way not to switch back ingest1 and stay on ingest2? ingest1 or ingest2 are both valid sources. 

I need: stay on ingest1 or ingest2, only switch to emergency if both ingest1 AND ingest2 are offline.

Regards,
Adam

smimram:
You can define the following function which will select the nth source in a list.
```
def switch.nth(~id="",~track_sensitive=true,~on_unavailable={()},n,l)
  n = to_int_getter(n)
  def p(i, s)
    ans = i == n()
    if ans and not (source.is_ready(s)) then on_unavailable() end
    ans
  end
  l = list.mapi(fun (i, s) -> ({p(i, s)}, s), l)
  switch(id=id, track_sensitive=track_sensitive, l)
end
```
Then you can use it to reselect only if the source becomes unavailable:
```
sources = [ingest1,ingest2,emergency]

n = ref(0)

def on_unavailable()
  n := list.index(source.is_ready, sources)
end

switch.nth(on_unavailable=on_unavailable, {!n}, sources)
```

kleinadam:
Thank you very much Samuel, I am not familiar with that kind of scripting language, but I will try out your code, Regards, Adam

PR #1076, opened 2020-01-16
Add switch.nth.

smimram:
See #1052.


smimram:
For the record, this should be superseeded by #1180 .

smimram:
There are better ways to do this (see #1180), closing.

PR #1097, opened 2020-02-23
Split documentation between essential and extra.

toots:


libs/hls.liq
```diff
@@ -3,7 +3,7 @@
 # @param ~id Force the value of the source ID.
 # @param ~reload How often (in seconds) the playlist should be reloaded.
 # @param uri Playlist URI
-# @flag experimental
+# @flags experimental
```
smimram: did you really mean `@flags` here?

libs/playlist.liq
```diff
@@ -4,6 +4,7 @@
 # reload function can optionally be called with a new playlist URI, otherwise,
 # it reloads the previous URI.
 # @category Source / Input
+# @flag extra
```
smimram: This one looks pretty important to me.
```suggestion
```

libs/video.liq
```diff
@@ -18,6 +18,7 @@ end
 
 # Add a static image on the first video channel.
 # @category Source / Video Processing
+# @flag extra
```
smimram: I expect that "many people" will use this in order to add logos.
```suggestion
```

libs/hls.liq
toots:
```suggestion
# @flag experimental
```

PR #1107, opened 2020-03-10
Fixed header comparison condition in doc example for harbor http proxy

naitsirch:
string.capitalize does not make string to uppercase


toots:
Thanks for these changes!

PR #1119, opened 2020-03-25
fix typo in ffmpeg error message

mcfiredrill:
Hi! Just noticed this when I was trying to get input.ffmpeg to work.


toots:
Thks!

PR #1124, opened 2020-03-28
Fix doc genration of settings

chrko:
Currently the generated html breaks as the browser interprets `<pid>`, `<script>` and `<sysrundir>` as html tags. Especially `<script>` hides the following output.


toots:
Good catch, thank you!

PR #1129, opened 2020-04-03
Fix rotate selection.

toots:
The current implementation of source selection on `rotate` operator is too obscure and actually bug prone. Here's one example:

- If children sources are: `[s1, s2, s3, s4]` all with weight `1`.
- `pos` is `2`
- `s2` ends its track
- `s2` is unavailable after finishing its track, all other sources are
- The calculated `ready_list` will then contain: `[s1, s3, s4]`
- The algorithm selects `s4` as `pos+1`

The issue here is that unavailable sources should be skipped over but the selection list should respect original positions and move on to the next source in the list if the source to be selected isn't available.

This PR implements that and, hopefully, clarifies the code a bit.

### DIFF
--- a/src/operators/switch.ml
+++ b/src/operators/switch.ml
@@ -475,7 +475,7 @@ let () =
         ~kind ~replay_meta ~override_meta ~transition_length:tl ts children)
 
 (** Random switch *)
-exception Found of child
+exception Found of child option
 
 class random ~kind ~override_meta ~transition_length ?replay_meta strict mode
   children =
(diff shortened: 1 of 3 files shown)


src/operators/switch.ml
```diff
@@ -486,30 +486,53 @@ class random ~kind ~override_meta ~transition_length ?replay_meta strict mode
...
+        let ready_list, ready_count =
+          List.fold_left
+            (fun (l, n) (_, s) ->
+              if s.source#is_ready then (Some s :: l, n + 1) else (None :: l, n))
+            ([], 0) children
+        in
+
+        if ready_count = 0 then raise (Found None);
+
+        let ready_list = List.rev ready_list in
```
smimram, line 520: Why not directly use a `fold_right` above?

src/operators/switch.ml
```diff
@@ -486,30 +486,53 @@ class random ~kind ~override_meta ~transition_length ?replay_meta strict mode
...
+        in
+        let rec f () =
+          match List.nth ready_list position with
+            | Some s -> raise (Found (Some s))
+            | None ->
+                set_next_position ();
+                f ()
+        in
+        set_next_position ();
+        f ()
```
smimram, line 534: `f` can do the advance, it will simplify the code:
```suggestion
        let rec f () =
          if strict then position <- (position + 1) mod children_count
          else position <- Random.int children_count
          match List.nth ready_list position with
            | Some s -> raise (Found (Some s))
            | None -> f ()
        in
        f ()
```

src/operators/switch.ml
```diff
@@ -486,30 +486,53 @@ class random ~kind ~override_meta ~transition_length ?replay_meta strict mode
...
+
+        if ready_count = 0 then raise (Found None);
+
+        let ready_list = List.rev ready_list in
+
+        let set_next_position () =
+          if strict then position <- (position + 1) mod children_count
+          else position <- Random.int children_count
+        in
+        let rec f () =
```
smimram, line 526: Find a better name than `f` :)

src/operators/switch.ml
toots: Not tail-rec?

smimram: Well, I doubt that we will have a rotate with thousands of sources...

toots: Well guess what.. The list is reverted when initializing the source.. Tricky!

src/operators/switch.ml
toots: giphy (1)

PR #1133, opened 2020-04-03
Playlists as lists.

smimram:
We can now play a list of files (as opposed to a playlist stored in a file). This is useful for scripts generating playlists.


toots:
Maybe add a `CHANGES` entry?

Issue #1132, opened 2020-04-03
Allow cross() to accept sources with video


dhannyz:
Currently cross only accepts `source(audio='#a+1, video=0, midi=0)`
It would be great if it could accept sources with video to apply some crossfades or other transitions

Edit: One final thing, with the `fade.initial` `fade.final`, they do accept sources with video, does this mean they fade both the audio volume and video to black?
Because that would be the default kind of transition for cross with video

smimram:
@toots after a quick look at the code for cross I don't see anything which really requires preventing video in cross, do you?

PR #1134, opened 2020-04-03
Crossfade videos

smimram:
Allow crossfading for videos too. Should solve #1132 once it is working (which it unfortunately does not right now...)


smimram:
Messed up this one, will do it again.

PR #1135, opened 2020-04-03
Allow crossfading video sources.

smimram:
Should solve #1132 once it is working.


libs/fades.liq
```diff
@@ -66,6 +66,15 @@ def mkfade(~type="lin",~start=0.,~stop=1.,~duration=3.,~on_done={()},s) =
   fade
 end
 
+# Scale source during fading.
+# @category Source / Sound Processing
+# @flag hidden
+def fade.scale(~id="fade.scale", x, s)
+  s = amplify(id=id, override="", x, s)
+  s = video.opacity(id=id, x, s)
+  s
```
toots, line 75:
```suggestion
  video.opacity(id=id, x, s)
```

libs/fades.liq
```diff
@@ -115,7 +124,7 @@ def fade.out(~id="fade.out",~duration=3.,
   delay = fun () -> !duration
   s = on_end(delay=delay,start_fade,s)
   s = on_track(stop_fade,s)
-  amplify(id=id,override="",apply,s)
+  fade.scale(id=id,apply,s)
```
toots, line 127: Isn't that gonna force all sources to be `video=1+#a`?

smimram: Nope, see https://github.com/savonet/liquidsoap/pull/1135/files#diff-05a71795aba4f59cf7098a4c39224665R48
Video effects don't actually require to have video (which I think is ok for this reason).

toots: Cool. If you've tested it then it's 👍 for me!

toots:
Make sure to add a `CHANGES` entry tho!

smimram:
So, this is kind of working, but Liquidsoap is switching to blank in `mksafe` after the first transition...

@toots any idea?

A test script:
```
s = playlist.list(["video1.mp4", "video2.mp4"])
s = crossfade(s)
s = mksafe(s)
output.sdl(drop_audio(s))
out(s)
```

Issue #1136, opened 2020-04-03
Blit error


smimram:
The following script
```
s = single("test.avi")
s = add([s,s])
s = mksafe(s)
output.sdl(drop_audio(s))
out(s)
```
gives rise to the following error
```
...
2020/04/03 14:23:05 [mksafe:3] Switch to add_36260.
2020/04/03 14:23:05 [clock.pulseaudio:2] Source sdl failed while streaming: (Invalid_argument "Array.iter2: arrays must have the same length")!
2020/04/03 14:23:05 [clock.pulseaudio:3] Raised at file "string.ml", line 115, characters 19-34
2020/04/03 14:23:05 [clock.pulseaudio:3] Called from file "sexp.ml", line 112, characters 13-47
2020/04/03 14:23:05 [mksafe:3] Switch to mksafe.
2020/04/03 14:23:05 [clock.pulseaudio:2] Source pulse_out(liquidsoap:) failed while streaming: (Invalid_argument "Array.iter2: arrays must have the same length")!
2020/04/03 14:23:05 [clock.pulseaudio:3] Raised at file "string.ml", line 115, characters 19-34
2020/04/03 14:23:05 [clock.pulseaudio:3] Called from file "sexp.ml", line 112, characters 13-47
2020/04/03 14:23:05 [clock.pulseaudio:3] Streaming loop stopped.
...
```
which comes from `Frame.blit_content`.

If I remove the `output.sdl(...)` in the script, the error is not present.

I strongly suspect that this is related to @toots #1083 PR...

toots:
![giphy](https://user-images.githubusercontent.com/871060/78412000-79329d00-75d7-11ea-919b-4b7571f4bc31.gif)

toots:
Here's another, more minimal script to reproduce:
```ruby
s = (noise():source(2,1,0))
s = add([s,s])
output.dummy(s)
```
Looks like the `add` is triggering it.

toots:
This one works:
```ruby
s = (noise():source(2,1,0))
s = (add([s,s]):source(2,1,0))
output.dummy(s)
```

Looks like a typing/initialization issue, seems like the `add` operator is initiated with an internal buffer that has only audio.

smimram:
Same error on same script.

toots:
Looks like it really wasn't me after all.. 😅

PR #1138, opened 2020-04-04
Check sources content when passing sources to an operator.

toots:
This is a fix for #1136. Now that we don't silently accept to change internal frame's content on the fly (see: #1083), we are now seeing cases of incompatible source content that were previously silently accepted at runtime.

These changes add a new check for these cases and raise an error with, hopefully, a user-friendly error:
Screenshot 2020-04-04 11 25 10


toots:
After further inspection, it's not clear if this is needed. The underlying issue was closed with 3ffe574. Keeping this one open for a minute to see if we need it.

PR #1174, opened 2020-04-25
[WIP] Avoid copying video content unless told to.

toots:
This PR establishes the basis for avoiding video content copy. The reasoning is that, unless an operator explicitly declares itself as needing fresh video data, we should be able to pass video frames untouched all the way through.

Early testing seems to indicate that this should be working.

### DIFF
--- a/src/operators/video_effects.ml
+++ b/src/operators/video_effects.ml
@@ -38,6 +38,8 @@ class effect ~name ~kind effect (source : source) =
 
     method abort_track = source#abort_track
 
+    method needs_fresh_video = true
+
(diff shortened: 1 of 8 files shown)


smimram:
I'll have a look in details, but I already have a question: is there a way of checking things? like an API to request frames for editing so that we can check that the flag is indeed set when it should...

toots:
Closing this one for now. We don't need to copy audio data when filling buffers, we'll see about video copies when it becomes a thing. I'm anticipating that most video use will be w/o using filters.

PR #1175, opened 2020-04-26
Documentation: fix a cookbook section

martinkirch:
This fixes and improves the "Dump a stream into segmented files" section of the cookbook.


toots:
That looks great, thanks so much!

PR #1178, opened 2020-04-27
Generalize audio_to_stereo to video frames.

smimram:
Also handle frames without audio.

### DIFF
--- a/src/conversions/audio_to_stereo.ml
+++ b/src/conversions/audio_to_stereo.ml
@@ -41,7 +41,9 @@ class basic ~kind source =
           (* Set audio layer. *)
           let audio =
             match Frame.(tmp_frame.content.audio) with
-              | [||] -> assert false
+              | [||] ->
+                  let buf = Audio.Mono.create (AFrame.size ()) in
+                  [| buf; buf |]
               | [| chan |] -> [| chan; chan |]
               | audio -> Array.sub audio 0 2
           in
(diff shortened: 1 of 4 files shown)


src/conversions/audio_to_stereo.ml
```diff
@@ -41,18 +41,23 @@ class basic ~kind source =
...
               | audio -> Array.sub audio 0 2
           in
           Frame.set_audio frame audio)
   end
 
 let () =
-  let input_kind = Lang.kind_type_of_kind_format Lang.audio_any in
+  let input_kind = Lang.kind_type_of_kind_format Lang.any in
+  let { Frame.audio; video; midi } = Lang.of_frame_kind_t input_kind in
+  ignore audio;
```
toots, line 56:
```suggestion
  let { Frame.video; midi; _ } = Lang.of_frame_kind_t input_kind in
```

src/conversions/audio_to_stereo.ml
```diff
@@ -41,18 +41,23 @@ class basic ~kind source =
           (* Set audio layer. *)
           let audio =
             match Frame.(tmp_frame.content.audio) with
-              | [||] -> assert false
+              | [||] ->
+                  let buf = Audio.Mono.create (AFrame.size ()) in
+                  [| buf; buf |]
```
toots, line 46: Are we sure that this is created empty?

PR #1181, opened 2020-04-28
Correct xxx_getter type.

smimram:
The attached patch corrects the type of `xxx_getter` (e.g. `int_getter`) to what it was in the past, which is what it should actually be: we should keep the link between the input and the output, both should be `xxx` or both should be `{xxx}`, but we cannot change them independently. However, with this patch I cannot seem to give default values to getter arguments, which is quite a limitation (and would justify in my opinion being slightly unsafe).

For instance, the following code is going on without any problem:
```
def f(~x)
  x = to_int_getter(x)
  print(x() + 2)
end

f(x=3)
f(x={4})
```
We can thus define functions with getter arguments.

However, as soon as we try to give a default value for the argument, things go bad. The following code
```
def g(~x=int_getter(2))
  x = to_int_getter(x)
  print(x() + 2)
end

g()
g(x=3)
g(x={4})
```
is not accepted with the patch in the PR, but is accepted in the current master. The result of `int_getter(2)` is an integer and not a getter to an integer and the last call (`g(x={4})`) is failing.

How do we properly fix this?

### DIFF
--- a/src/lang/lang_builtins.ml
+++ b/src/lang/lang_builtins.ml
@@ -103,12 +103,12 @@ let () =
       (fun p ->
         let getter = to_get (Lang.assoc "" 1 p) in
         Lang.val_fun [] ~ret_t:type_t (fun _ _ -> to_val (getter ())));
+    let a = get_t () in
     add_builtin ~cat:Liq (name ^ "_getter")
       ~descr:
         ( "Identity function over " ^ name ^ " getters. "
         ^ "This is useful to make types explicit." )
-      [("", get_t (), None, None)]
-      (get_t ())
+      [("", a, None, None)] a
       (fun p -> List.assoc "" p)
   in
   add_getters "string" Lang.string_getter_t Lang.string_t Lang.to_string_getter


dbaelde:
I have not looked at the code, but let's start by understanding your example: I imagine that type inference guesses `int` for the argument `x` of `g`. It would not be surprising given the default value -- iirc we do not delay typing constraints to obtain more general types, as would be the case here. Can you force `int or int getter` (using a constraint) manually, and does it fix the problem ?

smimram:
Yes @dbaelde, your suggestion is indeed working: this is basically what the current implementation does...

toots:
@smimram In line with what David wrote, it seems to me that it is still type-safe to disconnect input and output type, as long as they are both getters of the same type, right?

smimram:
I don't think so: currently, I think that the output type of `int_getter` means you can have either an `int` or a `() -> int`, so that we could break the typing system.

smimram:
My last comment seems to be wrong:
```
a = int_getter(3)
def f(x={1})
  print(x)
end
f(a)
```
raises
```
At line 3, char 4-17:
Error 5: this value has type
  int (inferred at line 3, char 15)
but it should be a subtype of the type of the value at line 7, char 2
  () -> _ (inferred at line 4, char 8-11)
```
whereas it should exhibit a runtime error according to my explanation...

So everything is alright as is?

toots:
Yes, getters are not accessible directly as a function through the scripting API, you need to first convert them to a function via `to_int_getter`. 

I don't have a current `master` at the moment but what you want to be testing is this:
```
def f(x=int_getter(2)) =
  x()
end
```

This should not work, instead you do:
```
def f(x=int_getter(2)) =
  fn = to_int_getter(x)
  fn()
end
```

This way you can be sure that wether you call `f(2)` or `f({2})`, you never have issues, which is why I believe that this whole magic trick is indeed type-safe.

toots:
I believe that this one can be closed. @smimram feel free to reopen/disagree.. :-)

PR #1183, opened 2020-05-01
Implement switch with dynamic sources

smimram:


smimram:
The following tests seem to be ok:
```
s = fallback([once(sine(duration=2., 440.)), sine(880.)])
```
```
s = sequence([sine(duration=2., 440.), sine(880.)])
```
```
flag = ref(true)
thread.run.recurrent(delay=2., fun()->begin flag := not !flag; 2. end)
s = mksafe(switch([({!flag},sine(440.)), ({true},sine(880.))]))
```
```
s = rotate([sine(duration=2.,400.), sine(duration=1.,500.), sine(duration=2.,600.)])
```

PR #1185, opened 2020-05-02
Native channel converter.

### LINKED ISSUE #1179: gtts: incompatible number of channels
In my script:
```none
set("server.telnet", true)
set("server.telnet.port", 1234)
...
# A special queue for smooth_add requests
special = request.queue(id="special")
...
radio = smooth_add(delay=0.5, p=0.15, normal=radio, special=special)
...
radio = audio_to_stereo(radio)
...
output.icecast(%mp3.vbr,
...,
radio)
```

Then, on the telnet server, I try:
`special.push gtts:testing testing`

In the logs, I get:
```none
2020/04/27 23:50:47 [decoder.ffmpeg:3] File "/tmp/liq-process1f578d.mp3" has an incompatible number of channels.
2020/04/27 23:50:47 [decoder.mad:3] File "/tmp/liq-process1f578d.mp3" has an incompatible number of channels.
2020/04/27 23:50:47 [decoder:3] Unable to decode "/tmp/liq-process1f578d.mp3" as {audio=2;video=0;midi=0}!
```
And no TTS output on the stream. What's going wrong?

toots:
This PR cleans-up the decoder API. It contains the following changes:
- Adds a generic channel converter API, based on the samplerate converter API.

Future plan: maybe add a ffmpeg channel converter? Seems a bit overkill, though, since we'd have to copy back and forth to ffmpeg frames. Might get to it if we ever actually have to support `5.1`.

- Clean-up & factor out the decoder API to apply a strict separation of concerns. In particular:
  - Decoder should have a declarative way of expressing their content, mime and file extensions and wether they can decode files or streams
  - Decoders should not have to worry about choosing what kind of buffer they need
  - Decoders should not have to worry about converting samplerate to liquidsoap's internal samplerate
  - Decoders should not have to worry about converting channel layout
  - Decoders should not have to worry about resizing video frames
  - Decoders should not have to worry about converting video `fps`.

**However** when that makes sense, it is possible for individual decoders to handle any and all of that. Typically, the `ffmpeg` decoder can and should do all the above as part of its native filters pipeline, before delivering its data to us.

This PR adds priorities to decoders, which should allow to make sure that decoders such as `ffmpeg` decoder are always tested first if present and decoders such as `gstreamer` as always tested last.

Additionally, we also introduce the beginning of a generic media testing framework. This is far from complete but it should be a good start.

Side note: with this PR, SRT I/Os and `input.harbor` now work with video, as well as many other potential stream-based I/Os. Tests need to be adapted as they are currently not working.

For now, the ffmpeg stream decoder only decodes stream with the (non-standard) `application/ffmpeg` mime type. This needs to be revisited too at some point.

This PR also introduces implicit channel conversion thus fixes: #1179


src/converters/audio/native_audio_converter.ml
```diff
@@ -50,3 +50,14 @@ let samplerate_converter () =
 
 let () =
   Audio_converter.Samplerate.converters#register "native" samplerate_converter
+
+let channel_layout_converter src dst data =
+  assert (src <> dst);
```
smimram, line 55: Why don't we allow the identity conversion (when `src = dst`)?

src/converters/audio/native_audio_converter.ml
```diff
@@ -50,3 +50,14 @@ let samplerate_converter () =
...
+  match dst with
+    | `Mono -> [| Audio.to_mono data |]
+    | `Stereo -> [| data.(0); data.(0) |]
```
smimram, line 58: What if the input is 5.1?

src/converters/audio_converter.ml
```diff
@@ -99,3 +99,69 @@ module Samplerate = struct
            in
            f converters_conf#get))
 end
+
+module Channel_layout = struct
+  exception Unsupported
+  exception Invalid_data
+
+  type layout = [ `Mono | `Stereo | `Five_point_one ]
```
smimram, line 107: It would be nice to add `` `Empty `` so that we can generate blank, which is a very very useful case (for video). This would however passing the length of data we want to generate around.

src/converters/audio_converter.ml
```diff
@@ -99,3 +99,69 @@ module Samplerate = struct
...
+
+  type converter =
+    layout -> layout -> Frame.audio_t array -> Frame.audio_t array
+
+  type t = {
+    src : layout;
```
smimram, line 113: What do we need this field for?

src/decoder/decoder.ml
```diff
@@ -215,6 +215,38 @@ let test_file ?(log = log) ~mimes ~extensions fname =
       if not ext_ok then log#info "Invalid file extension for %S!" fname;
       false ) )
 
+let can_decode_audio_type ~strict decoded_type target_kind =
+  match target_kind with
+    | { Frame.audio } when audio = Frame.Zero -> false
```
smimram, line 220: Again, I'd like to see `true` there...

toots: @smimram This should do it:
```ocaml
let can_decode_kind decoded_type target_kind =
  let can_convert_audio audio =
    Audio_converter.Channel_layout.(
      try
        ignore
          (create
             (layout_of_channels decoded_type.Frame.audio)
             (layout_of_channels audio));
        true
      with Unsupported -> false)
  in
  match Frame.type_of_kind target_kind with
    (* Either we can decode straight away. *)
    | _ when Frame.type_has_kind decoded_type target_kind -> true
    (* Or we can convert audio and/or drop video and midi *)
    | { Frame.audio; video; midi } ->
        let audio =
          if can_convert_audio audio then audio else decoded_type.Frame.audio
        in
        let video = if video = 0 then 0 else decoded_type.Frame.video in
        let midi = if midi = 0 then 0 else decoded_type.Frame.midi in
        Frame.type_has_kind { Frame.audio; video; midi } target_kind
```

src/converters/audio/native_audio_converter.ml
toots: This case is filtered before so we don't hit the converters for them.

src/converters/audio_converter.ml
toots: Just added support for it! It'll be done with the decoders, not as channel layout converter. Thanks for pointing it out..

src/converters/audio/native_audio_converter.ml
toots: We'll delegate to ffmepg for that. I'll implement it next.

src/converters/audio_converter.ml
toots: To test if the user is passing incorrect data later:
```ocaml
  let convert { src; converter } input =
    if Array.length input != channels_of_layout src then raise Invalid_data;
    converter input
```

tests/Makefile
```diff
@@ -1,28 +1,127 @@
...
-LIQ = $(sort $(filter-out test.liq, $(wildcard *.liq)))
+DISTFILES = Makefile mk_stream_test.sh $(wildcard *.liq) $(wildcard *.pl)
+SUBDIRS = fixme tests generic
+LIQ = $(sort $(wildcard tests/*.liq))
 TESTS = $(basename $(LIQ))
 
 top_srcdir = ..
 
-test: flactest $(TESTS)
+# TODO: fix test_stream_video
```
toots, line 10:
```suggestion
```

PR #1195, opened 2020-05-06
Remove get and set for refs out of the kernel.

smimram:
The smaller the kernel is, the safer we are, and the simpler functions are to write and read. The slight performance impact should be unnoticeable.


toots:
I like it but I think that you should check with @dbaelde about it. There are some iterations over types that include get/set that I'd like to confirm are safe to remove..

dbaelde:
@smimram The only reason I see is that references force some specific treatments in the type system (to avoid abusive generalizations) so treating them as "libraries" is odd. But I don't see anything that prevents get and set to be moved outside of the kernel.

PR #1197, opened 2020-05-06
Add support for records and methods.

smimram:


smimram:
Note for later: `GH1129.liq` is failing. Here is simplified way to reproduce:
```
a = ()

def a.f(l)
  def rec aux(l) = [] end
  aux(l)
end

ignore(a.f([1]))
ignore(a.f([blank()]))

output.dummy(blank())
```
which raises, at *execution time*,
```
Error 5: this value has type
  [source(audio=2, video=0, midi=0)]
but it should be a subtype of the type of the value at ../libs/list.liq, line 82, char 6
  [?{int}] (inferred at ../libs/list.liq, line 80, char 29)
```
I strongly suspect that this is due to the fact that I am not going through inner types anymore (because this was posing problems with generalization in presence of type schemes). I propose that we fix this by removing types at execution...

libs/fades.liq
```diff
@@ -290,8 +292,8 @@ def crossfade(~duration=5.,~override_duration="liq_cross_duration",
               ~default=(fun (a,b) -> sequence([a, b])),
               ~high=-15., ~medium=-32., ~margin=4.,
               ~minimum=(-1.),~width=2.,~conservative=true,s)
-  fade.out = fade.out(type="sin",duration=fade_out)
-  fade.in  = fade.in(type="sin",duration=fade_in)
+  let fade.out = fade.out(type="sin",duration=fade_out)
+  let fade.in  = fade.in(type="sin",duration=fade_in)
```
toots, line 296: What is the scope is this `let`? Can it leak into the main record type?

libs/externals.liq
```diff
@@ -127,15 +127,15 @@ def enable_external_flac_decoder() =
   if file.which(flac) != "" then
     log(label="external.decoder","Enabling EXTERNAL_FLAC external decoder.")
     flac_p = "#{flac} -d -c -"
-    def test_flac(file) =
```
toots, line 130: Does that mean that record variables cannot be overridden?

libs/hls.liq
```diff
@@ -86,7 +86,7 @@ def input.hls(~id="",~reload=10.,uri)
 end
 
 # @flag hidden
-def output.harbor.hls.base(~id,~fallible,~on_start,~on_stop,~streams_info,
+def output.harbor.hls_base(~id,~fallible,~on_start,~on_stop,~streams_info,
```
toots, line 89: Does the `hidden` flag apply to the `output.harbor` type? Is it carried over to prevent displaying it in the documented type?

libs/list.liq
```diff
@@ -30,8 +30,11 @@ end
 
 # Get the n-th element of a list (the first element is at position 0), or `default` if element does not exist.
 # @category List
-def rec list.nth(~default, l, n)
-  list.case(l, default, fun (x, l) -> if n == 0 then x else list.nth(default=default, l, n-1) end)
+def list.nth(~default, l, n) =
```
toots, line 33: So, no recursive `record.member` definition. Doesn't sound too restrictive, though.

libs/thread.liq
```diff
@@ -10,23 +10,25 @@ def thread.run(~fast=true, ~delay=0., ~every=float_getter(-1.), f)
     ignore(f() == ())
     every()
   end
-  thread.run.recurrent(fast=fast, delay=delay, f)
+  thread.run_recurrent(fast=fast, delay=delay, f)
 end
 
+let thread.run.recurrent = thread.run_recurrent
```
toots, line 16: Could we avoid having the function defined in two places in the `thread` module?

libs/utils.liq
```diff
@@ -39,21 +39,21 @@ end
...
 %ifdef output.portaudio
-  output.prefered = output.portaudio
+  let output.prefered = output.portaudio
 %endif
 %ifdef output.pulseaudio
-  output.prefered=output.pulseaudio
+  let output.prefered=output.pulseaudio
 %endif
 # Output to local audio card using the first available driver in pulseaudio,
 # portaudio, oss, alsa, ao, dummy.
```
toots, line 59: Maybe it's a good time to deprecate this and use `output.ao` for this purpose.

libs/video.liq
```diff
@@ -197,18 +202,18 @@ def video.add_text.ffmpeg(~color=0xffffff, ~font="", ~size=18, ~x=10, ~y=10, tex
 end
 %endif
 
-video.add_text = video.add_text.native
+let video.add_text.best = video.add_text.native
```
toots, line 205: While we're there, how about `video.text.add`?

src/lang/builtins_json.ml
```diff
@@ -47,6 +47,7 @@ let rec to_json_compact v =
           Printf.sprintf "[%s]" (String.concat "," (List.map to_json_compact l))
         )
     | Lang.Tuple l -> "[" ^ String.concat "," (List.map to_json_compact l) ^ "]"
+    | Lang.Meth _ -> failwith "TODO: JSON of field not yet implemented"
```
toots, line 50: I think it could be okay to do `"<field>"` or something like that here. That's what's done with requests and other abstract types. Also, let's update the doc accordingly.

src/lang/builtins_math.ml
```diff
@@ -22,14 +22,14 @@
...
+
 let () =
   let add op name descr =
     let t = Lang.float_t in
     add_builtin name ~cat:Math ~descr [("", t, None, None)] t (fun p ->
-        match p with
-          | [("", { Lang.value = Lang.(Ground (Ground.Float a)); _ })] ->
-              Lang.float (op a)
-          | _ -> assert false)
+        let a = Lang.to_float (List.assoc "" p) in
```
toots, line 31: Must have been some very old code!

src/lang/builtins_thread.ml
```diff
@@ -22,8 +22,10 @@
 
 open Lang_builtins
 
+let () = Lang.add_module "thread"
+
 let () =
-  add_builtin "thread.run.recurrent" ~cat:Control
+  add_builtin "thread.run_recurrent" ~cat:Control
```
toots, line 28: I see.. Is it possible to remove or define a module then (see comment in the corresponding `.liq` file..)

src/lang/lang.ml
```diff
@@ -488,11 +505,16 @@ let rec assoc label n = function
 
 let type_and_run ~lib ast =
   Clock.collect_after (fun () ->
+      if true || Lazy.force Term.debug then
```
toots, line 508: Hmm?

src/lang/lang.ml
```diff
@@ -488,11 +505,16 @@ let rec assoc label n = function
...
+        Printf.eprintf "Type checking...\n%!";
       (* Type checking *)
       Term.check ~ignored:true ast;
 
+      if Lazy.force Term.debug then
+        Printf.eprintf "Checking for unused variables...\n%!";
       (* Check for unused variables, relies on types *)
       Term.check_unused ~lib ast;
+      if true || Lazy.force Term.debug then Printf.eprintf "Evaluating...\n%!";
```
toots, line 517: Hmm?

src/lang/lang_values.ml
```diff
@@ -460,18 +481,104 @@ module V = struct
...
+let _ =
+  let x = "x" in
+  let v = "v" in
+  let rec aux prefix = function
+    | l::ll -> Printf.sprintf "%s.%s where (%s = %s)" x (String.concat "." (List.rev prefix)) l (aux (l::prefix) ll)
+    | [] -> v
+  in
+  let ll = ["l1"; "l2"; "l3"] in
+  Printf.sprintf "%s = %s" x (aux [] ll)
+    *)
```
toots, line 535:
```suggestion
```

src/lang/lang_values.ml
```diff
@@ -692,7 +819,10 @@ let rec check ?(print_toplevel = false) ~level ~env e =
 let check ?(ignored = false) e =
   let print_toplevel = !Configure.display_types in
   try
-    check ~print_toplevel ~level:(List.length builtins#get_all) ~env:[] e;
+    let env = default_typing_environment () in
+    (* Printf.printf "Typing env\n%!"; *)
+    (* List.iter (fun (x,(g,t)) -> Printf.printf "%s : %s\n%!" x (T.print ~generalized:g t)) env; *)
```
toots, line 824:
```suggestion
```

src/lang/modules.ml
```diff
@@ -0,0 +1,52 @@
...
+      "input.external";
+      "liquidsoap";
+      "metadata";
+      "midi";
+      "output";
+      "os";
+      "playlist";
+      "reopen";
+      "request";
+      "request.dynamic";
```
toots, line 41: Doesn't that work like `mkdir -p`?

src/operators/frei0r_op.ml
```diff
@@ -23,6 +23,10 @@
 open Source
 open Extralib
 
+let () =
+  Lang.add_module "frei0r";
+  Lang.add_module "video.frei0r"
```
toots, line 28: Both?

src/synth/dssi_op.ml
```diff
@@ -125,6 +125,11 @@ class dssi ~kind ?chan plugin descr outputs params source =
         done
   end
 
+let () =
+  Lang.add_module "dssi";
+  Lang.add_module "synth.dssi";
```
toots, line 130: Both?

libs/fades.liq
smimram: Usual scoping.

The instruction
```
let r.a = f
```
is a shorthand for
```
r = {{ r | a = f }}
```
so that the scoping conventions are as usual.

libs/externals.liq
smimram: You can override as usual. But if you define
```
file = "toto"
```
then `file` is precisely `"toto"` (you lose all the fields), so that you cannot use `file.which` anymore: this was the problem here. We could do things in another way, but that the most simple one.

libs/hls.liq
smimram: `hidden` applies to `output.harbor.hls_base`. When we register a function in the doc, we register it as it is declared (with all the dots and so on). We don't register the intermediate records. I don't think the flags are carried over (didn't change anything on this side): `hidden` simply means that *this* definition won't occur in the doc.

libs/list.liq
smimram: It could be added though. Again
```
def list.nth(...) = ...
```
is expanded to
```
list = {{ list | nth = (...) -> }}
```
I leave it to you as an exercise to generalize this expansion in the case where we have recursive definitions... :)

src/lang/lang_values.ml
smimram: It took me quite some time to get this one correctly, I'd rather leave this intermediate code for now just in case we need to make some changes on it...

src/lang/modules.ml
smimram: Nope. We could, but we don't :) The main reason is that we wouldn't detect typos in module names with "`-p`"...

src/operators/frei0r_op.ml
smimram: Fixed in b11641b8730a0d6bff639dcb7362aeefb7e0168e.

src/synth/dssi_op.ml
smimram: For this one we need both: we currently have all the plugins registered as `synth.dssi.*` and we have the function `dssi.register` to manually register a plugin.

libs/fades.liq
toots: Cool!

libs/externals.liq
toots: Word.

libs/list.liq
toots: Yeah I don't think we need to prioritize that just now :-)

libs/thread.liq
toots: IF you can, don't think that should be a blocker.

src/lang/lang_values.ml
toots: What's the purpose of this code then? How about making it a OCaml test?

src/lang/modules.ml
toots: Gotcha!

smimram:
To me this can be merged excepting for one thing: I'd like to discuss first the notations for records.

For now we have
```
{{ a = 4 , b = "b" }}
```
I think that
```
{ a = 4 , b = "b" }
```
would be much better but there is a possible confusion with the notation
```
{x}
```
which is a shorthand for
```
fun () -> x
```
I think that it is not ambiguous from the grammar point of view, though: the two uses can be distinguished syntactically.

Also, we have to find a notation for decorated values, the natural one would be
```
{ 4 | a = 5 , b = "a" }
```
What do you think?

I'll have another PR in order to properly replace values.

toots:
Regarding the notation, how about: 
```
{| a = 4 , b = "b" |}

{4| a = 5 , b = "a" |}
```

smimram:
Let's dive into records!...

PR #1202, opened 2020-05-10
Compute kinds by unification in order to get rid of types during evaluation.

smimram:
This PR removes all types at execution.

The typing phase is left unchanged but after that we don't mention types anymore. The reasons why it is a good thing are that
- this speeds up execution (execution times x3 higher because we don't have to maintain types which is quite costly, for instance each time we use a variable we have to instantiate its type which requires going through the whole type and copying it, each time we apply a function we have to perform unification, etc.)
- the kernel is simpler: for instance generalization had to consider all types occurring in all terms, whereas we can simply consider the type of the current term now, as usual
- the kernel of the language is smaller and thus safer and more modular
- as a particular cas of the above point, I needed this in for record (#1197) which would be hell to implement otherwise.

The only place where types were needed are in order to compute the number of (audio/video/midi) channels to produce for operators. This is now inferred by unification when creating the operators (I use the same mechanism than for clocks). The PR also simplifies much the typing of channels: they roughly all have the same type whereas we had something mildly different to type frame kinds and source kinds, etc. We could go even further, but I didn't want to ruin the readability of this PR.

The downsides:
- We  essentially do the work twice: we check that the typing of channels is correct at typing time, and once again at runtime. In my opinion, this could be entirely removed from typing, since unification is performed right at the beginning of the execution of the program which is roughly the same as typing phase. The only difference is for dynamically created sources, but those aren't used much (for now). And there is no reason to incorporate the number of channels in typing, but not say size of images, samplerate, etc. Let's keep that at runtime. I did not remove it for now because our users are "used" to the current types `source('a,'b,'c)` (to be honest, I am not even sure of that: many of them would simply prefer `source`...)
- Since we do the work twice there is a slight risk of incoherence with the two approaches. The only one I see for now is that if you force the number of channels by casting (e.g. `(s : source(1,2,3))`) this will not actually be taken in account since the number of channels is guessed anyway. This could be easily solved by adding explicit channel casting operators to the language instead of using the typing system for that. Anyway, the kind (= number of channels) computed by unification should be safe, it only might not be what the user expected.


src/io/pulseaudio_io.ml
```diff
@@ -71,12 +70,14 @@ class output ~infallible ~start ~on_start ~on_stop ~kind p =
 
     val mutable stream = None
 
+    method channels = AFrame.channels_of_kind self#kind
```
toots, line 73: Why not just inlining this below? Seems like the only place it's used at and adding extra methods to objects makes typing annoying, even when `private`..

src/lang/builtins_json.ml
```diff
@@ -23,8 +23,7 @@
 let log = Log.make ["lang"; "json"]
 
 let rec to_json_compact v =
-  (* Utils.escape implements
-   * JSON's escaping RFC. *)
+  (* Utils.escape implements JSON's escaping RFC. *)
```
toots, line 26: 👍

src/lang/lang.ml
```diff
@@ -461,6 +400,9 @@ let iter_sources f v =
              * Display a fat log message to warn about this risky situation,
              * which probably won't prevent users to get biffled... *)
```
toots, line 401:
```suggestion
             * which probably won't prevent users to get biffled [I spot a mrpingouin here! - RB]... *)
```

src/lang/lang.ml
```diff
@@ -471,11 +413,12 @@ let iter_sources f v =
...
                           | { T.name = "source"; _ } -> true
                           | _ -> false
-                      then raise Found)
+                      then raise Exit)
                     v.t
                 in
                 has_var_pos
-              with Found -> true
+              with Exit -> true
+                *)
```
toots, line 421: TODO: Look at this again.

tests/type_errors.pl
```diff
@@ -46,13 +46,13 @@ sub correct {
 
 section("SUBTYPING");
 incorrect('(1:unit)');
-correct('ignore(((blank():source(1,1,1)):source(*,*,*)))');
-incorrect('((blank():source(*,*,*)):source(1,1,1))');
+# correct('ignore(((blank():source(1,1,1)):source(*,*,*)))');
+# incorrect('((blank():source(*,*,*)):source(1,1,1))');
```
toots, line 50:
```suggestion
```

tests/type_errors.pl
```diff
@@ -46,13 +46,13 @@ sub correct {
...
 # Next one requires the inference of a subtype (fixed vs. variable arity)
 correct('ignore(audio_to_stereo(add([])))');
 correct('ignore((blank():source(audio=1,video=1,midi=1)))');
-correct('ignore((blank():source(*+1,0,0)))');
-correct('ignore((blank():source(1+*,0,0)))');
+# correct('ignore((blank():source(*+1,0,0)))');
+# correct('ignore((blank():source(1+*,0,0)))');
```
toots, line 55:
```suggestion
```

src/io/pulseaudio_io.ml
smimram: The number of channels might be used multiple times (it is not the case here, but I am using the same coding pattern everywhere) and I find it clearer this way. I'll add `private` though. Thanks much for the comment though, I have spotted many places where we were using the kind too early!...

smimram: Actually not, it seems that I can remove delaying channels computations from the diff! :)

tests/type_errors.pl
smimram: Nope, the type `*` does not exist anymore: I have remove `Any`

tests/type_errors.pl
smimram: Same as above.

tests/type_errors.pl
toots: So why not remove the test? 🙂

src/conversions/conversion.ml
```diff
@@ -37,13 +37,21 @@ class base ?(audio = false) ?(video = false) ?(midi = false) ~converter
 
     method self_sync = source#self_sync
 
-    (* The tmp_frame is intended to be filled by the underlying
-     * source. Content untouched by the converter are replaced by
-     * by content from the calling frame. Touched content get their
-     * own layer. *)
-    val mutable tmp_frame = Frame.create source#kind
+    val mutable tmp_frame = None
```
toots, line 40: Why not using `lazy` values for these like we do for frame settings?

src/decoder/decoder.ml
```diff
@@ -247,37 +247,36 @@ let can_decode_kind decoded_type target_kind =
...
-      | Zero, Zero, _ -> [`Midi]
-      | _ -> [])
+        target_type = { Frame.audio; video; midi }
+
+let decoder_modes ctype =
+  let audio = ctype.Frame.audio in
+  let video = ctype.Frame.video in
+  let midi = ctype.Frame.midi in
+  if audio = 0 && video = 0 && midi = 0 then
+    [`Audio_video; `Audio; `Video; `Midi]
```
toots, line 267:
```suggestion
    [`Metadata]
```

Let's make a `` `Metadata`` decoder type for this case.

src/decoder/external_decoder.ml
```diff
@@ -100,9 +100,10 @@ let test_kind f filename =
   else
     Some
       {
-        Frame.video = Frame.Zero;
-        midi = Frame.Zero;
-        audio = (if ret < 0 then Frame.Succ Frame.Any else Frame.mul_of_int ret);
+        Frame.video = 0;
+        midi = 0;
+        (* TODO: this is not perfect *)
+        audio = (if ret < 0 then Lazy.force Frame.audio_channels else ret);
```
toots, line 106: Yeah it's not perfect but it would be implicitly converted via the new decoder workflow.

src/decoder/external_decoder.ml
```diff
@@ -150,11 +149,12 @@ let external_input_oblivious process filename prebuf =
     try Process_handler.kill process with Process_handler.Finished -> ()
   in
   let input = { Decoder.read; tell = None; length = None; lseek = None } in
-  let kind = Frame.{ audio = Any; video = Zero; midi = Zero } in
-  let gen =
-    Generator.create ~log_overfull:false ~log:(log#info "%s") ~kind `Audio
+  (* TODO: is this really what we want for audio channels? *)
+  let ctype =
+    Frame.{ audio = Lazy.force Frame.audio_channels; video = 0; midi = 0 }
```
toots, line 154: Same thing here, we should be able to trust our implicit channel conversion now. I haven't looked at the details of this part of the code, though.

src/decoder/ffmpeg_decoder.ml
```diff
@@ -1,24 +1,27 @@
+;;
+1
+
```
toots, line 3:
```suggestion
```

src/decoder/metadata_decoder.ml
```diff
@@ -0,0 +1,94 @@
...
+  { Decoder.fill; fseek = (fun _ -> 0); close = ignore }
+
+let empty = { Frame.audio = 0; video = 0; midi = 0 }
+
+let () =
+  Decoder.file_decoders#register "META" (fun ~metadata:_ filename ctype ->
+      if ctype = empty then (
+        ignore (parse_file filename);
+        Some (fun () -> file_deco filename) )
+      else None)
```
toots, line 94: I don't think this compiles. The new decoder registration API is fully declarative, you might need to add this file back to the `Makefile` and finish converting it. As I said above, let's add a `` `Metadata`` decoder type for it.

src/io/portaudio_io.ml
```diff
@@ -64,6 +63,8 @@ class output ~kind ~clock_safe ~start ~on_start ~on_stop ~infallible buflen
           ~name:"output.portaudio" ~output_kind:"output.portaudio" val_source
           start as super
 
+    method private channels = self#ctype.Frame.audio
```
toots, line 66: Note for the future: at some point we might want to factor out the patterns that are repeated for I/O operators.

src/lang/builtins_ffmpeg_filters.ml
```diff
@@ -228,26 +228,36 @@ let () =
 
   add_builtin ~cat:Liq "ffmpeg.filter.audio.input"
     ~descr:"Attach an audio source to a filter's input"
-    [("", Graph.t, None, None); ("", audio_t, None, None)] Audio.t (fun p ->
+    [
+      ( "channels",
+        Lang.int_t,
+        Some (Lang.int 2),
+        Some "Number of audio channels." );
```
toots, line 235: Yes, that works!

src/lang/builtins_files.ml
```diff
@@ -243,7 +241,20 @@ let () =
...
+    [("", Lang.string_t, None, Some "Read metadata from a file.")]
+    Lang.metadata_t
+    ~descr:"Call a function when a file is modified. Returns unwatch function."
+    (fun p ->
+      let uri = Lang.to_string (List.assoc "" p) in
+      let r = Request.create uri in
+      let ctype = { Frame.audio = 0; video = 0; midi = 0 } in
+      if Request.resolve ~ctype:(Some ctype) r 30. = Request.Resolved then
+        Lang.metadata (Request.get_all_metadata r)
+      else Lang.metadata (Hashtbl.create 0))
```
toots, line 257: Why was this moved back to builtin?

src/lang/builtins_prometheus.ml
```diff
@@ -64,19 +64,15 @@ let add_metric metric_name create register set =
       let m =
         create ~label_names ?registry:None ~help ?namespace ?subsystem name
       in
-      Lang.val_fun
-        [("label_values", "label_values", Lang.list_t Lang.string_t, None)]
-        ~ret_t:set_t
-        (fun p _ ->
+      Lang.val_fun [("label_values", "label_values", None)] (fun p ->
```
toots, line 67: Let's make prometheus a `on_frame` scripted hook if we can after this.

src/lang/lang_types.ml
```diff
@@ -73,7 +73,7 @@ let print_pos ?(prefix = "at ") (start, stop) =
 (** Ground types *)
 
 type ground = ..
-type ground += Bool | Int | String | Float
+type ground += Bool | Int | String | Float | Request
```
toots, line 76: Nice. I can see why you would want to add `Source` here.

src/lang/lang_types.ml
```diff
@@ -622,11 +621,13 @@ let rec ( <: ) a b =
                   try (* TODO use variance info *)
                       h1 <: h2
                   with Error (a, b) ->
+                    let bt = Printexc.get_raw_backtrace () in
                     let post = List.map (fun (v, _) -> (v, `Ellipsis)) t1 in
-                    raise
+                    Printexc.raise_with_backtrace
```
toots, line 626: Didn't know this one!

src/conversions/conversion.ml
smimram: I've thought of that but you can't: you need to access `self` to get the content type (`self#ctype`) and OCaml won't let you have access to it when initializing instance variables (even under a lazy, which is legit but I guess difficult to check).

src/decoder/external_decoder.ml
smimram: Alright then :)

src/decoder/ffmpeg_decoder.ml
smimram: ouch

src/decoder/metadata_decoder.ml
smimram: Oh, you removed this and it got added back by the merge. Let's remove it.

src/io/portaudio_io.ml
smimram: Yes...

src/lang/builtins_ffmpeg_filters.ml
smimram: :)

src/lang/builtins_files.ml
smimram: Well done, you spotted a tricky point. Let me explain.

Because you don't determine the content type with typing anymore, I had to remove the type from requests: when you create you don't determine it content type. For instance, the type of `request.create` used to be
```
(...) -> request(audio='a, video='b, midi='c)
```
and is now
```
(...) -> request
```
Of course I did not remove the content type checks, but content type is determined at resolution time. For this reason, you cannot use `request.resolve` in Liquidsoap anymore, which explains why I had to move this to the OCaml side. All requests resolutions are performed in the OCaml side (not entirely true, see below).

Side note: there is still a `request.resolve` in Liquidsoap, but it resolves without content type: this corresponds to what you used to obtain with `request.create_raw` before and can only be used to download files which are not going to be decoded (e.g. playlists).

src/lang/builtins_prometheus.ml
smimram: Yeah!

src/lang/lang_types.ml
smimram: You cannot add `Source` here because it depends on the number of channels, so it has to remain a constructed type. I could do this with requests because, as explained above, they don't depend on other types anymore.

src/lang/lang_types.ml
smimram: Yeah, it's quite useful to keep all the backtraces.

src/decoder/decoder.ml
smimram: I used another way to solve the thing (`file.metadata` directly requests resolution of metadata), but you should be pleased: we now have `[]` as answer for this case.

smimram:
> Man what a tremendous PR!! I've had a first superficial pass, made some comments. I'll have another one later this weekend, I want to make sure that I understand how this works with dynamic sources.

Thanks!

> Also, how does it look when throwing an error? This should be user-friendly..

Yes, I have to test this. But to be honest, it is quite hard to think of an error which is not detected by the typing system but only by dynamic unification. Do you have an idea?... I'll try to think further about it.

libs/deprecations.liq
```diff
@@ -239,8 +239,14 @@ def request.dynamic(~id="",~conservative=false,~default_duration=30.,
...
-# Deprecated: this function has been replaced by fail.
+# Deprecated: this function has been replaced by `fail`.
 # @flag hidden
 def empty(~id="")
   fail()
 end
+
+# Deprecated: use `request.create` instead.
+# @flag hidden
+def request.create.raw(~indicators=[], ~persistent=false, uri)
```
toots, line 250:
```suggestion
def request.create.raw(~indicators=[], ~persistent=false, uri)
  deprecated("request.create.raw", "request.create")
```

src/conversions/conversion.ml
toots: Makes sense indeed thks.

src/decoder/decoder.ml
```diff
@@ -247,37 +247,35 @@ let can_decode_kind decoded_type target_kind =
...
-      | Zero, Zero, Zero -> []
-      | Zero, Zero, _ -> [`Midi]
-      | _ -> [])
+        target_type = { Frame.audio; video; midi }
+
+let decoder_modes ctype =
+  let audio = ctype.Frame.audio in
+  let video = ctype.Frame.video in
+  let midi = ctype.Frame.midi in
+  if audio = 0 && video = 0 && midi = 0 then []
```
toots, line 266: Cool thks!

src/lang/builtins_files.ml
toots: Ho I see, `request.resolve` passes the negotiated content type down to the decoder at runtime. Cool, it was always tricky to explain how to properly resolve requests without blocking the main thread anyway.

src/lang/lang_types.ml
toots: I meant to say: why you would want to also remove content types from source and then add it here :-)

smimram: Ah sorry! Let's keep this for the future though...

PR #1213, opened 2020-05-24
Restore request.dynamic.

smimram:
Following a question on slack, I don't think that it is a good idea to deprecate `request.dynamic`: since we have the `available` parameter, there is a perfectly good way of functioning by using this to indicate that we have no request to push, and when it is the case we typically want to add one request at a time. Of course, this is an easy variation over `request.dynamic.list`, but since some users were already used to this one, let's keep it simple for them.


toots:
Thanks, makes sense to me.

PR #1216, opened 2020-05-26
Allow different video sizes for different sources.

### LINKED ISSUE #1214: Support for different resolution / fps video
My 'broadcast' output is set with `set("frame.video....` at the top of my liq script. I am using `video.tile` to generate a `director preview` of each of the inputs so we know when it is safe to crossfade.

`video.tile` and the subsequent H.264 encoding are CPU heavy. I would like to be able to specify a lower resolution and lower fps value to ease the CPU burden.

smimram:
Should fix #1214.


smimram:
For now video size propagation by unification seems to work, but I gut size mismatchs in blits in generators (see the test). Help welcome:)

I think that this is a good test for the architecture before getting to raw contents...

toots:
I haven't looked at the code but I'm not sure if that's the right approach. It's a lot of code change for a CPU optimization.

If you look at modern live video producing software like OBS, they have a notion of canvas. You set a high size for the canvas and it is downsized to the output(s) size. Meanwhile, all elements on the canvas are added as layers and the final image is computed right before going into the encoder.

Could we do the same? Consider a global video frame size and allow video streams of any size under that, along with a position information (x/y offset)? Then we could compute the final image only once right before encoding it.

Screenshot 2020-05-28 08 52 59

toots:
That would actually finally give a good meaning to having multiple video channels like we allow..

gdhgdhgdh:
> [...] If you look at modern live video producing software like OBS, they have a notion of canvas. [...]

Any progress towards Liq becoming a 'headless OBS' would be wonderful!

toots:
I'm gonna close this as this should be done in the context of the upcoming extensible frame content work.

toots:
@gdhgdhgdh Forgot to respond to your comment. Yes! It's definitely in our minds. We want to implement a layered content type for video that would work like the layers in OBS. We have a big PR pending that allows for extensible frame contents here: https://github.com/savonet/liquidsoap/pull/1237 and we have another issue to remind us to implement the new content type: https://github.com/savonet/liquidsoap/issues/1261

Issue #1215, opened 2020-05-26
string.quote does not escape backticks 


naitsirch:
**Describe the bug**
The `string.quote` (in earlier versions it is just `quote`) should escape shell metacharacters. But it does not escape backticks "\`" which are used in shells to execute subcommands.

**To Reproduce**
str = "foo\`echo 1 > /tmp/foo\`"
str = string.quote(str)
get_process_lines("php my-script.php #{str}")

**Expected behavior**
The php script should get the plain string "foo\`echo 1 > /tmp/foo\`" as argument. But instead it only gets "foo" and `echo 1 > /tmp/foo` is executed as command. This could be seen as a security vulnerability.
Especially in regard to the authorization function shown [here](https://www.liquidsoap.info/doc-1.4.2/harbor.html).

**Version details**
 - OS: Debian
 - Version e.g. `1.3.7` (but as far as I see, there were no changes in earlier versions)

**Install method**
distribution package

toots:
Thanks for this report. Gonna look into it asap.

naitsirch:
Some further investigation into this topic show that there is much more to consider. Here are a few resources that may help:

- https://unix.stackexchange.com/a/270979/208546
- https://gist.github.com/Zenexer/40d02da5e07f151adeaeeaa11af9ab36 (this is especially for PHP, but may be interesting for you, too)
- https://fireshellsecurity.team/restricted-linux-shell-escaping-techniques/

toots:
Hi, I agree with you. The escape game is a loosing battle, it's much better to provide place holders for program arguments. I'll look into it ASAP. Or, else, feel free to send us a PR!

dtcooper:
I observe the dollar sign is not escape as well,

`get_process_output('echo ' ^ string.quote('$HOME'))` should produce the string `$HOME`, however is produces the string `/home/user`

dtcooper:
In case anyone else stumbles on this, as a stop-gap, I'm putting my variables in the `env` argument to `get_process_output`, ie `get_process_output(env=[('VAR', '$HOME')], 'echo "$VAR"')`.

dtcooper:
Consequently, my approach above _does_ sidestep this bug,

For example, per above,

```
str = 'foo`echo 1 > /tmp/foo`'
get_process_lines(env=[('STR', str)], 'echo php my-script.php "$STR"')
```

Works as expected.

toots:
Thanks for these details. This issue is on my radar.

[...]

PR #1220, opened 2020-05-27
Added hint about shell argument escaping to harbor doc

naitsirch:
For more background see https://github.com/savonet/liquidsoap/issues/1215#issuecomment-634522140


toots:
Thank you!

Issue #1191, opened 2020-05-04
Replace the Bubble example with Beets


martinkirch:
I've read with much interest the [Bubble example](https://www.liquidsoap.info/doc-1.4.2/bubble.html) - but it seems Bubble has been offline for a while. So instead I tried to plug [Beets](http://beets.io/), with pretty good results. Before writing a pull request to update the documentation, I'd like to validate the interest for this method and my implementation: 

```ocaml
def beets_protocol(~rlog,~maxtime,arg) =
  tmp_playlist = file.temp("beetsplaylist", ".m3u8")  # see playlists_cleaner below
  ignore(get_process_lines(
    "/home/me/beets/bin/beet random -f '$path' #{arg} > #{tmp_playlist}"
  ))
  [tmp_playlist]
end

add_protocol("beets", beets_protocol,
  syntax = "same arguments as beet's random module, see https://beets.readthedocs.io/en/stable/reference/query.html"
)

exec_at(freq=3600., pred={ true },
  fun () -> list.iter(fun(msg) -> log(msg, label="playlists_cleaner"),
    get_process_lines("find /tmp -iname beetsplaylist*m3u8 -mtime +0 -delete"),
  )
)

### EXAMPLES

# pick one hour of random songs, reload every 2000 seconds because `beet random -t60` returns less-than-60-minutes playlists
playlist_1h = mksafe(id="playlist 1h", playlist("beets:-t60", mode="normal", reload=2000))

# pick one hour of songs whose `genre` tag contains `Rock`
playlist_1h = mksafe(id="playlist 1h", playlist("beets:-t60 genre:Rock", mode="normal", reload=2000))
```
Writing to temporary files is not elegant but I couldn't find another way ... or would it be equivalent to call `request.create(indicators=get_process_lines...` ?

(by the way, following #1013 I guess a documentation is needed around `requests`, but I don't understand this part enough yet to contribute this)

martinkirch:
If you can read French, I've just published an article about this Beets integration https://martinkirch.github.io/2020-05-24/liquidsoap-musique-avec-beets.html

toots:
Hi! Sorry I let this slip. I'm down to swap the example, thanks for working on a new protocol! I'll put a tab on this to review it, feel free to ping us again if this gets delayed.

toots:
Yeah, this looks great! Feel free to drop us a PR updating the doc, I'll be happy to merge it!

martinkirch:
[...]

PR #1222, opened 2020-06-01
Replace bubble by beets

martinkirch:
... because Bubble is no longer available, whereas Beets is actively maintained.

Discussed in #1191


toots:
This looks great, really well written documentation, thank you so much. I'm merging here and will update the website as well.

PR #1227, opened 2020-06-06
Allow replacing the value of a decorated value.

smimram:
This PR introduces the new notation
```
def replaces f(x) = ... end
```
which mans that the value of `f` is replaced by the new one, but all the previous methods are kept.

Typical use:
```
def output.harbor.hls.base(...) = ... end

def output.harbor.hls(...) = output.harbor.hls.base(...) end
```
i.e. we need to define a subfield before defining the field.


smimram:
@toots this one looks fine to me, can you comment on it?

smimram:
I just merged master without any problem (didn't try to rebase).

toots:
> I just merged master without any problem (didn't try to rebase).

Yeah, rebase don't work well with multiple commits.. Looking now.

PR #1234, opened 2020-06-07
Adapt API to records.

### LINKED ISSUE #1231: Convert the rest of the API
* Convert all relevant APIs to records, e.g. `http.get` return type etc.
* Add method wrappers for e.g. `source.skip`
* Deprecate old API e.g. `source.skip`

smimram:
Fixes #1231.


libs/request.liq
```diff
@@ -55,18 +55,5 @@ def request.queue.pushable(~id="", ~conservative=false, ~default_duration=30., ~
...
-# @param ~id Force the value of the source ID.
-# @param ~conservative If true, estimated remaining time on the current track is not considered when computing queue length.
-# @param ~default_duration When unknown, assume this duration (in sec.) for files.
-# @param ~interactive Should the queue be controllable via telnet?
-# @param ~length How much audio (in sec.) should be queued in advance.
-# @param ~queue Initial queue of requests.
-# @param ~timeout Timeout (in sec.) for a single download.
-def request.queue(~id="", ~conservative=false, ~default_duration=30., ~interactive=true, ~length=40., ~queue=[], ~timeout=20.)
-  snd(request.queue.pushable(id=id, conservative=conservative, default_duration=default_duration, interactive=interactive, length=length, queue=queue, timeout=timeout))
+  s.{push=push}
```
toots, line 58: Do you think it would be possible to support `s.{support}` here? That would be really nice. I've been thinking about adding that to function calls for a while as well (`f(foo=foo)`)

smimram: Yes but let's leave this for the future :) This means more grammar productions, but apart from that I don't see any obvious obstacle.

PR #1235, opened 2020-06-08
Prometheus implementation on Liquidsoap side.

smimram:
@toots it would be nice if you could help me restore this code in order to test frames... Thanks!

### DIFF
--- a/libs/pervasives.liq
+++ b/libs/pervasives.liq
@@ -25,4 +25,5 @@
 %include "gstreamer.liq"
 %include "visualization.liq"
 %include "profiler.liq"
+%include "prometheus.liq"
 %include "deprecations.liq"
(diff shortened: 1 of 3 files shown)


gibix:
Hello! there is a plan to merge this for 2.0?

smimram:
Prometeus support will definitely be part of 2.0. This PR is only to test its implementation in Liquidsoap side instead of OCaml.

toots:
Prometheus is already supported in `1.4.4` actually but there was an issue that prevented it from being added to the online doc. I'll see about this and will also add a specific section about it in the documentation.

toots:
You can find an example of its use here: https://github.com/mbugeia/srt2hls/blob/master/radio/live.liq

PR #1236, opened 2020-06-08
Implement methods as maps

### LINKED ISSUE #1233: Make bytecode more efficient
Things I can think of
- make records with multiple fields (in a `Map`)

smimram:
Should be more efficient. Fixes #1233.


smimram:
Only my erathostenes: without
```
***** Performance tests *****
Computing Erathostenes (imperative)...
Computing Erathostenes (imperative) took 3.13447284698 seconds.
Computing Erathostenes (recursive)...
Computing Erathostenes (recursive) took 3.17217898369 seconds.
```
and with map:
```
***** Performance tests *****
Computing Erathostenes (imperative)...
Computing Erathostenes (imperative) took 3.3241519928 seconds.
Computing Erathostenes (recursive)...
Computing Erathostenes (recursive) took 3.35814499855 seconds.
```
The gain is thus negative... (I am pretty sure that I observed a gain in the past...)

I guess for programs intensively accessing big records this should make the difference. Let's quantify this before merging.

toots:
Cool thanks. Yeah I don't doubt that this should be more efficient but I've learned to be cautious with premature optimizations. Also, getting a framework in place for measuring this now is guaranteed to be super helpful in the future.

smimram:
I added a new test (big-record) in order to observe the gain on big records and we still see an increase from 1.26365709305 sec to 1.36542487144 sec...

toots:
Looks like using `Hashtbl` also doesn't yield any improvement.. This is weird.

PR #1242, opened 2020-06-10
Add support for errors and optional arguments.

smimram:
This PR introduces the notion of error:
- it can be raised with `error.raise("message")`
- it can be catched with `error.catch({code}, fun(e) -> print("got error: " ^ e))`
We can thing of syntactic sugar later on. As you can see runtime errors are parametrized by an integer (that what's most simple for now). Uncatched exception will raise a toplevel error.

The main use of errors is for optional arguments. For instance, we don't want to always specify default value in `list.hd` and raise an exception if the list is empty and we don't have a default value. For this to work without too many changes, I had to add _optional values_. A value of type `int?` is either nothing (which can be generated with `maybe.nothing()`) or an integer (a plain integer, not "some integer" as in OCaml). I use this to have `nothing` as default value for `list.hd` for instance (as well as other). All this should of course be type safe.

Comments are welcome!


src/lang/lang_values.ml
```diff
@@ -212,6 +215,7 @@ and in_term =
   | Encoder of Encoder.format
   | List of term list
   | Tuple of term list
+  | Nothing
```
toots, line 218: Since `Nothing` is essentially a ground value, should we register it via `Lang.MkAbstract`? I can see arguments in both sides.

smimram: I'll let you decide on that, I am not so familiar with abstract values. The only thing which makes this act weird is the subtyping `'a <: `'a?`, but I don't know if this is relevant here.

smimram:
I am completely in favor of adding a new syntax for this. I did not do it because I was lazy, plus we should have a conversation about which syntax extensions we want to put here: I also see `while` and `for` loops... Maybe in a second time?

For the abstract error, this is a nice idea, but won't it make it complicated for users to handle errors?

src/lang/lang_values.ml
toots: I think it needs to be in the main language file since it is a global definition that should be clearly visible to all callers of this API.

toots:
> I am completely in favor of adding a new syntax for this. I did not do it because I was lazy, plus we should have a conversation about which syntax extensions we want to put here: I also see `while` and `for` loops... Maybe in a second time?
> 
> For the abstract error, this is a nice idea, but won't it make it complicated for users to handle errors? Isn't the string enough?

I think error classes wouldn't be too complicated to use if we have a dedicated syntax. The reason I think they would be needed if b/c otherwise we're stuck with analyzing string content to know which kind of error are being caught, which is weak, and, also, each catch blocks interrupts any possible error, which is not a good extensive pattern to surface internal errors etc.

src/lang/builtins_error.ml
```diff
@@ -22,8 +22,28 @@
...
 
+let () =
+  add_builtin "error.register" ~cat:Liq
+    ~descr:"Return an error of the given type"
+    [("", Lang.string_t, None, Some "Type of the error")] Error.t (fun p ->
+      let s = Lang.to_string (List.assoc "" p) in
+      Error.to_value s)
+
+let () =
+  add_builtin "error.type" ~cat:Liq ~descr:"Get the type of an error"
```
smimram, line 43: I would use "kind" instead of "type" in order to avoid confusion with the typing system.

src/lang/lang.ml
```diff
@@ -162,11 +162,7 @@ let metadata m =
 
 let compare_values a b =
   let rec aux = function
-    | Ground (Ground.Float a), Ground (Ground.Float b) -> compare a b
-    | Ground (Ground.Int a), Ground (Ground.Int b) -> compare a b
-    | Ground (Ground.String a), Ground (Ground.String b) -> compare a b
-    | Ground (Ground.Bool a), Ground (Ground.Bool b) -> compare a b
-    | Ground a, Ground b -> compare (Ground.to_string a) (Ground.to_string b)
+    | Ground a, Ground b -> Ground.compare a b
```
smimram, line 165: Yay! :)

smimram: I am actually worried: shouldn't we use `demeth` in order to remove methods before comparing the values?

toots: I believe the type is `Ground of Ground.t` so there shouldn't be any value there.

smimram: Nope, and integer decorated with methods is considered as a valid integer. See: https://github.com/savonet/liquidsoap/blob/master/src/lang/lang.ml#L498
I would do as suggested below for safety (or at least ensure mentally that we won't use it on values with methods, but I don't think that this is ensured).

toots:
Ok, I have changed the following:

### Maybe:
*  Added `maybe.something` (useful for tests)
* Removed `maybe.case`, replaced by new syntax: `x | y`, which returns `y` when `x` is nothing and `x` value otherwise. This will be very intuitive for a lot of programers!

### Errors
* Added a generic `error` variable, with kind and message, registrable with `error.register`
* Changed `error.raise` to type: `(error, string) -> 'a`, where `<string>` is the error message. Messages can only be set via `error.raise`.
* Added `error.kind` and `error.message`

### try/catch
* `try <expr> catch <var> do <expr> done` which catches all errors
* `try <expr> catch <var>: [error] do <expr> done` , which catches all errors with same kind as at least one error in the list.

### Some tests:
```
success = ref(true)

def t(x, y)
  if x != y then
    print("Failure: got #{x} instead of #{y}")
    success := false
  end
end

def f() =
  # Works as expected with no errors
  ret = try
    4
  catch _ do
    5
  end

  t(ret, 4)

  e = error.register("foo")
  t(e, error.register("foo"))

  # Works as expected with no errors
  ret = try
    4
  catch _: [e] do
    5
  end

  t(ret, 4)

  # Can report kind
  ret = try
    error.raise(e);
    "bla"
  catch err do
    error.kind(err)
  end
  t(ret, "foo")

  # Can report empty message
  ret = try
    error.raise(e)
    "bla"
  catch err do
    error.message(err) | "blo"
  end
  t(ret, "blo")

  # Can report set message
  ret = try
    error.raise(e, "msg");
    "bla"
  catch err do
    error.message(err) | "blo"
  end
  t(ret, "msg")

  e' = error.register("bla")
  t(false, (e == e'))

  # Ignores errors when not in list
  ret =
    try
      try
        error.raise(e, "msg");
        "bla"
      catch err: [e'] do
        "blo"
      end
    catch err: [e] do
      "gni"
    end
  t(ret, "gni")

  # Ignore errors when list is empty
  ret =
    try
      try
        error.raise(e, "msg");
        "bla"
      catch err: [] do
        "blo"
      end
    catch err: [e] do
      "gni"
    end
  t(ret, "gni")

  # Catches error when in list
  ret =
    try
      try
        error.raise(e, "msg");
        "bla"
      catch err: [e, e'] do
        "blo"
      end
    catch err: [e] do
      "gni"
    end
  t(ret, "blo")

  if !success then test.pass() else test.fail() end
end

test.check(f)
```

src/lang/lang.ml
toots: Hmm. I really do believe this cannot happen. The values must have been `demeth`'d before reaching this function.

The `Ground` variant is defined via:
```
module Term = Lang_values
...
module Ground = Term.Ground
open Ground
...
```
And, in turn:
```
module Ground = struct
  type t = ..
...
end
```
That type never contains values. Likewise, `Ground.compare` only deals with values of type `Ground.t`.

That's also the type returned by the function you pointed to:
```ocaml
let to_int t =
  match (demeth t).value with Ground (Int s) -> s | _ -> assert false
```

src/lang/builtins_maybe.ml
```diff
@@ -31,21 +31,22 @@ let () =
...
     ]
-    b
+    a
     (fun p ->
       let x = Lang.assoc "" 1 p in
       let d = Lang.assoc "" 2 p in
-      let f = Lang.assoc "" 3 p in
-      let f = Lang.to_fun f in
-      match Lang.to_option x with None -> d | Some x -> f [("", x)])
+      match Lang.to_option x with None -> Lang.apply d [] | Some x -> x)
```
smimram, line 52: This is not the proper eliminator, please keep the `case` as it was...

src/lang/lang_parser.mly
```diff
@@ -293,6 +296,25 @@ expr:
   | BEGIN exprs END                  { $2 }
   | FUN LPAR arglist RPAR YIELDS expr{ mk_fun ~pos:$loc $3 $6 }
   | LCUR exprss RCUR                 { mk_fun ~pos:$loc [] $2 }
+  | expr LBAR expr                   { let maybe = mk ~pos:$loc($1) (Var "maybe") in
```
smimram, line 299: Again, I don't like much this syntax and this is not a proper `case` that you are using here :)

src/lang/lang_values.ml
```diff
@@ -1059,8 +1061,8 @@ and apply f l =
   (* Record error positions. *)
   let f pe =
     try f pe with
-      | Runtime_error (poss, e) ->
-          raise (Runtime_error (Option.to_list pos @ poss, e))
+      | Runtime_error err ->
+          raise (Runtime_error { err with pos = Option.to_list pos })
```
smimram, line 1065: No, no, no, the idea was that we keep the stacktrace! Revert this!

src/lang/builtins_error.ml
```diff
@@ -22,49 +22,82 @@
...
+      let { kind } = Error.of_value (Lang.assoc "" 1 p) in
+      let msg =
+        Utils.maybe Lang.to_string (Lang.to_option (Lang.assoc "" 2 p))
+      in
+      raise (Lang_values.Runtime_error { Lang_values.kind; msg; pos = [] }))
 
 let () =
   let a = Lang.univ_t () in
-  add_builtin "error.catch" ~cat:Liq
+  add_builtin "error.catch" ~cat:Liq ~flags:[Lang.Hidden]
```
smimram, line 81: There is no shame in using this function!
```suggestion
  add_builtin "error.catch" ~cat:Liq
```

src/lang/builtins_error.ml
```diff
@@ -22,49 +22,82 @@
...
     ~descr:"Execute a function, catching eventual exceptions."
     [
-      ("", Lang.fun_t [] a, None, Some "Function to execute.");
-      ( "",
-        Lang.fun_t [(false, "", Lang.string_t)] a,
+      ( "errors",
+        Lang.list_t Error.t,
         None,
-        Some "Error handler." );
+        Some "Kinds of errors to catch. Catches all errors if empty." );
```
smimram:
```suggestion
        Some "Kinds of errors to catch. );
```
I am not so fond of non-monotonous behaviors like this. Someday someone will use this lis in a complicated way and will have bugs because the empty list is not the expected limit case. I would rather have a list option here and have `nothing` catching all.

toots: Just did it!

src/lang/builtins_error.ml
toots: This is the same pattern as is done for `if .. then .. else .. end`. The syntax uses a runtime function and the function is hidden from the API. Also, in this case `catch` is now a reserved work do ya can't call `error.catch` anyways.

src/lang/lang_values.ml
toots: Hmm. I must be missing something here. All I'm intending to do is structure the arguments, more than 3 I like to use records. I think you're referring to dropping the `@ poss`, I'll add it back.

src/lang/lang_parser.mly
toots: I really do actually. Couple of reasons:
* The syntax is consistent with similar patterns in dynamic languages like Javascript. This will make adoption easier.
* The syntax is short
* The left side is implicitly scoped, which avoids unnecessary `{<expr>}`
* It is a case, `x | y` is the same as, in OCaml:
```ocaml
match x with
  | Some v -> v
  | None -> y
```

src/lang/lang_values.ml
smimram: Yes, the `@`!

src/lang/builtins_error.ml
```diff
@@ -22,49 +22,82 @@
...
 let () =
-  add_builtin "error.type" ~cat:Liq ~descr:"Get the type of an error"
+  add_builtin "error.kind" ~cat:Liq ~descr:"Get the kind of an error"
     [("", Error.t, None, None)] Lang.string_t (fun p ->
-      Lang.string (Error.of_value (List.assoc "" p)))
+      Lang.string (Error.of_value (List.assoc "" p)).kind)
+
+let () =
+  add_builtin "error.message" ~cat:Liq ~descr:"Get the message of an error"
+    [("", Error.t, None, None)] (Lang.maybe_t Lang.string_t) (fun p ->
```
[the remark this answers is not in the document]
toots: I disagree. With `nothing` we should on the contrary clean up the API in all places where we assume any kind of semantics for the empty string or empty list and use `nothing` instead. I believe you just mentioned a similar change for the empty list below.

libs/errors.liq
```diff
@@ -0,0 +1 @@
+let error.nothing = error.register("nothing")
```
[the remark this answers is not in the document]
toots: Yeah, this is more explicit I agree.

src/lang/builtins_maybe.ml
```diff
@@ -0,0 +1,52 @@
...
+ *****************************************************************************)
+
+open Lang_builtins
+
+let () = Lang.add_module "maybe"
+
+let () =
+  add_builtin "maybe.nothing" ~cat:Liq ~descr:"Nothing." []
+    (Lang.maybe_t (Lang.univ_t ()))
+    (fun _ -> Lang.nothing)
```
smimram, line 30: I am actually thinking that it could be a good idea to call it `null` as in other languages?
```suggestion
  add_builtin "maybe.null" ~cat:Liq ~descr:"Nothing." []
    (Lang.maybe_t (Lang.univ_t ()))
    (fun _ -> Lang.null)
```

libs/list.liq
```diff
@@ -16,7 +16,7 @@ end
 # @category List
 # @param ~default Default value if key does not exist.
 def list.hd(~default=maybe.nothing(), l)
-  list.dcase(l, {maybe.raise(default, "no default value for list.hd")}, fun (x, _) -> x)
+  list.dcase(l, {default | error.raise(error.nothing, "no default value for list.hd")}, fun (x, _) -> x)
```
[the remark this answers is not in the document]
toots: As said below, this is a pattern from dynamically typed languages.

src/lang/lang_parser.mly
smimram: I agree with that, but I don't think it is particularly intuitive (I had to look up in the grammar to understand the intended meaning). Do you think that it will be use that often? Otherwise, I'd rather leave something more explicit...

src/lang/builtins_maybe.ml
toots: Yes!

toots: Why not at top level then?

src/lang/lang_parser.mly
toots: I plan on using it a bunch yes 🙂

smimram: Last question before I declare you won this one: can I have a reference where this is used in JS? Could not find it :(

toots: My bad it's double `||` not a single one in JS: https://pietschsoft.com/post/2008/10/14/javascript-gem-null-coalescing-using-the-or-operator

Looks like there's a new operator being worked on there, via '??`: https://medium.com/better-programming/how-to-use-the-null-coalescing-operator-in-javascript-15adefe7c655

Would you rather use this one? I prefer the `|` or `||` but following the same syntax could be beneficial in the long run.

smimram: I find `??` much better: it makes it clear that it applies to an option value. The rationale behind `||` was that is overloaded the disjunction, but since it is written `or` in Liq it does not make much sense here.

toots: Let's do it!

CHANGES.md
```diff
@@ -3,6 +3,8 @@
 
 New:
 
+- Add support for errors with `error.*` (#1242).
+- Add support for optional values with `null.*` (#1242).
```
toots, line 7: Let's mention the syntax changes here too.

libs/null.liq
```diff
@@ -0,0 +1,5 @@
+# Determine whether a nullable value is null.
+# @category Liquidsoap
+def null.is(x) =
```
toots, line 3: I think a better name would be `null.check`

src/lang/lang_lexer.ml
```diff
@@ -162,6 +162,9 @@ let rec token lexbuf =
     | "ref" -> REF
     | "def" -> PP_DEF
     | "replaces" -> REPLACES
+    | "try" -> TRY
+    | "catch" -> CATCH
+    | "do" -> DO
```
toots, line 167: Aren't there some other files to edit when adding new syntax? I saw some in your other PR, I'm just not familiar with this.

src/lang/lang_parser.mly
```diff
@@ -293,6 +296,25 @@ expr:
   | BEGIN exprs END                  { $2 }
   | FUN LPAR arglist RPAR YIELDS expr{ mk_fun ~pos:$loc $3 $6 }
   | LCUR exprss RCUR                 { mk_fun ~pos:$loc [] $2 }
+  | expr COALESCE expr               { let null = mk ~pos:$loc($1) (Var "null") in
+                                       let op =  mk ~pos:$loc($1) (Invoke (null, "default")) in
+                                       let handler = mk_fun ~pos:$loc($3) [] $3 in
+                                       mk ~pos:$loc (App (op, ["",$1;"",handler])) }
+  | TRY exprs CATCH bindvar COLON varlist DO exprs END
```
toots, line 303: If you're planning on adding `in` in the other PR, we should switch to `in` here too instead of `:`. We can do that after both are merged.

tests/language/null.liq
```diff
@@ -0,0 +1,18 @@
...
+
+%include "test.liq"
+
+def t(x,y) =
+  if x != y then test.fail() end
+end
+
+def f() =
+  t(null() ?? "bla", "bla")
+  t(null("foo") ?? "bla", "foo")
```
toots, line 11: I'm a little bit worried about this being potentially confusing but let's go with it for now.

src/lang/lang_lexer.ml
smimram: liquidsoap.xml in the documentation (syntax coloration) and the emacs mode.

libs/null.liq
toots: giphy (1)

src/lang/lang_parser.mly
smimram: good idea, in a second time.

PR #1248, opened 2020-06-16
Don't try to find sources in methods.

### LINKED ISSUE #1247: liquidsoap 1.5.x hangs with crossfade
**Describe the bug**
As soon as you use crossfading, (I haven't been able to determine if smart=true is required) in any script, liquidsoap will hang, even with a simple --check. Nothing is printed to the terminal.
**To Reproduce**
This script is enough, no matter if you run liquidsoap or liquidsoap --check:

```
source = mksafe(crossfade(smart=true,sine()))
output.file(%mp3(bitrate=320),"test.mp3",source)
```

**Expected behavior**
Liquidsoap should start streaming or at least print text to the terminal.

**Version details**
 - OS: ArchLinux
 - Version 1.5.0+git@493557d1

**Install method**
Installed via opam.

smimram:
Otherwise the complexity of `iter_sources` becomes too big.

Fixes #1247.

### DIFF
--- a/src/lang/lang.ml
+++ b/src/lang/lang.ml
@@ -392,7 +392,12 @@ let iter_sources f v =
       | List l -> List.iter iter_value l
       | Tuple l -> List.iter iter_value l
       | Meth (_, a, b) ->
-          iter_value a;
+          (* WARNING: we assume that there are no sources in methods. Otherwise,
+             since fields are defined by overriding previous values, the checking
+             ends up being quadratic and unusable in practice with out large
+             standard library, see #1247. *)
+          (* iter_value a; *)
+          ignore a;
           iter_value b
       | Fun (proto, pe, env, body) ->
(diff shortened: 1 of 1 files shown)


src/lang/lang.ml
```diff
@@ -392,7 +392,12 @@ let iter_sources f v =
       | List l -> List.iter iter_value l
       | Tuple l -> List.iter iter_value l
       | Meth (_, a, b) ->
-          iter_value a;
+          (* WARNING: we assume that there are no sources in methods. Otherwise,
+             since fields are defined by overriding previous values, the checking
+             ends up being quadratic and unusable in practice with out large
```
toots, line 397:
```suggestion
             ends up being quadratic and unusable in practice with our large
```

smimram:
I actually found a clean solution: simply avoid checking multiple times the same value.

PR #1252, opened 2020-06-17
Add support for for and while loops.

smimram:


doc/liquidsoap.xml
```diff
@@ -26,6 +26,11 @@
       <item>then</item>
       <item>else</item>
       <item>elsif</item>
+      <item>while</item>
+      <item>for</item>
+      <item>to</item>
+      <item>do</item>
+      <item>done</item>
```
toots, line 33: When working on the syntax for errors, I realized that we should be careful introducing too many reserved keywords. I was trying to use `in` for `try ... catch err in [errs] do ... end` and `in` was already wildly used (`in()`, `fade.in` etc.)

At the very least we should re-use `end` instead of `done` and perhaps also refrain from adding too many intermediary words, for instance:
```
for n = 1 .. n do
  [...]
end
```

Another thing I noticed reading from `go` is that they re-use `for` for `while` loops. I think I prefer `while` but just wanted to note it.

libs/loop.liq
```diff
@@ -0,0 +1,26 @@
...
+# @param cond Boolean condition
+# @param f Function to be called
+# @cateogry Liquidsoap
+def rec while(cond, f)
+  c = to_bool_getter(cond)
+  if c() then
+    f()
+    while(cond, f)
+  end
+end
```
toots, line 26: Why should we have both the functional and syntactic API?

libs/predicate.liq
```diff
@@ -34,13 +34,13 @@ end
 # @flag extra
 # @param p Predicate.
 def predicate.once(p)
-  done = ref(false)
+  finished = ref(false)
```
toots, line 37: See...

libs/loop.liq
smimram: This file was meant to go away.

doc/liquidsoap.xml
smimram: Alright, I changed the `do` to `end`. I am not fond of the `..` syntax: I would be alright to add it if we had proper iterators (which we don't for now :))

toots: Cool. I was just thinking, why don't we make the iteration more functional and generic? Something like:
```
for n = array do
   (...)
end
```

Then add a syntax for:
```
x = 1..5 // [1, 2, 3, 4, 5]
```

I'd keep the syntax for brackets away because we might want to do this at some point:
```
x = [2,3]
x = [1, ...x] // [1,2,3]
```

I think `1..5` syntax would be useful to our users as well.

smimram: I had this in mind too, I was lazy but since you propose that I'll dive into it. I propose
- to add a constraint `Iterable` for iterable types (lists, etc.)
- the syntax should be `for i in l do ... end` (I'd rather have `in` than `=`)
- I don't like the idea of building a list, I think that `1..1000` would generate something like [Seq.t](https://caml.inria.fr/pub/docs/manual-ocaml/libref/Seq.html) in OCaml, i.e. a function of type `() -> int?` which returns the integers from 1 to 1000 and then `null` (this avoids uselessly building the list)
Do you like it?

toots: I'm down with the following remarks:
* Do we really need `iterable`? Couldn't we just start with `() -> 'a?` for `for` and `() -> true` for while?
* I have nothing against using `in` per-say but what will you do with `fade.in`?

smimram: Yes we could start with `() -> 'a?` but it would be nice to be able to iterate over a list without having to convert it first...

I think I should be able to keep `fade.in` with some easy parser tweaking.

doc/content/language.md
```diff
@@ -95,7 +95,9 @@ You can form expressions by using
   without `def .. end`, and in the body of anonymous functions, for example `fun
   (x) -> f1(x) ; f2(x)` will be read as `(fun (x) -> f1(x)) ; f2(x)` not as `fun
   (x) -> begin f1(x) ; f2(x) end`,
-- _code blocks_: `{ expr }` is a shortcut for `fun () -> expr`.
+- _code blocks_: `{ expr }` is a shortcut for `fun () -> expr`,
+- _loops_: the syntax for "for" and "while" loops is `for i = first to last do
```
toots, line 99: Needs update!

src/lang/lang.ml
```diff
@@ -138,10 +138,7 @@ let record = meth unit
 let source s = mk (Source s)
 let request r = mk (Ground (Request r))
 let reference x = mk (Ref x)
-
-let val_fun p f =
-  let p' = List.map (fun (l, x, d) -> (l, x, d)) p in
```
toots, line 143: Huh..

src/lang/lang_builtins.ml
```diff
@@ -702,3 +702,75 @@ let () =
       print_string v;
       flush stdout;
       Lang.unit)
+
+(** Loops. *)
+
+let () =
+  add_builtin "while" ~cat:Liq ~descr:"A while loop."
```
toots, line 709: Flags hidden!

src/lang/lang_builtins.ml
```diff
@@ -702,3 +702,75 @@ let () =
...
+    (fun p ->
+      let c = Lang.to_bool_getter (Lang.assoc "" 1 p) in
+      let f = Lang.to_fun (Lang.assoc "" 2 p) in
+      while c () do
+        ignore (f [])
+      done;
+      Lang.unit)
+
+let () =
+  add_builtin "for" ~cat:Liq ~descr:"A for loop."
```
toots, line 724: Flags hidden!

src/lang/lang_parser.mly
```diff
@@ -296,11 +298,14 @@ expr:
...
   | LCUR exprss RCUR                 { mk_fun ~pos:$loc [] $2 }
+  | WHILE expr DO exprs END          { mk ~pos:$loc (App (mk ~pos:$loc($1) (Var "while"), ["", mk_fun ~pos:$loc($2) [] $2; "", mk_fun ~pos:$loc($4) [] $4])) }
+  | FOR VAR IN expr DO exprs END     { mk ~pos:$loc (App (mk ~pos:$loc($1) (Var "for"), ["", $4; "", mk_fun ~pos:$loc($6) ["", $2, T.fresh_evar ~level:(-1) ~pos:(Some $loc($2)), None] $6])) }
+  | expr DOTDOT expr                 { mk ~pos:$loc (App (mk ~pos:$loc($2) (Invoke (mk ~pos:$loc($2) (Var "iterator"), "int")), ["", $1; "", $3])) }
   | expr COALESCE expr               { let null = mk ~pos:$loc($1) (Var "null") in
                                        let op =  mk ~pos:$loc($1) (Invoke (null, "default")) in
                                        let handler = mk_fun ~pos:$loc($3) [] $3 in
                                        mk ~pos:$loc (App (op, ["",$1;"",handler])) }
-  | TRY exprs CATCH bindvar COLON varlist DO exprs END
+  | TRY exprs CATCH bindvar IN varlist DO exprs END
```
toots, line 308: Thank you!

src/lang/lang_parser.mly
```diff
@@ -453,6 +458,10 @@ replaces:
   | { false }
   | REPLACES { true }
 
+var:
+  | VAR { $1 }
+  | IN  { "in" }
```
toots, line 463: Woo I see.

tests/language/loop.liq
```diff
@@ -0,0 +1,42 @@
...
+  
+def f() =
+  n = ref(1)
+  while !n < 10 do
+    n := !n * 2
+  end
+  t(!n, 16)
+
+  n = ref(0)
+  for i in 11 do
```
toots, line 22: Do we really want to make `int` iterable? This looks weird to me, I think it'll be confusing.

smimram: I have removed the `Iterable` type constraint for now, it was buggy... Support for iterating over integers has gone too for now!

PR #1265, opened 2020-06-24
Add support for opening modules.

### LINKED ISSUE #1259: Implement module open

smimram:
Fixes #1259.

For now, we restrict the main value to be unit, do we want more generality?


src/lang/lang_parser.mly
```diff
@@ -182,6 +182,7 @@
 %token COALESCE
 %token TRY CATCH DO
 %token IF THEN ELSE ELSIF
+%token OPEN
```
toots, line 185: I think it's cool to add `open` as reserved word in this case.

src/lang/lang_values.ml
```diff
@@ -314,6 +315,7 @@ let rec print_term v =
     | Null -> "null"
     | Meth (l, v, e) -> print_term e ^ ".{" ^ l ^ " = " ^ print_term v ^ "}"
     | Invoke (e, l) -> print_term e ^ "." ^ l
+    | Open (m, e) -> "open " ^ print_term m ^ " in " ^ print_term e
```
toots, line 318: Is that really the implemented syntax?

src/lang/lang_parser.mly
```diff
@@ -243,11 +244,10 @@ g: | {} | GETS {}
...
 exprs:
   | expr s                   { $1 }
-  | expr exprs               { mk ~pos:$loc (Seq ($1,$2)) }
-  | expr SEQ exprs           { mk ~pos:$loc (Seq ($1,$3)) }
+  | expr s exprs             { mk ~pos:$loc (Seq ($1,$3)) }
   | binding s                { mk_let ~pos:$loc($1) $1 (mk ~pos:$loc unit) }
-  | binding exprs            { mk_let ~pos:$loc($1) $1 $2 }
-  | binding SEQ exprs        { mk_let ~pos:$loc($1) $1 $3 }
+  | binding s exprs          { mk_let ~pos:$loc($1) $1 $3 }
+  | OPEN expr s exprs        { mk ~pos:$loc (Open ($2,$4)) }
```
toots, line 250: Should we restrict `OPEN` to `var` only?

src/lang/lang_values.ml
```diff
@@ -750,6 +754,17 @@ let rec check ?(print_toplevel = false) ~level ~(env : T.env) e =
                 x
         in
         e.t >: aux a.t
+    | Open (a, b) ->
+        check ~level ~env a;
+        a.t <: mk T.unit;
```
toots, line 759: I think it makes sense to restrict `open` to `unit` modules for now. We can change that if we ever see a need for it but it seems nice to say `unit` modules are "proper" modules, non-unit modules are decorated values.

src/lang/lang_parser.mly
smimram: Nope, we want to be able to open `a.b` for instance, so let's not add arbitrary restrictions.

PR #1266, opened 2020-06-24
Add support for `x ? y : z`

toots:
Fixes https://github.com/savonet/liquidsoap/issues/1263


toots:
> I am fine with the PR itself, but do we really need to add such a syntax (which might conflict with other things we might want to add later on)?

I use this pattern a lot in other languages and I really believe this is something that is missing in ours (and in OCaml) and that will make adoption easier.

What kind of conflict do you envision?

smimram:
Nothing particular in mind, just a general comment. If you have use for it, go ahead!

Issue #1271, opened 2020-06-30
Rename clock relationships


toots:
Avoid using `master`/`slave` terminology in our software. Same goes for `git` branches.

Issue #1256, opened 2020-06-22
Memory Leak


stimpy23:
**Describe the bug**
I think, it's the same problem as described in https://github.com/savonet/liquidsoap/issues/1226
but since I'm not sure, I'm starting a new issue...
I'm currently in the process of moving from some self-written scripts to AzuraCast for broadcasting our radiostation.
All is working fine until it doesn't. Then, liquidsoap starts eating up all the memory, it can.
When it happens, it stops logging, strace doesn't give any output but:
`mmap(NULL, 737374208, PROT_READ|PROT_WRITE, MAP_PRIVATE|MAP_ANONYMOUS, -1, 0) = 0x7fa6d2da3000` every minute or so, which makes debugging pretty hard for me.. ;)
Only way to make it stop is deleting all media from AzuraCast.

First time around, I just added everything at once: Some AzuraCast managed playlists, some managed by liquidsoap itself (due to special options "Play 1 track only" and "Start at the exact time" in AzuraCast). It all worked fine until I was finished, then the leakage began.
So I deleted all the media and started over again.
This time, I added 10 playlists at a time, rebooted the container and waited for the leakage to begin. It never did, until I was done adding all the playlists, again. Last playlists where reruns of shows, this time without special options, so AzuraCast manages the playlists. Still: 5 Minutes later, the leakage began again.

This time, I tried to remove only these rerun-playlists: Nothing. Then, I took a look at the liquidsoap.liq file, AzuraCast is generating, and I saw it didn't remove the rerun playlists from the .liq file, yet. So I did it manually. Still no luck.

AzuraCast (and therefore liquidsoap) is running in a Linux Container with Ubuntu Bionic. Normally the container has 2CPUs and 4GB of RAM, but while debugging, I already tried giving it up ro 32GB RAM and 4 CPUs. No difference: It just eats up all the RAM whithin a minute, then it starts swapping.
Funny enough: It's only swapping up to 2Gigs, no more. But it's constantly reading/writing these 2 gigs...

This is, where I'm stuck at the moment.. Already asked the AzuraCast developer for help, but he sent me here, since it's liquidsoap causing the trouble:
https://github.com/AzuraCast/AzuraCast/issues/2970

**To Reproduce**
I honestly have no idea..! I guess adding a lot of media could help.. ;)
What I know by now is, that it doesn't matter if you're using internal or external playlists and that it doesn't happen immediately.
[...]

Issue #1122, opened 2020-03-25
Memory leak on windows using telnet commands


ignaciovenezia:
Hello!
I've had this issue of liquidsoap memoryleaking for a while and tried to fix it to no avail. Whilst trying to pinpoint why was I having memory issues, I've come to the conclusion that using telnet commands, on windows, is the cause.

Why is this I'm not sure, I'm quite new to liquidsoap so know for sure.

Testing script to replicate: https://pastebin.com/VhdjAqus

I interact with liquidsoap through telnet from my own application. I have a service that sends the tcp commands "decoder.status" and the custom command "lm" every second to check the status of the decoder and the latest title provided by the notify_metadata procedure. I then manage a secondary process based off the result of these tcp commands.

If my approach is very wrong, I ask of you to guide me in a better direction. 

Anyway, whilst searching for an answer I stumbled upon issue #622 in which in a response @toots asked this very question, so I guess there must be something to it.

Some more information: 
- Liquidsoap version 1.4.0
- OS: on both Windows 7 and Windows 10 I've had the same issue
- Memory usage increases when calling any telnet command, like "help" for example, not just custom commands.

[It can get very wild](https://imgur.com/a/O8icYja)

If this is entirely my fault, don't hesitate to call me on it. Tough love works.

Thanks.

toots:
Thanks for this report. I'll have another try at it.

ignaciovenezia:
> Thanks for this report. I'll have another try at it.

Thanks for looking into it. Whilst looking for a non-memoryleak way of interacting with the server I found out that using http harbor also leaks in the same way.

`# Return the json content
# of meta
def get_meta(~protocol,~data,~headers,uri) =
  m = "asd"
  http_response(
    protocol=protocol,
    code=200,
    headers=[("Content-Type","application/json; charset=utf-8")],
    data=json_of(m)
  )
end

# Register get_meta at port 700
harbor.http.register(port=7000,method="GET","/getmeta",get_meta)` 

Using this, which is pretty much the code on the docs, every time I do an http request to localhost:7000/getmeta memory usage  increases and it never goes down.

I tried calling 'garbage_collect()' in the script in many different locations with no effect whatsoever in the memory usage.

pranza:
hi!
[...]

Issue #1108, opened 2020-03-11
Liquidsoap error after AzuraCast update


dbeiner:
**Describe the bug**
After update AzuraCast, Liquidsoap won't start

**To Reproduce**
Update AzuraCast

**Expected behavior**
Help me to fix the issue

**Version details**
Ubuntu 18.04 LTS

**Install method**
AzuraCast on Docker

**Common issues**
```
Fatal error: exception Stack overflow
Raised by primitive operation at file "source.ml", line 209, characters 46-51
Called from file "list.ml", line 110, characters 12-15
Called from file "camlinternalOO.ml", line 370, characters 12-17
Called from file "camlinternalOO.ml", line 380, characters 24-40
Called from file "lang/lang.ml", line 409, characters 30-37
Called from file "lang/lang_values.ml", line 1027, characters 14-20
Called from file "lang/lang_values.ml", line 1090, characters 17-30
Called from file "lang/lang.ml", line 654, characters 14-38
Called from file "tools/tutils.ml", line 83, characters 14-18
Re-raised at file "tools/tutils.ml", line 83, characters 48-55
Called from file "lang/lang_errors.ml", line 67, characters 8-12
Re-raised at file "lang/lang_errors.ml", line 155, characters 48-55
Called from file "lang/lang.ml", line 671, characters 8-163
Called from file "lang/lang.ml", line 678, characters 4-76
Called from file "main.ml", line 125, characters 34-40
Called from file "main.ml", line 668, characters 6-16
Called from file "runner.ml", line 30, characters 14-31
Line 211, char 15 before "# Remote Relays":
Error -1: Unknown error
```
Thank you !

dbeiner:
https://github.com/AzuraCast/AzuraCast/issues/2495

smimram:
Il would help if we had the script (in particular around line 211...)

dbeiner:
Hi @smimram 

I've asked AzuraCast team to help me to give it to you...

dbeiner:
Hi @smimram 

I've try to restart LiquidSoap and the error is a little bit most shortly, "only" : 

×Liquidsoap encountered an error
Re-raised at file "lang/lang_errors.ml", line 155, characters 48-55
Called from file "lang/lang.ml", line 671, characters 8-163
Called from file "lang/lang.ml", line 678, characters 4-76
Called from file "main.ml", line 125, characters 34-40
Called from file "main.ml", line 668, characters 6-16
[...]

PR #1272, opened 2020-06-30
Avoid exponentially growing memory allocation when unifying clocks.

toots:
This PR is a follow-up for https://github.com/savonet/liquidsoap/issues/1108 and a fix for the underlying issue, as reported in https://github.com/savonet/liquidsoap/issues/1256

## How to reproduce?

The following script exhibits the issue with release `1.4.2` and, most likely, also with the current `master`:

```
s = sine()
s = cue_cut(s)

def rec f(l) =
  if list.length(l) == 30 then
    l
  else
    f(list.add((random.bool, s), l))
  end
end

s = switch(f([]))

output.dummy(fallible=true,s)
```

## Root cause

The root cause of the issue is an exponential increase in the list of sub-clocks associated with a given unknown clock. This list is compiled during clock unification and the unification algorithm was missing two safe guards:
- Being careful to not unify a variable against itself
- Being careful to filter out the list to only keep one instance of each subclock

## Reproduction scenario

The operator `cue_cut` requires a full control of its underlying source. For this purpose, it initiates a slave sub clock (and soon to be renamed to avoid `master`/`slave` terminology: https://github.com/savonet/liquidsoap/issues/1271) as follows:
```ocaml
    method private set_clock =
      let slave_clock = Clock.create_known (new Clock.clock self#id) in
      (* Our external clock should stricly contain the slave clock. *)
      Clock.unify self#clock
        (Clock.create_unknown ~sources:[] ~sub_clocks:[slave_clock]);
      (* The source must belong to our clock, since we need occasional
       * control on its flow (to seek at the beginning of a track). *)
      Clock.unify slave_clock source#clock;
      (* When this source disappears the slave clock becomes useless.
       * To allow for its collection, remove references to it as a subclock
       * of the master clock. *)
      Gc.finalise (fun self -> Clock.forget self#clock slave_clock) self
```
This introduces a first sub-clock in the list of dependent clocks for the `rotate` operator, during the first unification.

In turn, the `rotate` operator unifies its own clock with its sibling's clocks as follows:
```ocaml
    method private set_clock =
      List.iter (fun s -> unify self#clock s#clock) sources
```
Which is where the exponential patten occurs. When unifying two unknown clock, the code goes as:
```ocaml
    | ( Link ({ contents = Unknown (sa, ca) } as ra),
        Link ({ contents = Unknown (sb, cb) } as rb) ) ->
        occurs_check a b;
        let merge =
          (* Using List.rev_append to remain tail-recursive, see #1108. *)
          Link (ref (Unknown (List.rev_append sa sb, List.rev_append ca cb)))
        in
        ra := Same_as merge;
        rb := Same_as merge
```

But, in our examples, all the siblings are the same exact source, so this ends up unifying each clock with itself over and over, doubling the length of the list of dependent clocks at each iteration, making it grow at a `2^n` rate:

```
set clock for rotate
s: 0, sc: 1
s: 0, sc: 2
s: 0, sc: 4
s: 0, sc: 8
s: 0, sc: 16
s: 0, sc: 32
s: 0, sc: 64
s: 0, sc: 128
s: 0, sc: 256
s: 0, sc: 512
s: 0, sc: 1024
s: 0, sc: 2048
s: 0, sc: 4096
s: 0, sc: 8192
s: 0, sc: 16384
s: 0, sc: 32768
s: 0, sc: 65536
s: 0, sc: 131072
s: 0, sc: 262144
s: 0, sc: 524288
s: 0, sc: 1048576
s: 0, sc: 2097152
s: 0, sc: 4194304
...
```

This behavior is not exhibited in the absence of `cue_cut` because in this case, the list of sub-clocks is empty so never starts growing.

## Post-mortem

The fix for #1108 had identified the faulty path, which was triggering a stack overflow. By making the list concatenation tail-recursive, it switched resources exhaustion away from the stack onto allocated resources, allowing the unification algorithm to survive a slightly better increase in sources. However, as with exponential growth, a larger `n` was bound to make it fail again, which is what got reported in #1256

## Fix

The proposed patch does two things that are both sensitive and have been confirmed to fix the issue separately:
- Make sure to not unify the same clock variable against itself
- Make sure to filter the associated lists to make sure they keep only one occurence of each item

## Release plans

This fix should be be released soon. A `1.4.3` release will probably need to be scheduled. Meanwhile, we have discussed plans with @SlvrEagle23 to let them use a tagged branch for their own production deployment. To that effect, a `1.4.3` bugfix branch should be pushed shortly so they can begin sending the fix to their users.

## Possibly related issues

It is possible that this bug also triggered memory leaks issues long-reported on win32 (see: https://github.com/savonet/liquidsoap/issues/1122), specially since the user's script seems to also some complicated situations with clocks (see: https://github.com/savonet/liquidsoap/issues/985). An early windows build would be great to check about this with @ignaciovenezia and @pranza

We should also investigate similarly reported issues with memory leaks such as https://github.com/savonet/liquidsoap/issues/541 for instance.


src/source.ml
```diff
@@ -205,13 +205,16 @@ let rec unify a b =
     | Known s, Known s' ->
         if s <> s' then
           raise (Clock_conflict (variable_to_string a, variable_to_string b))
+    | Link ra, Link rb when ra == rb -> ()
     | ( Link ({ contents = Unknown (sa, ca) } as ra),
         Link ({ contents = Unknown (sb, cb) } as rb) ) ->
         (* TODO perhaps optimize ca@cb *)
```
toots, line 211:
```suggestion
```

src/source.ml
```diff
@@ -205,13 +205,15 @@ let rec unify a b =
...
           raise (Clock_conflict (variable_to_string a, variable_to_string b))
+    | Link ra, Link rb when ra == rb -> ()
     | ( Link ({ contents = Unknown (sa, ca) } as ra),
         Link ({ contents = Unknown (sb, cb) } as rb) ) ->
-        (* TODO perhaps optimize ca@cb *)
         occurs_check a b;
         let merge =
+          let s = List.sort_uniq Stdlib.compare (List.rev_append sa sb) in
+          let sc = List.sort_uniq Stdlib.compare (List.rev_append ca cb) in
           (* Using List.rev_append to remain tail-recursive, see #1108. *)
```
toots, line 215:
```suggestion
```

src/test/source_test.ml
```diff
@@ -0,0 +1,16 @@
...
+
+let () =
+  let c1 = create_unknown ~sources:[] ~sub_clocks:[] in
+  let c2 = create_unknown ~sources:[] ~sub_clocks:[c1] in
+  let c3 = create_unknown ~sources:[] ~sub_clocks:[c1] in
+  (* Make sure unification of a variable with itself
+     works as expected. *)
+  unify c2 c2;
+  assert (subclocks c2 = [c1]);
+  (* Make sure subclocks don't get dubplicated during inification. *)
```
toots, line 13:
```suggestion
  (* Make sure subclocks don't get duplicated during inification. *)
```

pranza:
hello!
оhh, so you finally found something! the bug was being awakened by my 
fallback system, right? should i re-do my config to avoid it or should i 
wait for a [windows x64] build to resolve it?

regardsss,
p

SlvrEagle23:
@toots Thank you for continuing to look at this issue and working to resolve it! We do indeed have provisions in place now to pick a specific commit and pin it during our Docker image build process, so whenever new fixes are rolled out, we can incorporate them directly without needing them to be tagged as releases.

pavlos2:
Many thanks Romain and to the rest team for the great work and support! Looking forward to 1.4.3 release.

src/source.ml
```diff
@@ -205,13 +205,14 @@ let rec unify a b =
...
           raise (Clock_conflict (variable_to_string a, variable_to_string b))
+    | Link ra, Link rb when ra == rb -> ()
     | ( Link ({ contents = Unknown (sa, ca) } as ra),
         Link ({ contents = Unknown (sb, cb) } as rb) ) ->
-        (* TODO perhaps optimize ca@cb *)
         occurs_check a b;
         let merge =
-          (* Using List.rev_append to remain tail-recursive, see #1108. *)
-          Link (ref (Unknown (List.rev_append sa sb, List.rev_append ca cb)))
+          let s = List.sort_uniq Stdlib.compare (List.rev_append sa sb) in
```
smimram, line 213: Are lists the good data structure here? Since we maintain everything sorted, maybe should we use Sets instead, no?

src/source.mli
```diff
@@ -317,4 +317,5 @@ module Clock_variables : sig
   val forget : clock_variable -> clock_variable -> unit
   val get : clock_variable -> clock
   val is_known : clock_variable -> bool
+  val subclocks : clock_variable -> clock_variable list
```
smimram, line 320: Some doc please :)

src/source.ml
toots: Good idea.

src/source.mli
toots: This is for testing purposes for now so I'll just add that.. 😅

toots, line 320:
```suggestion

  (* This is exported for testing purposes only at the moment. *)
  val subclocks : clock_variable -> clock_variable list
```

toots:
> @toots Thank you for continuing to look at this issue and working to resolve it! We do indeed have provisions in place now to pick a specific commit and pin it during our Docker image build process, so whenever new fixes are rolled out, we can incorporate them directly without needing them to be tagged as releases.

Great! There's a `1.4.3` branch now that I'll use to collect the bugfix for that release. I think you should be able to push it for production, it's essentially `1.4.2` with this fix at the moment.

toots:
> Many thanks Romain, and to the rest team, for the great work and support! Looking forward to 1.4.3 release.

Thank for your patience helping us with the issue!

toots:
> hello! оhh, so you finally found something! the bug was being awakened by my fallback system, right? should i re-do my config to avoid it or should i wait for a [windows x64] build to resolve it? regardsss, p

I sure hope that this was your issue though I cannot guarantee but I have hopes.. I'll make an early build for testing shortly.

arusi01:
Thanks!

SlvrEagle23:
AzuraCast has finished building an updated `stations` image with this update pulled from the 1.4.3 development branch of Liquidsoap. You can now update your installations to take advantage of the updated code.

pavlos2:
Thanks a lot Buster! Looking forward to this AzuraCast update.

pavlos2:
I updated AzuraCast, still see Liquidsoap 1.4.2

SlvrEagle23:
@pavlos2 This is because they haven't tagged it as a new release yet, it's a development branch that's branched off of 1.4.2. As long  as you're updated to the latest AzuraCast version, the fix should be applied.

toots:
Ha yeah I forgot about tagging a new version. I'll do it asap.

pavlos2:
I set it to 100 playlists switch, the script runs with no issues! THANKS A MILLION Romain and the rest team!!! Looking forward to start working again with my project I started almost six months ago and to run more scripts!

toots:
Nice! @pranza, there's a windows build of the pre-release for 1.4.3 here: https://github.com/savonet/liquidsoap/releases/tag/v1.4.3-pre-release. It should be updated every time we commit in the branch.

pranza:
Hi Romain!

thanks, running this one since morning:
Liquidsoap 1.4.3+git@b2b28211

and it behaves just like before - already close to 1 gig of ram eaten, also 
opening telnet connection amplifies the effect - you can easily see 
kilobytes start ticking up every second...
so that must be not the clock within a clock issue then?..

regards,
p

pranza:
hello! so, a report after one day.
i've disabled automatic morningly kill-restart to test this new version and 
found out:
one liquidsoap process dead (exited)
another eaten up 1.3G of RAM and still running...
so it's just as before, weeee! :)

p

toots:
@pranza Let's begin a proper ticket to track that okay? Could you file a bug report with:
* Latest liquidsoap version
* OS
and, most importantly:
* script
* logs
Of a run with your memory leak?

Thanks!

pranza:
Hi Romain!

I think i've submitted the case here:
https://github.com/savonet/liquidsoap/issues/985

and even logs are of a little help as i observed that when it just crashes 
log just stops without any warnings...  and then the new session begins from 
there... and it's hard to predict when it will get so bad that it crashes... 
and crashing is only a part of the game, main thing is the memory leak 
(which i think eventually messes things up that the process stops streaming 
or crashes)

nothing is logged when it starts gradually leaking, nothing suspicious comes 
up when it starts telnet induced memory leak... but with telnet it at least 
starts the very second connection is made so it's easy to start tracking.

a simple script that it allows telnet, and connecting to telnet and leaving 
it like that, observing memory usage could be a way to start 
troubleshooting. if that can be sorted maybe all the rest will fall into 
place...

regards,
pranza

PR #1293, opened 2020-07-22
Enhance typing of lists.

smimram:
We simplify the typing of lists (I don't understand why we needed to compute the sup beforehand in previous version, this looks like a useless hack).

If the simple method does not work, we try to remove methods in order not to
uselessly force methods. Before that the following was not working:

```
queue = request.queue()
radio = fallback([queue, playlist])
```

because `queue` has a method `push` and `playlist` does not have it.

### DIFF
--- a/src/lang/lang_types.ml
+++ b/src/lang/lang_types.ml
@@ -845,7 +845,7 @@ type explanation = bool * t * t * repr * repr
 
 exception Type_Error of explanation
 
-let print_type_error error_header (flipped, ta, tb, a, b) =
+let print_type_error error_header ((flipped, ta, tb, a, b) : explanation) =
   error_header (print_pos_opt ta.pos);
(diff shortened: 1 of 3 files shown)


toots:
Also, while we're at it, could we deal with a list of value and optional values? Any other case we could think of?

toots:
Also: could we try to compute the intersection of methods? Otherwise, we would loose all common methods, i.e. 
```
# all sources have a skip() method
l = [source_1, source_2, ...]

# s below doesn't have a skip method
s = list.hd(l) 
```
That would be unfortunate..

smimram:
Your comments are interesting, and in fact I had been thinking about this, but this would be hackish. My idea was to add only those which are needed for now. Let me try to do better with at least
- common methods
- active sources / sources

PR #1296, opened 2020-07-23
Fix typing of methods invocations.

### LINKED ISSUE #1295: Typing of functions and methods
The following is very bad:
```
$ ./liquidsoap -i "def f(x) = x.b end"
f     : ('a) -> 'b
```

smimram:
It should fix #1295.


smimram:
Note for later:
```
def f(x) =
  if x.b then
    x.a
  else
    0
  end
end
```
fails with
```
At line 7, char 4-7:
Error 5: this value has type
  ?A (inferred at line 6, char 5-8)
but it should be a supertype of the type of the value at line 7, char 4-7
  ?A.{b : ?B, a : ?C}
```

PR #1298, opened 2020-07-24
Raise errors when files are not found.

smimram:


src/lang/builtins_files.ml
```diff
@@ -355,12 +361,12 @@ let () =
   add_builtin "file.which" ~cat:Sys
     ~descr:
       "`file.which(\"progname\")` looks for an executable named \"progname\" \
-       using directories from the PATH environment variable and returns \"\" \
+       using directories from the PATH environment variable and returns `null` \
```
toots, line 364: Is that really the case? Looks like you're returning `""`

smimram: Nope, I started this change but figured out that it would not be very convenient, I'll revert this. The reason is that we would like to do something like
```
f = file.which("prog")
if f != null then
  ...execute f...
end
```
which we cannot do easily because the typing does not know that we are not null in the then branch...

toots: Word

smimram: Half word: I'd really like to have a simple way to achieve this in Liq :)

smimram:
> Still thing a `file.error` would be nice :-)

Is it `file.error` or `error.file`?

toots:
> > Still thing a `file.error` would be nice :-)
> 
> Is it `file.error` or `error.file`?

`file.error` so it belongs to the `file` module? Then we can reuse this pattern w/ other modules..

Issue #1368, opened 2020-09-30
Liquidsoap as a Live Video HEVC (h.265) / Audio (AAC) SRT Server?


b3ck:
I saw someone use Liquidsoap for live streaming audio using the SRT protocol, so I guess I'm just wondering if there is a way to use Liquidsoap as an input / output for the SRT protocol using the HEVC (h.265) codec, basically I want to stream to it using either my phone or a hdmi encoder that supports SRT, then pull the stream into OBS Studio as a Media Source, and I would also need at stats page in json format to use with a script.

Can this be done? And if so, how? where do I start?
or is this a pipe dream?

Kind Regards,
-b3ck

toots:
Hi! The latest `master` should be able to support that. It is still experimental but we would love to work with you on this use-case to make sure it is stable for it.

Would you be getting encoded data to be sent back without re-encoding?

b3ck:
Using a `DDMALL HEV-4K` encoder, it will be taking audio & video via HDMI from a Sony FDR-x3000, then send it to the 'Liquidsoap SRT' server, after which it needs to be pulled into a Media Source in OBS Studio for further manipulation, OBS Studio will be doing the decoding using its built-in FFMPEG.

So, SRT(Encoder) > SRT SERVER(Liquidsoap) > SRT into OBS Studio (ffmpeg).

The Liquidsoap SRT server will need to support all of the latest SRT features and functions. (SRT v1.4.2)

The SRT Server will also need to provide real-time statistics via HTTP in JSON format, this should include but be limited to;
```JSON
{
  "publishers": {
    "publish/live/feed1": {
      "bitrate": 1441,
      "bytesRcvDrop": 0,
      "bytesRcvLoss": 0,
      "mbpsBandwidth": 453.732,
      "mbpsRecvRate": 1.777003845982417,
      "msRcvBuf": 101,
      "pktRcvDrop": 0,
      "pktRcvLoss": 0,
      "rtt": 56.011,
      "uptime": 3
    }
  },
  "status": "ok"
}
```

This will be for further scripting outside the scope of Liquidsoap, unless you can get Liquidsoap to connect to OBS Studio via Websockets to control scenes based on the current BITRATE and RTT coming from the SRT Server.

[...]

PR #1377, opened 2020-10-10
SRT two ways

toots:
This PR provides the first step toward implementing the SRT server described in #1368 by providing a `listener` and `caller` mode for both `input.srt` and `output.srt`. It also fixes a couple of glitches with encoders and SRT, allowing to support ffmpeg copy mode, which passes data without re-encoding.

Test script:
```ruby
set("log.level", 4)

s = input.srt(port=9000, mode="listener")

e = %ffmpeg(format="mp3",%audio.copy, %video.none)

output.srt(mode="listener", e, fallible=true, s)
```

Then, on a different console:
```
% liquidsoap 'output.srt(mode="caller", %mp3, single("/tmp/bla.mp3"), port=9000)'
```

And, finally, to check the resulting stream:
```
% liquidsoap 'output.ao(fallible=true,input.srt(mode="caller",clock_safe=false))
```


toots:
Steps to test with video and OBS are described here: https://github.com/savonet/liquidsoap/issues/1368#issuecomment-706729387

PR #1384, opened 2020-10-17
Add script convenience wrappers

toots:
This PR adds a couple of much needed convenience wrappers, namely:

### `%argsof`

* `%argsof(<function>)`
* `%argsof(<function>[<list of params>])`

These can be used to re-use a builtin function's argument in order to re-define it or re-use default values without having to re-enter them. 

`<list of params>` takes two types of entries:
* `<name>`: include this named argument
* `!<name>`: exclude this named argument

If no arguments are specified for inclusion, all arguments are included.

`%argsof` patterns can be used as:
* Function arguments: `%argsof(foo[...])` adds the selected named arguments from function `foo` with the same type and same default value
* Function application: `%argsof(foo[...])` is equivalent to adding the selected named arguments from `foo` as `<argument_name>=<argument_name>`.

### `@docof`, `@argsof`

* `# @docof <function`
* `# @argsof <argsof params>`

These can be used to re-use a previous function's doc. `@docof` copies the whole doc from the given function while `@argsof` copies only the given arguments. `<argsof params>` follows the same semantics as`%argsof`.

### Limitations
* Unlabelled parameters are not supported
* Only default values which can be converted back to a `term` are supported. Typically, a `source` default value is not supported.

Example with `output.shoutcast`:
```ruby
# Output to shoutcast.
# @category Source / Output
# @argsof output.icecast[!headers,!description,!protocol]
# @param ~icy_reset Reset shoutcast source buffer upon connecting (necessary for NSV).
# @param ~icy_id Shoutcast source ID. Only supported by Shoutcast v2.
# @param ~dj Callback to set dj name.
# @param e Encoding format. For shoutcast, should be mp3 or AAC(+).
# @param s The source to output
def output.shoutcast(
  %argsof(
   output.icecast[!headers,!description,!protocol]
  ),
  ~icy_reset=true,~icy_id=1,~dj={""},
  ~aim="",~icq="",~irc="",
  e, s) =
  icy_reset = if icy_reset then "1" else "0" end

  headers = [("icy-aim",aim),("icy-irc",irc),
             ("icy-icq",icq),("icy-reset",icy_reset)]

  def map(m) =
    dj = dj()
    if dj != "" then
      list.add(("dj",dj),m)
    else
      m
    end
  end
  s = map_metadata(map,s)

  output.icecast(
    %argsof(
     output.icecast[!headers,!description,!protocol]
    ),
    headers=headers,description="UNUSED",protocol="icy",
    e, s)
end
end
```


smimram:
This looks wonderful but I would like to first discuss the syntax which I find a bit heavy. Would it be possible to have, instead:
```
# Output to shoutcast.
# @category Source / Output
# @param ~icy_reset Reset shoutcast source buffer upon connecting (necessary for NSV).
# @param ~icy_id Shoutcast source ID. Only supported by Shoutcast v2.
# @param ~dj Callback to set dj name.
# @param e Encoding format. For shoutcast, should be mp3 or AAC(+).
# @param s The source to output
def output.shoutcast(
  ~icy_reset=true,~icy_id=1,~dj={""},
  ~aim="",~icq="",~irc="",
  e, s) extends output.icecast[!headers,!description,!protocol] =

  headers = ...
  ...

  output.icecast(headers=headers,description="UNUSED",protocol="icy", e, s)
end
```
This means that the keyword `extends` would
- add the labeled parameters to the doc
- add the labeled parameters to the function with their previous default value
- locally define (at the very begining of the body of the function) the extended function (here `output.icecast`), by setting all labeled parameters to the value of the arguments as default value, e.g.
  ```
  let output.icecast = output.icecast(id=id, connection_timeout=connection_timeout, ...)
  ```
  (we should make sure that there is at least one non-labeled or non-exported parameter so that the definition does not fully evaluate)

What do you think of that?

toots:
> This looks wonderful but I would like to first discuss the syntax which I find a bit heavy. Would it be possible to have, instead:
> 
> ```
> # Output to shoutcast.
> # @category Source / Output
> # @param ~icy_reset Reset shoutcast source buffer upon connecting (necessary for NSV).
> # @param ~icy_id Shoutcast source ID. Only supported by Shoutcast v2.
> # @param ~dj Callback to set dj name.
> # @param e Encoding format. For shoutcast, should be mp3 or AAC(+).
> # @param s The source to output
> def output.shoutcast(
>   ~icy_reset=true,~icy_id=1,~dj={""},
>   ~aim="",~icq="",~irc="",
>   e, s) extends output.icecast[!headers,!description,!protocol] =
> 
>   headers = ...
>   ...
> 
>   output.icecast(headers=headers,description="UNUSED",protocol="icy", e, s)
> end
> ```
> 
> This means that the keyword `extends` would
> 
> * add the labeled parameters to the doc
> * add the labeled parameters to the function with their previous default value
> * locally define (at the very begining of the body of the function) the extended function (here `output.icecast`), by setting all labeled parameters to the value of the arguments as default value, e.g.
>   ```
>   let output.icecast = output.icecast(id=id, connection_timeout=connection_timeout, ...)
>   ```
>   
>   
>   (we should make sure that there is at least one non-labeled or non-exported parameter so that the definition does not fully evaluate)
> 
> What do you think of that?

It looks nice and is doable but I'm not sure what it would bring more compares to the proposed syntax? 

My concern with your proposed syntax is that it will complicate the implementation a bunch so it's important to weight the benefits.

Regarding the documentation, I believe this won't work like that. Documentation is added via the pre-processor and is heavily dependent on code comments (and also quite ad-hoc but that's not for this PR).

Furthermore, for documentation, I'm really aiming at being able to replace the whole do to support easy wrappers for scripted server commands.

Now that I think about it, this would also be an issue with your proposed syntax. Here's how I'm envisioning it with mine:
```ruby
# @docof input.http
def input.http(%argsof(input.http)) =
  s = input.http(%argsof(input.http))

  # Register server command
  server.register("some command", fun c -> ...)

 # Return s
 s
end
```
Here, not just the arguments doc is copied but the whole doc. Furthermore, the function is fully evaluated as we are not touching any of its arguments..

toots:
Also, for what it's worth, I'm really intending this syntax for developers or advanced users, not for regular users.

toots:
After some discussions with @smimram we're gonna consider this good for now and meant for internal libs use only so we can merge now as-is.

PR #1438, opened 2020-12-15
Add a missing line in docstring (fade.skip's override_skip)

martinkirch:
It's too bad to find the correct value for `liq_skip_meta` only in the source !


toots:
Thanks!!

PR #1448, opened 2020-12-25
Switch to `Yuva420p

toots:
Turns out we actually support YUB420 planar with an alpha channel internally. This PR updates naming conventions and pixel format conversions with ffmpeg. ~~We're keeping `yuv420p` as the default encoder pixel format as most encoders don't support `yuva420p`. Also, we should auto-detect pixel format according to the encoder.~~ Done!


src/operators/video_effects.ml
```diff
@@ -259,6 +259,34 @@ let () =
...
+    [
+      ("width", Lang.int_getter_t (), None, Some "Box width.");
+      ("height", Lang.int_getter_t (), None, Some "Box height.");
+      ("x", Lang.int_getter_t (), Some (Lang.int 0), Some "x offset.");
+      ("y", Lang.int_getter_t (), Some (Lang.int 0), Some "y offset.");
+      ("alpha", Lang.float_getter_t (), None, Some "alpha value.");
+      ("", Lang.source_t return_t, None, None);
+    ]
+    ~return_t ~category:Lang.VideoProcessing
+    ~descr:"Set alpha value on a give box inside the image."
```
toots, line 274:
```suggestion
    ~descr:"Set alpha value on a given box inside the image."
```

PR #1465, opened 2021-01-11
Don't require source methods by default.

smimram:
By subtyping, it is not nececessary to require the methods for source arguments.

With this patch the type of `amplify` changes from
```
(?id : string, ?override : string?, {float},
 source(audio='b, video='c, midi='d)
 .{
   time : () -> float,
   shutdown : () -> unit,
   fallible : () -> bool,
   skip : () -> unit,
   seek : (float) -> float,
   is_up : () -> bool,
   remaining : () -> float,
   on_track : ((([string * string]) -> unit)) -> unit,
   on_shutdown : ((() -> unit)) -> unit,
   on_metadata : ((([string * string]) -> unit)) -> unit,
   is_ready : () -> bool,
   id : (() -> string)
 }) -> source(audio='b, video='c, midi='d)
.{
  time : () -> float,
  shutdown : () -> unit,
  fallible : () -> bool,
  skip : () -> unit,
  seek : (float) -> float,
  is_up : () -> bool,
  remaining : () -> float,
  on_track : ((([string * string]) -> unit)) -> unit,
  on_shutdown : ((() -> unit)) -> unit,
  on_metadata : ((([string * string]) -> unit)) -> unit,
  is_ready : () -> bool,
  id : (() -> string)}
```
to
```
(?id : string, ?override : string?, {float},
 source(audio='b, video='c, midi='d)) -> source(audio='b,
video='c, midi='d)
.{
  time : () -> float,
  shutdown : () -> unit,
  fallible : () -> bool,
  skip : () -> unit,
  seek : (float) -> float,
  is_up : () -> bool,
  remaining : () -> float,
  on_track : ((([string * string]) -> unit)) -> unit,
  on_shutdown : ((() -> unit)) -> unit,
  on_metadata : ((([string * string]) -> unit)) -> unit,
  is_ready : () -> bool,
  id : (() -> string)}
```
which is arguably more readable...

PR #1466, opened 2021-01-12
Print methods separately

smimram:
Documents returned methods outside the type. For instance, the doc of `amplify` was
```
Multiply the amplitude of the signal.

Type: (?id : string, ?override : string?, {float},
 source(audio='b, video='c, midi='d)) -> source(audio='b, video='c, midi='d)
.{
  time : () -> float,
  shutdown : () -> unit,
  fallible : () -> bool,
  skip : () -> unit,
  seek : (float) -> float,
  is_up : () -> bool,
  remaining : () -> float,
  on_track : ((([string * string]) -> unit)) -> unit,
  on_shutdown : ((() -> unit)) -> unit,
  on_metadata : ((([string * string]) -> unit)) -> unit,
  is_ready : () -> bool,
  id : (() -> string)}

Category: Source / Sound Processing

Parameters:

 * id : string (default: "")
     Force the value of the source ID.

 * override : string? (default: "liq_amplify")
     Specify the name of a metadata field that, when present and well-formed,
     overrides the amplification factor for the current track. Well-formed
     values are floats in decimal notation (e.g. `0.7`) which are taken as
     normal/linear multiplicative factors; values can be passed in decibels
     with the suffix `dB` (e.g. `-8.2 dB`, but the spaces do not matter).

 * (unlabeled) : {float} (default: None)
     Multiplicative factor.

 * (unlabeled) : source(audio='b, video='c, midi='d) (default: None)
```
and with this PR it becomes
```
Multiply the amplitude of the signal.

Type: (?id : string, ?override : string?, {float},
 source(audio='b, video='c, midi='d)) -> source(audio='b, video='c, midi='d)

Category: Source / Sound Processing

Parameters:

 * id : string (default: "")
     Force the value of the source ID.

 * override : string? (default: "liq_amplify")
     Specify the name of a metadata field that, when present and well-formed,
     overrides the amplification factor for the current track. Well-formed
     values are floats in decimal notation (e.g. `0.7`) which are taken as
     normal/linear multiplicative factors; values can be passed in decibels
     with the suffix `dB` (e.g. `-8.2 dB`, but the spaces do not matter).

 * (unlabeled) : {float} (default: None)
     Multiplicative factor.

 * (unlabeled) : source(audio='b, video='c, midi='d) (default: None)

Methods:

 * id : () -> string

 * is_ready : () -> bool

 * on_metadata : ((([string * string]) -> unit)) -> unit

 * on_shutdown : ((() -> unit)) -> unit

 * on_track : ((([string * string]) -> unit)) -> unit

 * remaining : () -> float

 * is_up : () -> bool

 * seek : (float) -> float

 * skip : () -> unit

 * fallible : () -> bool

 * shutdown : () -> unit

 * time : () -> float
```
(the above example supposes that #1465 has been merged).

Next step would be to add proper documentation for methods, but this PR gets readability of documentation back to acceptable I think.

### DIFF
--- a/src/lang/lang.ml
+++ b/src/lang/lang.ml
@@ -228,6 +228,26 @@ let builtin_type p t =
 
 let to_plugin_doc category flags main_doc proto return_t =
   let item = new Doc.item ~sort:false main_doc in
+  let meths, return_t =
+    let rec aux t =
+      match (T.deref t).T.descr with
+        | T.Meth (l, mt, t) ->
+            let m, t = aux t in
+            ((l, mt) :: m, t)
+        | _ -> ([], t)
+    in
+    aux return_t
+  in
+  let meths =
+    let meths_items = new Doc.item "" in
+    List.iter
+      (fun (m, (generalized, t)) ->
+        let i = new Doc.item "" in
+        i#add_subsection "type" (T.doc_of_type ~generalized t);
+        meths_items#add_subsection m i)
+      meths;
+    meths_items
+  in
   let t = builtin_type proto return_t in
   let generalized = T.filter_vars (fun _ -> true) t in
   item#add_subsection "_category" (Doc.trivial category);
@@ -235,6 +255,7 @@ let to_plugin_doc category flags main_doc proto return_t =
   List.iter
     (fun f -> item#add_subsection "_flag" (Doc.trivial (string_of_flag f)))
     flags;
+  item#add_subsection "_methods" meths;
   List.iter
     (fun (l, t, d, doc) ->
       item#add_subsection
--- a/src/lang/lang.ml
+++ b/src/lang/lang.ml
@@ -238,16 +238,6 @@ let to_plugin_doc category flags main_doc proto return_t =
     in
     aux return_t
   in
-  let meths =
-    let meths_items = new Doc.item "" in
-    List.iter
-      (fun (m, (generalized, t)) ->
-        let i = new Doc.item "" in
-        i#add_subsection "type" (T.doc_of_type ~generalized t);
-        meths_items#add_subsection m i)
-      meths;
-    meths_items
-  in
   let t = builtin_type proto return_t in
   let generalized = T.filter_vars (fun _ -> true) t in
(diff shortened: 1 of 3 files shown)


toots (commented):
Is there a way to connect the methods to a certain type/value? If I look at the message right now, it's not clear where the methods do apply..

smimram:
For now I only extract the methods from the returned value, so there should be no ambiguity there (the methods in the arguments are still indicated as before, thus #1465 to reduce their number).

toots:
What happens if the operator returns more than one source?

toots:
I was thinking of something like this:
```
Multiply the amplitude of the signal.

Type: (?id : string, ?override : string?, {float},
 source(audio='b, video='c, midi='d)) -> source(audio='b, video='c, midi='d) as var

Category: Source / Sound Processing

Parameters:

 * id : string (default: "")
     Force the value of the source ID.

 * override : string? (default: "liq_amplify")
     Specify the name of a metadata field that, when present and well-formed,
     overrides the amplification factor for the current track. Well-formed
     values are floats in decimal notation (e.g. `0.7`) which are taken as
     normal/linear multiplicative factors; values can be passed in decibels
     with the suffix `dB` (e.g. `-8.2 dB`, but the spaces do not matter).

 * (unlabeled) : {float} (default: None)
     Multiplicative factor.

 * (unlabeled) : source(audio='b, video='c, midi='d) (default: None)

var methods:

 * id : () -> string

 * is_ready : () -> bool

 * on_metadata : ((([string * string]) -> unit)) -> unit

 * on_shutdown : ((() -> unit)) -> unit

 * on_track : ((([string * string]) -> unit)) -> unit

 * remaining : () -> float

 * is_up : () -> bool

 * seek : (float) -> float

 * skip : () -> unit

 * fallible : () -> bool

 * shutdown : () -> unit

 * time : () -> float
```

With possibly more `var`

smimram:
I agree with you, but I think we can leave this for a second time, correctly naming subtypes will take some more work...

toots:
Works for me, this is clearly an improvement.

PR #1471, opened 2021-01-13
Add support for documenting methods.

smimram:
Document methods in scripts with comments of the form
```
@method meth This documents the method
```

### DIFF
--- a/src/lang/lang_pp.ml
+++ b/src/lang/lang_pp.ml
@@ -261,12 +261,13 @@ let parse_comments tokenizer =
            (fun (n, _) -> String.length n = 0 || n.[0] <> '_')
            doc#get_subsections)
     in
-    let rec parse_doc (main, special, params) = function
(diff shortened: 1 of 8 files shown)


src/lang/lang_pp.ml
```diff
@@ -261,12 +261,13 @@ let parse_comments tokenizer =
            (fun (n, _) -> String.length n = 0 || n.[0] <> '_')
            doc#get_subsections)
     in
-    let rec parse_doc (main, special, params) = function
-      | [] -> (main, special, params)
+    let rec parse_doc (main, special, params, methods) = function
```
toots, line 264: Might be time to structure this with a record..

toots (approved):
Looks great, thanks!

src/lang/lang_pp.ml
smimram: Yes but well this is only used very locally...

PR #1472, opened 2021-01-14
Add documentation for internal methods.

smimram:
This add documentation for builtins. For instance:
```
Multiply the amplitude of the signal.

Type: (?id : string, ?override : string?, {float},
 source(audio='b, video='c, midi='d)) -> source(audio='b,
video='c, midi='d)

Category: Source / Sound Processing

Parameters:

 * id : string (default: "")
     Force the value of the source ID.

 * override : string? (default: "liq_amplify")
     Specify the name of a metadata field that, when present and well-formed,
     overrides the amplification factor for the current track. Well-formed
     values are floats in decimal notation (e.g. `0.7`) which are taken as
     normal/linear multiplicative factors; values can be passed in decibels
     with the suffix `dB` (e.g. `-8.2 dB`, but the spaces do not matter).

 * (unlabeled) : {float} (default: None)
     Multiplicative factor.

 * (unlabeled) : source(audio='b, video='c, midi='d) (default: None)

Methods:

 * fallible : () -> bool
     Indicate if a source may fail, i.e. may not be ready to stream.

 * id : () -> string
     Identifier of the source.

 * is_ready : () -> bool
     Indicate if a source is ready to stream, or currently streaming.

 * is_up : () -> bool
     Check whether a source is up.

 * on_metadata : ((([string * string]) -> unit)) -> unit
     Call a given handler on metadata packets.

 * on_shutdown : ((() -> unit)) -> unit
     Register a function to be called when source shuts down.

 * on_track : ((([string * string]) -> unit)) -> unit
     Call a given handler on new tracks.

 * remaining : () -> float
     Estimation of remaining time in the current track.

 * seek : (float) -> float
     Seek forward, in seconds (returns the amount of time effectively
     seeked).

 * shutdown : () -> unit
     Deactivate a source.

 * skip : () -> unit
     Skip to the next track.

 * time : () -> float
     Get a source's time, based on its assigned clock.
```

This is complementary to #1471 which add supports for adding documentation in scripts. Merging the two is expected to be non-trivial (although not very difficult).

### DIFF
--- a/src/lang/lang_parser.mly
+++ b/src/lang/lang_parser.mly
@@ -88,7 +88,7 @@
       in
       let get_meth_type () =
         match t.T.descr with
-          | T.Meth (_, _, t) -> t
+          | T.Meth (_, _, _, t) -> t
           | _ -> assert false
       in
       let term =
--- a/src/sources/harbor_input.ml
+++ b/src/sources/harbor_input.ml
@@ -271,15 +271,18 @@ module Make (Harbor : T) = struct
         [
           ( "stop",
             ([], Lang.fun_t [] Lang.unit_t),
+            "Stop the input.",
             fun s ->
               Lang.val_fun [] (fun _ ->
                   s#stop_cmd;
                   Lang.unit) );
           ( "status",
             ([], Lang.fun_t [] Lang.string_t),
+            "Current statu of the input.",
             fun s -> Lang.val_fun [] (fun _ -> Lang.string s#status_cmd) );
           ( "buffer_length",
             ([], Lang.fun_t [] Lang.float_t),
+            "Length of the buffer (in seconds).",
             fun s -> Lang.val_fun [] (fun _ -> Lang.float s#buffer_length_cmd)
           );
         ]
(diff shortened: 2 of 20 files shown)


src/sources/harbor_input.ml
```diff
@@ -271,15 +271,18 @@ module Make (Harbor : T) = struct
...
           ( "stop",
             ([], Lang.fun_t [] Lang.unit_t),
+            "Stop the input.",
             fun s ->
               Lang.val_fun [] (fun _ ->
                   s#stop_cmd;
                   Lang.unit) );
           ( "status",
             ([], Lang.fun_t [] Lang.string_t),
+            "Current statu of the input.",
```
smimram, line 281:
```suggestion
            "Current status of the input.",
```

src/lang/lang_parser.mly
```diff
@@ -88,7 +88,7 @@
       in
       let get_meth_type () =
         match t.T.descr with
-          | T.Meth (_, _, t) -> t
+          | T.Meth (_, _, _, t) -> t
```
toots, line 91: It could be time to structure this with a record..

toots (commented):
Looking good! One minor remark & also it looks like some methods are still missing doc: https://github.com/savonet/liquidsoap/pull/1472/checks?check_run_id=1701266830

PR #1480, opened 2021-01-22
Native implementation of RTMP.

smimram:
The protocol does not seem that difficult to implement, although quite dirty...

Let me try.


smimram:
Moved development to https://github.com/savonet/ocaml-rtmp for now.

PR #1490, opened 2021-02-12
Add length to playlist.reloadable.

smimram:
Add a method to obtain the length of a playlist. You can test with
```
p = playlist.reloadable("~/Music")
print("**** length: #{p.length()}")
output.pulseaudio(mksafe(p))
```
This was requested by @dhannyz on slack.


toots (commented):
I've been meaning to go deeper and actually re-implement the queues on the script side. How do you feel about that? There are a lot of things that could be cleaned up real good and better exposed to the user like primary/secondary queues, default duration etc.

smimram:
I mistakenly merged #1495 instead of this one, which had the effect of merging this one too.

smimram:
Yes, it would be great to do this! But with another PR :)

PR #1493, opened 2021-02-16
Implement interactive values on script side

smimram:
This is a first step twoard having interactive values on other things than telnet.

My plan is
- add interaction through an harbord webpage
- add the possibility to store the values so that they are persistent


toots (approved):
This looks great thank you!

smimram:
I mistakenly merged #1495 instead of this one, which had the effect of merging this one too.

PR #1496, opened 2021-02-16
Add support for setting in ref getters.

smimram:
Adding a `set` method to getters from refs.

It currently fails because subtyping is apparently not smart enough:
```
$ ./getter.liq 
At line 27, char 6:
Error 5: this value has type
  (() -> int).{set : ((int) -> unit)} (inferred at ../../libs/ref.liq, line 4, char 2-43)
but it should be a subtype of the type of the value at line 7, char 7
  {int}
```


smimram:
Strangely enough, the following works:
```
r = {4}.{f = fun(x)->print(x)}
print(r())
```

smimram:
So the problem was subtyping vs constraints... Should be fixed now.

PR #1502, opened 2021-02-20
Serve telnet over http

smimram:
Following an idea of @martinkirch, if you type
```
server.harbor()
```
You get a nice webpage on <http://localhost:8080/telnet> where you can play with the telnet:
telnet


toots (commented):
That's a good one!

toots (approved):

martinkirch:
Can confirm it's working ! I'm thankful I won't have to maintain a telnet client forever :pray: 

"One more thing": `localhost` is hardcoded in the log message, but it seems to bind according to the `harbor.bind_addrs` parameter (for example, on my test machine the correct URL is http://192.168.1.33:8080/telnet ) would it be possible to show it in the log ?

martinkirch:
On second thought, re-using `harbor.bind_addrs` may create a security breach : as opposed to `server.telnet.bind_addr`, `harbor.bind_addrs` will likely be a public address. The current implem assumes users have set-up port-level filtering.

I think this page should bind to `server.telnet.bind_addr` ; otherwise the function documentation should include a warning.

PR #1506, opened 2021-02-21
Allow reopening pipe on error.

smimram:
Add an option to reopen pipe operators when an error happens. Requested by @dhannyz.


toots:
Should this be backported to `1.4.4`?

smimram:
It can I think, yes. (although it is an improvement, not a bugfix)

toots:
This might also fix https://github.com/savonet/liquidsoap/issues/1510 🙂

PR #1507, opened 2021-02-21
Nice printing of internal dataypes

smimram:
For instance, this patch changes the type of `blank` fom

```
(?id : string, ?duration : float) -> source(audio='a, video='b, midi=none)
where
  'a, 'b is an internal media type (none, pcm, yuva420p or midi)
```

to 

```
(?id : string, ?duration : float) -> source(audio=internal, video=internal, midi=none)
```

**Question**. Before we merge the above patch I'd like to understand what it the difference between the two above types and

```
(?id : string, ?duration : float) -> source(audio=pcm('a), video=yuva420p('b), midi=none)
```

and thus, do we really need/want the "internal type" constraint for type variables? Is it thtat *internal = pcm or none*?

PS: we could easily generate blank midi I think...

### DIFF
--- a/src/lang/lang_types.ml
+++ b/src/lang/lang_types.ml
@@ -496,6 +496,12 @@ let print_repr f t =
     | `UVar (_, [Getter a]) when !pretty_getters ->
         Format.fprintf f "{%s}" (print_ground a);
         vars
+    | `EVar (_, [InternalMedia]) ->
+        Format.fprintf f "?internal";
+        vars
+    | `UVar (_, [InternalMedia]) ->
+        Format.fprintf f "internal";
+        vars
     | `EVar (name, c) | `UVar (name, c) ->
         Format.fprintf f "%s" name;
         if c <> [] then DS.add (name, c) vars else vars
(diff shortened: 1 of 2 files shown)


toots:
If I remember well, this comes from a historical leniency w.r.t frame content. A lot of operators accept either some content or no content at all and are happy to skip over non-existent content. This is particularly true for all video operators.

This leads to situations where where we have to say: "this operator accepts either nothing or <internal format type>", so pcm for audio or yuv420p for video. The internal media constraint collects this into a single constraint for simplicity.

A better solution would be to strictly enforce content type and introduce sub-modules that can be opened to enforce certain constraint, for instance a default opened module that requires audio but no video, a `audio_video` module that has the same operators but expects video as well and a `video` (or `video_only`) module, etc etc.

In this case we should be able to drop the internal constraint and properly enforce single media type for each operator.

Hope this makes sense.

smimram:
Ok, I am fine with this explanation, let's keep this like this for now.

Do you agree with my patch to simplify the printing?

toots (approved):

PR #1523, opened 2021-03-03
Use an explicit cast in the language

### LINKED ISSUE #1522: Type casts have no effect on final content kinds
Consider the following script
```
log.level.set(5)
s = single("test.mp3")
output((s:source(audio=pcm(mono))))
```
Even though I explicitly require mono pcm, we still get stereo (which is the default). It should be the case that typecasts constrain the kind at execution however.

smimram:
In this way the cast is correctly propagated to the runtime kind for sources, which fixes #1522.

### DIFF
--- a/src/lang/lang.ml
+++ b/src/lang/lang.ml
@@ -310,6 +310,7 @@ let iter_sources f v =
       | Term.List l -> List.iter (iter_term env) l
       | Term.Tuple l -> List.iter (iter_term env) l
       | Term.Null -> ()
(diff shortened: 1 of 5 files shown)


toots (approved):
This looks good! Do you know why the previous way of doing it wasn't working?

smimram:
The previous way was not working because we are not casting the result of a FFI application (and I am not sure we want to do this because this would make things less efficient).

PR #1531, opened 2021-03-09
Allow getters on any type

smimram:
This should simplify much getter-related code...

### DIFF
--- a/libs/getter.liq
+++ b/libs/getter.liq
@@ -17,7 +17,7 @@ def getter.float_of_int(x)
 end
 
 # Convert a float getter to a int getter.
-# @category Math
+# @category Liquidsoap
 def getter.int_of_float(x)
   getter.map(int_of_float, x)
 end
--- a/libs/getter.liq
+++ b/libs/getter.liq
@@ -11,7 +11,7 @@ def getter.is_constant(x)
 end
 
 # Convert an int getter to a float getter.
-# @category Math
+# @category Liquidsoap
 def getter.float_of_int(x)
   getter.map(float_of_int, x)
 end
--- a/libs/getter.liq
+++ b/libs/getter.liq
@@ -7,7 +7,7 @@ end
 # Determine if a getter is a constant.
 # @category Liquidsoap
 def getter.is_constant(x)
-  getter.elim(x, fun (_) -> true, fun (_) -> false)
+  getter.case(x, fun (_) -> true, fun (_) -> false)
 end
 
 # Convert an int getter to a float getter.
--- a/libs/getter.liq
+++ b/libs/getter.liq
@@ -1,20 +1,45 @@
+# Construct a function returning the value of a getter.
+# @category Liquidsoap
+def getter.function(x)
+  {getter.get(x)}
+end
+
+def getter.is_constant(x)
+  getter.elim(x, fun (_) -> true, fun (_) -> false)
+end
+
+# Apply a function to a getter.
+# @category Liquidsoap
+def getter.map(f, x)
+  getter(f(getter.get(x)))
+
+  # TODO: both optimizations below are not typable
+
+  # getter.elim(x, fun(x) -> getter(f(x)), fun (x) -> getter({f(x())}))
+
+  # if getter.is_constant(x) then
+    # getter(f(getter.get(x)))
+  # else
+    # getter({f(getter.get(x))})
+  # end
+end
+
 # Convert an int getter to a float getter.
 # @category Math
 def getter.float_of_int(x)
-  x = to_int_getter(x)
-  float_getter({float_of_int(x())})
+  getter.map(float_of_int, x)
 end
 
 # Convert a float getter to a int getter.
 # @category Math
 def getter.int_of_float(x)
-  x = to_float_getter(x)
-  int_getter({int_of_float(x())})
+  getter.map(int_of_float, x)
 end
 
 # Execute a function when the value of the getter changes.
 # @category Math
 def getter.on_change(f, x)
+  x = {getter.get(x)}
   old = ref(x())
   fun () -> begin
     new = x()
@@ -26,9 +51,9 @@ end
 # Detect whether the value of the getter changes.
 # @category Math
 def getter.changes(x)
-  old = ref(x())
+  old = ref(getter.get(x))
   fun () -> begin
-    new = x()
+    new = getter.get(x)
     if !old != new then old := new; true
     else false end
   end
--- a/libs/getter.liq
+++ b/libs/getter.liq
@@ -35,7 +35,7 @@ def getter.on_change(f, x)
 end
 
 # Detect whether the value of the getter changes.
-# @category Math
+# @category Liquidsoap
 def getter.changes(x)
   old = ref(getter.get(x))
   fun () -> begin
--- a/libs/getter.liq
+++ b/libs/getter.liq
@@ -8,22 +8,6 @@ def getter.is_constant(x)
   getter.elim(x, fun (_) -> true, fun (_) -> false)
 end
 
-# Apply a function to a getter.
-# @category Liquidsoap
-def getter.map(f, x)
-  getter(f(getter.get(x)))
-
-  # TODO: both optimizations below are not typable
-
-  # getter.elim(x, fun(x) -> getter(f(x)), fun (x) -> getter({f(x())}))
-
-  # if getter.is_constant(x) then
-    # getter(f(getter.get(x)))
-  # else
-    # getter({f(getter.get(x))})
-  # end
-end
-
 # Convert an int getter to a float getter.
 # @category Math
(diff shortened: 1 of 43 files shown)


libs/http.liq
```diff
@@ -132,7 +132,9 @@ def harbor.http.static.base(serve,~content_type,~basepath,~port,~headers,directo
   end
 
   def register(method)
-    serve(port=port,method=method,"^#{basepath}",handler(method=method))
+    # TODO: restore me...
+    # serve(port=port,method=method,"^#{basepath}",handler(method=method))
```
toots, line 136: This seems unrelated?

libs/video.liq
```diff
@@ -27,6 +27,14 @@ end
 def video.add_image(~id="", ~width=int_getter(-1), ~height=int_getter(-1), ~x=int_getter(0), ~y=int_getter(0), ~file, s)
   image =
     if getter.int.is_constant(x) and getter.int.is_constant(y) and getter.int.is_constant(width) and getter.int.is_constant(height) then
+      width = to_int_getter(width)
```
toots, line 30: Isn't `to_int_getter` working even if `getter.int.is_constant(width)` is constant?

toots: Actually I'm not sure what this code means..

src/lang/lang.ml
```diff
@@ -501,6 +502,11 @@ let to_string_list l = List.map to_string (to_list l)
 let to_int_list l = List.map to_int (to_list l)
 let to_source_list l = List.map to_source (to_list l)
 
+let to_getter t =
+  match (demeth t).value with
+    | Fun ([], _, _, _) | FFI ([], _, _) -> fun () -> apply t []
```
toots, line 507: This seems arbitrary? What about functions with multiple arguments? We just return `fun () ->  fn` in this case?

toots (commented):
This looks interesting. I don't see why not in general, but do you have a particular usage in mind?

Also, any concern with cases where we might end up comparing abstract values? I'm not sure if that could happen but in this wouldn't work well.

smimram:
There are two main reasons for this PR:
- one is that it is tiring to write the type in the function eg `int_getter(4)`, whereas you can now write `getter(4)`
- the other is that we needed more than ground types : I have an example where I need a getter on metadata
I agree that for some types it can be fishy, but
- nobody will use that for complex types anyway :)
- more seriously, we could easily add a constraint to enforce "resonable" types (e.g. not functions), which would still include more than ground types (e.g. lists of ground types)

src/lang/lang.ml
smimram: Not really arbitrary: a getter is either x or fun () -> x, we detect here when we are in the second case.

libs/http.liq
smimram: It was related: the change induced a subtle typing problem related to the fact that we have to correctly handle variance now that we have non-ground types. This is fixed anyway.

smimram:
@toots the PR looks ok to me, do you want to review/comment again?

libs/getter.liq
```diff
@@ -1,20 +1,31 @@
...
+end
+
+# Determine if a getter is a constant.
+# @category Liquidsoap
+def getter.is_constant(x)
+  getter.case(x, fun (_) -> true, fun (_) -> false)
+end
+
 # Convert an int getter to a float getter.
 # @category Math
```
toots, line 14:
```suggestion
# @category Liquidsoap
```

libs/getter.liq
```diff
@@ -1,20 +1,31 @@
...
 def getter.float_of_int(x)
-  x = to_int_getter(x)
-  float_getter({float_of_int(x())})
+  getter.map(float_of_int, x)
 end
 
 # Convert a float getter to a int getter.
 # @category Math
```
toots, line 20:
```suggestion
# @category Liquidsoap
```

libs/getter.liq
```diff
@@ -1,20 +1,31 @@
...
 def getter.int_of_float(x)
-  x = to_float_getter(x)
-  int_getter({int_of_float(x())})
+  getter.map(int_of_float, x)
 end
 
 # Execute a function when the value of the getter changes.
 # @category Math
```
toots, line 26:
```suggestion
# @category Liquidsoap
```

libs/getter.liq
```diff
@@ -26,9 +37,9 @@ end
 # Detect whether the value of the getter changes.
 # @category Math
```
toots, line 38:
```suggestion
# @category Liquidsoap
```

There might be more like that that are not in the diff.

libs/gstreamer.liq
```diff
@@ -33,7 +33,7 @@ let gstreamer.single = ()
 # @param ~on_error Callback executed when an error happens.
 # @param uri URI of the file to be played.
 def gstreamer.single.audio(~id="",~on_error=fun(_)->3.,uri) =
-  uri = to_string_getter(uri)
+  uri = getter.function(uri)
```
toots, line 36: This is great.

libs/http.liq
```diff
@@ -105,7 +105,7 @@ def harbor.http.static.base(serve,~content_type,~basepath,~port,~headers,directo
     ret = string.extract(pattern="^#{basepath}([^?]*)",uri)
 
     if list.length(ret) == 0 then
-      http.response(protocol=protocol,code=404,headers=responseHeaders,data="Not Found!")
+      getter(http.response(protocol=protocol,code=404,headers=responseHeaders,data="Not Found!"))
```
toots, line 108: Yes!

src/lang/lang.ml
```diff
@@ -67,11 +67,12 @@ let nullable_t t = T.make (T.Nullable t)
...
-let string_getter_t () = getter_t T.String
-let float_getter_t () = getter_t T.Float
-let int_getter_t () = getter_t T.Int
-let bool_getter_t () = getter_t T.Bool
+let getter_t a = T.make (T.Getter a)
+let ground_getter_t a = getter_t (T.make (T.Ground a))
+let string_getter_t () = ground_getter_t T.String
+let float_getter_t () = ground_getter_t T.Float
+let int_getter_t () = ground_getter_t T.Int
+let bool_getter_t () = ground_getter_t T.Bool
```
toots, line 75: Is this still needed?

src/lang/lang_types.ml
```diff
@@ -120,16 +117,12 @@ let print_ground v =
...
+type constr = Num | Ord | Dtools | InternalMedia
 type constraints = constr list
 
 let print_constr = function
   | Num -> "a number type"
   | Ord -> "an orderable type"
-  | Getter t ->
-      let t = print_ground t in
-      if !pretty_getters then Printf.sprintf "{%s}" t
-      else Printf.sprintf "either %s or ()->%s" t t
```
toots, line 132: Pretty cool that this is no longer a constraint, indeed.

src/lang/lang_types.ml
```diff
@@ -492,8 +486,10 @@ let print_repr f t =
         let vars = print ~par:false vars t in
         Format.fprintf f "]@]";
         vars
-    | `Getter a ->
-        Format.fprintf f "{%s}" (print_ground a);
+    | `Getter t ->
+        Format.fprintf f "{";
+        let vars = print ~par:false vars t in
+        Format.fprintf f "}";
```
toots, line 492: Are we concerned that this will clash with the syntax `{v}` for `()->v`?

src/lang/lang_types.ml
```diff
@@ -1019,6 +999,11 @@ let rec ( <: ) a b =
     | Ground (Format k), Ground (Format k') -> (
         try Frame_content.merge k k' with _ -> raise (Error (repr a, repr b)) )
     | Ground x, Ground y -> if x <> y then raise (Error (repr a, repr b))
+    | Getter t1, Getter t2 -> (
+        try t1 <: t2 with Error (a, b) -> raise (Error (`Getter a, `Getter b)) )
+    | Arrow ([], t1), Getter t2 -> (
+        try t1 <: t2
+        with Error (a, b) -> raise (Error (`Arrow ([], a), `Getter b)) )
```
toots, line 1006: So, a getter can be unified with a function with no arguments. I see. Are we assuming that, for all practical purposes, they are the same but with a different underlying implementation just like `Fun` and `FFI`?

toots (approved):
Looks great. Couple of comments but I don't think they should be blocking.

src/lang/lang_types.ml
smimram: It's not worse than earlier and the two are "mostly consistent".

src/lang/lang_types.ml
smimram: The point it that a function with no argument *is* a valid getter: it is not just the way we implement that, we can actually provide `fun() -> ...` where a getter is expected.

PR #1538, opened 2021-03-18
Switch to dune!

toots:


src/bin/main.ml
```diff
@@ -63,8 +63,8 @@ let allow_root =
 (* Do not run, don't even check the scripts. *)
 let parse_only = ref false
 
-(* Should we load the pervasives? *)
-let pervasives = ref true
+(* Should we load the load_libs? *)
```
toots, line 66:
```suggestion
(* Should we load the pervasives? *)
```

smimram:
BTW, is there a good reason why this has not been merged? Are we waiting for some missing support from dune?

toots:
Yes, there are several issues with `dune` preventing this from happening. I'm keeping track of them.

PR #1560, opened 2021-04-06
Add handler when source is leaving.

smimram:
A common problem consists in skipping a source when it is not used anymore in a `switch`.  When `switch` changes sources is "leaves" the old source. The current patch allows registering a callback on this situation. We can thus do something like:

```
m = playlist("~/Music")
m.on_leave({print("leaving!"); m.skip()})
s = sine()
s = switch(track_sensitive=false, [({time () mod 10. <= 1.}, s), ({true}, m)])
output(s)
```

Each time `s` kicks in `m` is left and thus skipped.

### DIFF
--- a/src/lang/builtins_source.ml
+++ b/src/lang/builtins_source.ml
@@ -140,6 +140,23 @@ let () =
       Lang.unit)
 
 let () =
+  add_builtin "source.on_leave" ~cat:Sys
+    [
+      ("", Lang.source_t (Lang.univ_t ()), None, None);
+      ("", Lang.fun_t [] Lang.unit_t, None, None);
+    ]
+    Lang.unit_t
+    ~descr:
+      "Register a function to be called when source is not used anymore by \
+       another souce."
+    (fun p ->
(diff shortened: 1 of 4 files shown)


src/lang/builtins_source.ml
```diff
@@ -139,6 +139,23 @@ let () =
...
+    Lang.unit_t
+    ~descr:
+      "Register a function to be called when source is not used anymore by \
+       another souce."
+    (fun p ->
+      let s = Lang.to_source (Lang.assoc "" 1 p) in
+      let f = Lang.assoc "" 2 p in
+      let wrap_f () = ignore (Lang.apply f []) in
+      s#on_leave wrap_f;
+      Lang.unit)
```
toots, line 157: Do we need this?

toots (approved):
This looks good. I don't think that we need the top-level `source.on_leave`. If anything it should be in the script side.

src/lang/builtins_source.ml
smimram: Yes in order to have decent types for functions in scripts.

Issue #1550, opened 2021-03-25
Add handlers for asynchronous errors


toots:
Now that we have support for proper errors (yay!), we need a way to handle errors raised in the asychronous tasks.

Once this is done, we'll need to re-activate the disabled check in `tests/lang/list.liq`

PR #1564, opened 2021-04-10
Add handlers for asynchronous errors.

toots:
This PR ads support for catching runtime errors inside asynchronous queues. It should be recommended to ad such handlers every time one makes heavy use of asynchronous tasks.

Additionally, we should start visiting the code and see how much of the runtime OCaml exceptions that we raise should be converted to `Lang_value.Runtime_error` so that we can give the user more flexibility with how to handle errors.

Example:
```ruby
e = error.register("foo")

def on_error(~backtrace, ~queue_name, e) =
  print("caught error #{e} from queue #{queue_name} and backtrace:\n#{backtrace}")
end

thread.on_error(e, on_error)

thread.run(fun () -> error.raise(e, "Asynchronous error"))
```

Fixes: #1550

### DIFF
--- a/src/lang/builtins_thread.ml
+++ b/src/lang/builtins_thread.ml
@@ -80,7 +80,7 @@ let () =
         (false, "queue_name", Lang.string_t);
         (false, "", Builtins_error.Error.t);
       ]
-      Lang.bool_t
+      Lang.unit_t
   in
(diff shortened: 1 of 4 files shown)


smimram (approved):
Nice idea!

PR #1567, opened 2021-04-12
Support quality parameter ala ffmpeg CLI

smimram:
When trying
```
output.file(%ffmpeg(format="mp3", %audio(codec="libmp3lame", samplerate=44100, q=4)), "/tmp/out", s)
```
Liquidsoap complained that the option `q` was not defined (whereas it should be). It seems to me that this is because the logic for detecting unused parameters is inverted.

@toots: do you confirm this?

### DIFF
--- a/src/lang_encoders/lang_ffmpeg.ml
+++ b/src/lang_encoders/lang_ffmpeg.ml
@@ -24,6 +24,20 @@ open Lang_values
 open Lang_values.Ground
 open Lang_encoders
 
+(* Looks like this is how ffmpeg CLI does it.
+   See: https://github.com/FFmpeg/FFmpeg/blob/4782124b90cf915ede2cebd871be82fc0267a135/fftools/ffmpeg_opt.c#L1567-L1570 *)
+let set_global_quality q f =
+  let flags =
+    match Hashtbl.find_opt f.Ffmpeg_format.other_opts "flags" with
+      | Some (`Int f) -> f
+      | Some _ -> assert false
+      | None -> 0
+  in
+  let flags = flags lor Avcodec.flag_qscale in
+  Hashtbl.replace f.Ffmpeg_format.other_opts "flags" (`Int flags);
+  Hashtbl.replace f.Ffmpeg_format.other_opts "global_quality"
+    (`Float (float Avutil.qp2lambda *. q))
+
 let ffmpeg_gen params =
   let defaults =
     {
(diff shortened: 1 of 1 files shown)


smimram:
However, I am not really sure : `mediainfo /tmp/out` claims that the output file is CBR whereas it should be VBR if `q` was taken in account...

toots (approved):
Thanks for that. The video options below has the same problem.

toots:
One thing to note, tho:
```
q (-V)
Set constant quality setting for VBR. This option is valid only using the ffmpeg command-line tool. For library interface users, use global_quality.
```
From: https://ffmpeg.org/ffmpeg-codecs.html

We are a library interface.. Let me investigate this a little more.

toots:
Ok, no, the logic is correct. When creating the encoder, the ffmpeg API removes all dictionary values that have been used. Then, what we do is get a copy of the (imperative) hashtbl table for the audio options and only keep the key/values that are still present in the global options hash. If the resulting hash is not empty this means that some key/values haven't been used.

I can confirm that using `global_quality` works. Let me see if I can replicate the trick that the `ffmpeg` CLI does to support `q` (and `qscale`).

smimram:
Do you really confirm that it works? When I try
```
s = mksafe(playlist("~/Music"))
output.file(%ffmpeg(format="mp3", %audio(codec="libmp3lame", samplerate=44100, global_quality=4)), "/tmp/out.mp3", s)
```
`mediainfo` gives
```
General
Complete name                            : /tmp/out
Format                                   : MPEG Audio
File size                                : 89.2 KiB
Duration                                 : 5 s 694 ms
Overall bit rate mode                    : Constant
Overall bit rate                         : 128 kb/s
Writing library                          : LAME3.100
FileExtension_Invalid                    : m1a mpa mpa1 mp1 m2a mpa2 mp2 mp3

Audio
Format                                   : MPEG Audio
Format version                           : Version 1
Format profile                           : Layer 3
Format settings                          : Joint stereo / MS Stereo
Duration                                 : 5 s 695 ms
Bit rate mode                            : Constant
Bit rate                                 : 128 kb/s
Channel(s)                               : 2 channels
Sampling rate                            : 44.1 kHz
Frame rate                               : 38.281 FPS (1152 SPF)
Compression mode                         : Lossy
Stream size                              : 89.0 KiB (100%)
Writing library                          : LAME3.100
```
whereas we should have VBR, isn't it?

Otherwise, supporting `q` as a parameter is a good thing!

toots:
Hmm weird, I just did this:
```
output.file(%ffmpeg(format="mp3", %audio(codec="libmp3lame", samplerate=44100, q=4)), fallible=true, on_stop=shutdown, "/tmp/out.mp3", sine(duration=20.))
```
And got that:
```
❯ mediainfo /tmp/out.mp3
General
Complete name                            : /tmp/out.mp3
Format                                   : MPEG Audio
File size                                : 87.2 KiB
Duration                                 : 20 s 9 ms
Overall bit rate mode                    : Variable
Overall bit rate                         : 35.6 kb/s
Writing library                          : LAME3.100

Audio
Format                                   : MPEG Audio
Format version                           : Version 1
Format profile                           : Layer 3
Format settings                          : Joint stereo / MS Stereo
Duration                                 : 20 s 10 ms
Bit rate mode                            : Variable
Bit rate                                 : 35.6 kb/s
Channel(s)                               : 2 channels
Sampling rate                            : 44.1 kHz
Frame rate                               : 38.281 FPS (1152 SPF)
Compression mode                         : Lossy
Stream size                              : 86.9 KiB (100%)
Writing library                          : LAME3.100
```

PR #1574, opened 2021-04-15
Switch to ocurl

toots:
This PR switches our internal `http` query support to use `libcurl`. As it is right now, it would make `ocurl` and `libcurl` mandatory dependencies.

In details:
- `http.{get,put,post,head,delete}` are re-implemented using `libcurl` and now support `https` urls, which has been very confusing for our users in the past.
- `https.{get,put,post,head,delete}` are deprecated .

A couple of new APIs are introduced:
-  `file.write` can now receive a nullable string getter, which allows for streamed file write. It was added only to realize that this wasn't what I needed but it's cool to have!
- `file.write.stream` returns a `() -> string?` callback that allows to actively push streamed data to the file to write to
- `http.{get,put,post,delete,head}.stream` accepts a `() -> stream?` callback that can be used to stream a query's response body.

With these new APIs, the real exciting change here is:
- Request resolutions are now implemented on the script side with no need for an external dependency on the `curl` binary! 🎉

Couple more notes:
- I am considering fully re-implementing `input.http` using `libcurl` and also renaming it. This operator is overloaded (why does it support playlists?) and confusing (it only support real-time data from icecast). But this won't be part of this PR.
- The harbor operators still require their own implementation and won't work with `libcurl`. This means that for now, we'd still have a dependency on the `ssl` module.
- The windows build are tricky. MXE has a recipe for `curl` but it needs the latest MXE recipes. Ideally, we should rebuild a new `ocaml-cross` docker image with but it currently fails to build the `ocaml` compiler (see: https://github.com/ocaml-cross/opam-cross-windows/issues/192). This is a lot of work to fix so we might have to disable the windows build for a minute.


toots:
I can confirm that a clean `mxe` clone successfully builds `libcurl` so we should be able to build again on `win32` as soon as the cross-compiler is fixed. That was the last blocker for me here.

PR #1586, opened 2021-04-19
Add test that exhibits file.watch problem at toplevel.

smimram:


toots:
I think this is good to go.

Issue #1580, opened 2021-04-18
Arguments for harbor auth function.


smimram:
The change in 6e8f895d080c09b1eee40d02ad1349df74f3449e adds the IP of the user to the authentication function. Because of this, all scripts will have to rewrite their auth functions in order to add the `address` argument. This is ok for 2.0 but not very future-proof: we might want to add the connection time or so in the future... Maybe a better style, would be to pass a record for the user as unique argument, ie switch from
```
def auth(~address, login, pass)
  login == "admin" and pass == ""
end
```
to
```
def auth(user)
  user.login == "admin" and user.password == ""
end
```
so that we can add more fields without impact on existing scripts?

toots:
Great idea!!

PR #1587, opened 2021-04-19
Use records for cross arguments.

### LINKED ISSUE #1585: Use records for cross' transition function
We should do the same as in #1580 for `cross`. Currently, the type is
```
(?id : string, ?active : bool, ?conservative : bool,
 ?duration : {float}, ?minimum : float,
 ?override_duration : string, ?width : float,
 ((float, float, [string * string], [string * string],
   source(audio=pcm('a), video='b, midi='c),
   source(audio=pcm('a), video='b, midi='c)) ->
  source(audio=pcm('a), video='b, midi='c)),
 source(audio=pcm('a), video='b, midi='c)) ->
source(audio=pcm('a), video='b, midi='c)
```
which is quite complicated: we want the crossing function to take two sources as argument (not 6 arguments), each one being decorated by fields indicating properties such as metadata, power, etc.

smimram:
Fixes #1585.

### DIFF
--- a/src/operators/cross.ml
+++ b/src/operators/cross.ml
@@ -375,15 +375,17 @@ class cross ~kind (s : source) ~cross_length ~override_duration ~rms_width
               let params =
                 [
                   ( "",
-                    Lang.meth (Lang.source a)
+                    Lang.meth Lang.unit
                       [
-                        ("power", Lang.float db_before);
+                        ("source", Lang.source a);
+                        ("db_level", Lang.float db_before);
                         ("metadata", Lang.metadata before_metadata);
                       ] );
                   ( "",
-                    Lang.meth (Lang.source b)
+                    Lang.meth Lang.unit
                       [
-                        ("power", Lang.float db_after);
+                        ("source", Lang.source b);
+                        ("db_level", Lang.float db_after);
                         ("metadata", Lang.metadata after_metadata);
                       ] );
                 ]
@@ -433,9 +435,10 @@ let () =
   let kind = Lang.audio_pcm in
   let k = Lang.kind_type_of_kind_format kind in
   let transition_arg =
-    Lang.method_t (Lang.source_t k)
+    Lang.method_t Lang.unit_t
       [
-        ("power", ([], Lang.float_t), "Power of the source (in dB).");
+        ("source", ([], Lang.source_t k), "Source");
+        ("db_level", ([], Lang.float_t), "dB level of the source.");
         ("metadata", ([], Lang.metadata_t), "Metadata of the source.");
       ]
   in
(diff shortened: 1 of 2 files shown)


src/operators/cross.ml
```diff
@@ -374,12 +374,18 @@ class cross ~kind (s : source) ~cross_length ~override_duration ~rms_width
...
-                  ("", Lang.float db_before);
-                  ("", Lang.float db_after);
-                  ("", Lang.metadata before_metadata);
-                  ("", Lang.metadata after_metadata);
-                  ("", Lang.source a);
-                  ("", Lang.source b);
+                  ( "",
+                    Lang.meth (Lang.source a)
+                      [
+                        ("power", Lang.float db_before);
```
toots, line 380: I think `power` is a bit vague. `dbLevel` would be better.

toots:
> We have a typing error there for good reasons in `fades.liq`. Basically, the transition function does not consider power which thus does not show up in the type.
> 
> We could work around by adding
> 
> ```
> ignore(a.power)
> ignore(b.power)
> ```
> 
> in the transition function, but the standard users will not find this by themselves. The problem here is that our typing system does not allow to say that we take as input a type of the form
> 
> ```
> source.{metadata : metadata, ...}
> ```
> 
> (with the `...` indicating that we might have more...)

If I remember correctly, adding this would be pretty complex right?

smimram:
> If I remember correctly, adding this would be pretty complex right?

This should be doable, but definitely post-2.0. I have some ideas that I want to try, but this would mean heavily re-thinking the type system once again... :)

toots:
> > If I remember correctly, adding this would be pretty complex right?
> 
> This should be doable, but definitely post-2.0. I have some ideas that I want to try, but this would mean heavily re-thinking the type system once again... :)

I hear you. Should we revert https://github.com/savonet/liquidsoap/issues/1580?

smimram:
> > > If I remember correctly, adding this would be pretty complex right?
> > 
> > 
> > This should be doable, but definitely post-2.0. I have some ideas that I want to try, but this would mean heavily re-thinking the type system once again... :)
> 
> I hear you. Should we revert #1580?

Nope, I had a thought about that: we can represent `{x:int,...}` because it is a `'a` decorated with a field `x:int` (i.e. `'a.{x:int}`). However we cannot represent a type such as `source.{x:int,...}`. So for pure records it's fine, but not for decorated values. For cross we could take 4 four arguments: a source + a record of parameter (for each of the two sources).

toots:
> > > > If I remember correctly, adding this would be pretty complex right?
> > > 
> > > 
> > > This should be doable, but definitely post-2.0. I have some ideas that I want to try, but this would mean heavily re-thinking the type system once again... :)
> > 
> > 
> > I hear you. Should we revert #1580?
> 
> Nope, I had a thought about that: we can represent `{x:int,...}` because it is a `'a` decorated with a field `x:int` (i.e. `'a.{x:int}`). However we cannot represent a type such as `source.{x:int,...}`. So for pure records it's fine, but not for decorated values. For cross we could take 4 four arguments: a source + a record of parameter (for each of the two sources).

Neat!

PR #1617, opened 2021-05-03
Add sync API for input.ffmpeg.

### LINKED ISSUE #1604: Synchronization issue with input.ffmpeg
Trying `input.ffmpeg` on `testsrc` with
```
s = (input.ffmpeg(buffer=0.1, format="lavfi", "testsrc=size=1280x720:rate=25"):source(audio=none))
output.video(s)
```
The timing is clearly wrong, seconds go faster than they should:
out

toots:
This PR adds the possibility to run a buffered source synchronously. We still need the buffer to simplify the process of queueing decoded data but we delegate the control of the clock and synchronization back to the source.

This should fix #1604 and introduce a pattern that could be re-used for other inputs, e.g. `input.http`

### DIFF
--- a/src/io/ffmpeg_io.ml
+++ b/src/io/ffmpeg_io.ml
@@ -23,21 +23,21 @@
 module Generator = Generator.From_audio_video_plus
 module Generated = Generated.Make (Generator)
 
-class input ~bufferize ~log_overfull ~kind ~start ~on_start ~on_stop ?format
-  ~opts url =
+class input ~self_sync ~clock_safe ~bufferize ~log_overfull ~kind ~start
+  ~on_start ~on_stop ?format ~opts url =
   let max_ticks = 2 * Frame.main_of_seconds bufferize in
   (* A log function for our generator: start with a stub, and replace it
    * when we have a proper logger with our ID on it. *)
   let log_ref = ref (fun _ -> ()) in
   let log x = !log_ref x in
   object (self)
-    inherit Source.source ~name:"input.ffmpeg" kind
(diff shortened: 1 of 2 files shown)


src/io/ffmpeg_io.ml
```diff
@@ -153,6 +157,19 @@ let () =
...
+          Some
+            "Set to `true` if the input is a real-time stream, i.e. icecast, \
+             SRT, etc. and `false` is not, i.e. local file, etc.." );
+        ( "clock_safe",
+          Lang.bool_t,
+          Some (Lang.bool false),
+          Some
+            "Set to `true` if the source needs to be attached to its own \
+             clock. Should be set to `true` when `self_sync` is set to `true` \
+             in most cases." );
```
smimram, line 172: Do we really need this parameter since it seems to be implied by `self_sync`? At least I would put a `Lang.nullable_t Lang.bool_t` so that the default value is the same as `self_sync`.

src/sources/generated.ml
```diff
@@ -84,6 +82,28 @@ module Make (Generator : Generator.S) = struct
...
+
+      method virtual private clock : Source.clock_variable
+
+      val mutable clock = None
+
+      method private get_clock =
+        match clock with
+          | Some c -> c
+          | None ->
+              let c = new Clock.clock "srt" in
```
smimram, line 98: `"srt"`, really?

smimram (approved):

mcfiredrill:
Sounds promising ! :tada:

Issue #1616, opened 2021-05-03
ALSA source crashes at startup, when POSTing metadata


martinkirch:
With v2.0.0+git@d42c7ecb, my test radio crashes right after start-up. It happens in a long script, so I tried to iteratively rebuild the script from its output to see when this happens, so here is a smaller script that crashes:
```
set("log.level",10)
SHOWERGEL = "http://localhost:2345"

radio = fallback(track_sensitive=false, [
  playlist("/home/martin/Musique"),
  blank()
])

def post_to_showergel(metadata)
   response = http.post(
    headers=[("Content-Type", "application/json")],
    "#{SHOWERGEL}/metadata_log",
    data=json_of(metadata)
  )
end
source.on_track(radio, post_to_showergel)

output.alsa(radio)
```

See the complete log [here](https://gist.github.com/martinkirch/ee5e9a2b4c8e62224f399933fabdbfa0)

This script will work
 - if `source.on_track` is commented
 - when replacing `http.post...` with only `log(json_of(metadata))`

Reminds me #1590 where
> Showergel receives [ ] when starting the stream (I guess that's the only sample ever sent by blank() at the bottom of my fallback)

martinkirch:
and it's triggered by `blank()` because this script works when replacing `radio` by
```
radio = fallback(track_sensitive=false, [
  mksafe(playlist("/home/martin/Musique")),
])
```

smimram:
With some better logging, the error is more understandable:
```
2021/05/03 22:46:27 [lang:2] Could not perform http request: CurlException(Couldnt_connect, 7, CURLE_COULDNT_CONNECT).
2021/05/03 22:46:27 [clock.alsa:4] Source alsa_out(default) failed while streaming: Runtime_error.Runtime_error(_)!
2021/05/03 22:46:27 [clock.alsa:4] Raised at file "tools/runtime_error.ml", line 31, characters 2-52
2021/05/03 22:46:27 [clock.alsa:4] Called from file "lang/builtins_http.ml", line 143, characters 10-50
2021/05/03 22:46:27 [clock.alsa:4] Called from file "lang/lang_values.ml", line 1154, characters 8-12
...
```
Do you get the same error? (I obviously don't have a server on localhost 2345)

martinkirch:
Yes - I forgot I had shut down the server on this machine...

Even with a running server I got, after a few tracks
```
2021/05/03 23:29:54 [switch_29450:5] Got metadata at position 0: calling handlers...
[...]

PR #1627, opened 2021-05-10
Generic error handler

toots:
This PR extends `thread.on_error` to work on all runtime exceptions, including exceptions thrown in the main streaming thread.

The idea is to provide users with a default handler for runtime errors so that they do not have to handle them unless they need/want to and also so that runtime errors never interrupt the streaming loop, which is most important to the application.

Errors handlers are registered as a stack that is, last registered handlers is the first one to be called. We register a default handler in the runtime library that will be called if the user hasn't registered their own afterward.

Lastly, it is still possible to catch a runtime error via the usual `try .. catch` syntax, but only when needed by the user.

This should, along with the `Not_found` fix, address #1616

### DIFF
--- a/libs/thread.liq
+++ b/libs/thread.liq
@@ -30,3 +30,9 @@ def thread.when(~fast=true, ~every=1., ~once=false, ~changed=true, p, f)
   end
   thread.run.recurrent(fast=fast, delay=0., check)
 end
+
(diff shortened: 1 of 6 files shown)


smimram (approved):
Looks fine to me!

Issue #1576, opened 2021-04-16
string.quote vs process.quote


smimram:
Right now we have two functions for quoting strings: `string.quote` and `process.quote`. We should decide whether we want to keep the two or not, and in the second case we should make clear the difference between the two (I could not really understand that from the doc).

dtcooper:
FYI: `process.quote(...)` was a suggestion I made and @toots implemented to fix a security hole, that `string.quote(...)` still suffers from. [See relevant discussion here](https://github.com/savonet/liquidsoap/issues/1215). In my opinion we should remove (or alias) `string.quote`. Thanks!

smimram:
Thanks @dtcooper for the explanation. I find this both confusing and worrying since the stnadard library uses `string.quote` where some process execution is involved (and thus `process.quote` should be used instead if I understand correctly):
```
$ git grep string.quote
deprecations.liq:# Deprecated: this function has been replaced by `string.quote`.
deprecations.liq:  deprecated("quote", "string.quote")
deprecations.liq:  string.quote(s)
externals.liq:    json = process.read("#{ffprobe} -print_format json -show_streams #{string.quote(fname)}")
externals.liq:          process.read.lines("#{mpcdec} -i #{string.quote(file)} 2>&1 \
externals.liq:    mpcdec_p = fun(f) -> "#{mpcdec} #{string.quote(f)} -"
externals.liq:                                              --show-channels #{string.quote(fname)}"))
externals.liq:                   #{string.quote(fname)}")
externals.liq:    faad_p = (fun (f) -> "#{faad} -w #{string.quote(f)}")
externals.liq:                           process.read.lines("#{faad} -i #{string.quote(file)} 2>&1 | \
externals.liq:        ret = process.read.lines("#{faad} -i #{string.quote(file)} 2>&1")
externals.liq:  osd = 'osd_cat -p #{position} --font #{string.quote(font)}'
externals.liq:    ignore(process.run("echo #{string.quote(display(m))} | #{osd} &"))
externals.liq:       ^ ' -t #{time} #{string.quote(title)} '
externals.liq:  s.on_metadata(fun (m) -> ignore(process.run(send^string.quote(display(m)))))
externals.liq:                       -vc null -vo null #{string.quote(s)} 2>/dev/null")
file.liq:      list.hd(default="", process.read.lines("file -b --mime-type #{string.quote(fname)}"))
http.liq:  list.hd(default="", process.read.lines("file -b -I #{string.quote(file)}"))
[...]

PR #1635, opened 2021-05-19
Remove string.quote. Fixes: #1576

toots:


smimram:
Sorry to be picky, but I would rather write `string.quote` than `process.quote`. Would it be possible to change both to the current behavior of `process.quote` and deprecate `process.quote` instead?

toots:
Done!

src/lang/builtins_string.ml
```diff
@@ -394,10 +394,14 @@ let () =
       Lang.string (Utils.interpolate (fun k -> List.assoc k l) s))
 
 let () =
-  add_builtin "string.quote" ~cat:String ~descr:"Escape shell metacharacters."
+  add_builtin "string.quote" ~cat:String
+    ~descr:
+      "Return a quoted copy of the given string, suitable for use as one \
```
smimram, line 399:
```suggestion
      "Return a quoted copy of the given string, suitable for use as an \
```

smimram (approved):
Go!

PR #1639, opened 2021-05-20
Update README.md

dromer:
There are probably other docs, but this one was the most obvious to me.


toots:
IRC slack bridge updated, merging this thks! 

Is there an admin password that we can store somewhere so we don't loose access to the channel?

Thanks for taking care of this!

dromer:
Chanserv registrations do not work with admin passwords. I can add another user as founder though as a fail-safe. Just point me to the right username :)

PR #1656, opened 2021-05-28
Input http curl reboot

toots:
This PR shares some of the same technical concerns described in #1659. It also re-implements `input.http` using `libcurl` and deprecates `input.https`.

The implementation is slightly tricky but I believe that we'll be happy to have it in the future in particular because `libcurl` is widely used and we can expect a wide range of support for different HTTP protocols and technical details that we do not want to handle on our own.

Additionally, this PR makes it so that we only require a single thread for all running inputs.


toots:
Closing this. We'll re-implement `input.http` using `ffmpeg` instead, the code improvement and cleanliness is pretty hard to beat.. See: https://github.com/savonet/liquidsoap/pull/1659

PR #1664, opened 2021-06-03
Fix equality of records

### LINKED ISSUE #1583: Equality of records
Right now we have
```
# {x=3} == {x=5};;
- : bool = true
```
because only the unit part of records is compared...

smimram:
Should fix #1583.


src/lang/builtins_bool.ml
```diff
@@ -27,9 +27,29 @@ let () =
...
     add_builtin name ~cat:Bool ~descr:"Comparison of comparable values."
       [("", t, None, None); ("", t, None, None)] Lang.bool_t (fun p ->
-        let a = Lang.assoc "" 1 p |> Lang.demeth in
-        let b = Lang.assoc "" 2 p |> Lang.demeth in
-        Lang.bool (op (Lang.compare_values a b)))
+        let a = Lang.assoc "" 1 p in
+        let b = Lang.assoc "" 2 p in
+        let a' = Lang.demeth a in
+        let b' = Lang.demeth b in
+        (* For records, we also compare fields. *)
```
toots, line 34: Why only for records?

src/lang/builtins_bool.ml
```diff
@@ -27,9 +27,29 @@ let () =
...
+        let ans =
+          if a'.Lang.value = Lang.Tuple [] && b'.Lang.value = Lang.Tuple [] then (
+            let r a =
+              let m, _ = Lang_values.V.split_meths a in
+              m
+            in
+            let a = r a in
+            let b = r b in
+            (* TODO: the order is not the expected one on records (for < we want
+               one field to be < and the other to be <=). *)
```
toots, line 44: What do you mean?

src/lang/builtins_bool.ml
```diff
@@ -27,9 +27,29 @@ let () =
...
+            List.for_all
+              (fun (l, v) ->
+                let v' = List.assoc l b in
+                op (Lang.compare_values v v'))
+              a )
```
toots, line 49: Do you need to also test the other direction, i.e. that all methods from `b` are in `a`?

src/lang/lang_types.ml
```diff
@@ -666,12 +667,20 @@ let rec bind a0 b =
...
                   let rec check b =
-                    let b = demeth b in
+                    let m, b = split_meths b in
                     match b.descr with
                       | Ground _ -> ()
                       | EVar (j, c) ->
                           if List.mem Ord c then ()
                           else b.descr <- EVar (j, Ord :: c)
+                      | Tuple [] ->
+                          (* For records, we want to ensure that all fields are ordered. *)
```
toots, line 677: What does that entail?

tests/language/record.liq
```diff
@@ -70,7 +70,11 @@ def f() =
   ignore(list.hd(l).x)
   l = [b,a]
   ignore(list.hd(l).x)
-  def f(c) = [a,c] end      
+  def f(c) = [a,c] end
+
+  # Equality
+  t({a = 5} == {a = 5}, true)
+  t({a = 5} == {a = 6}, false)
```
toots, line 77:
```suggestion
  t({a = 5} == {a = 6}, false)
  t({a = 5, b = 3} == {a = 6}, false)
  t({a = 5} == {a = 6, b = 4}, false)
```

toots (commented):
Some questions!

tests/language/record.liq
toots: Any maybe some more, like positive tests etc.?

src/lang/builtins_bool.ml
smimram: My idea is that we would tend to write something like
```
if http.get("http://...") == "bla" then
  ...
end
```
to compare the result of getting a page. And we don't wan't to take methods in accounts in this case.

And most of the time, excepting for records, fields are functions, which we cannot compare...

src/lang/builtins_bool.ml
smimram: Given records `r` and `s` with two fields `a` and `b`, the current implementations declares that `r < s` when both `r.a < s.a` and `r.b < s.b` whereas the natural order is rather
- `r.a < r.a` and `r.b <= s.b`, or
- `r.a <= r.a` and `r.b < s.b`

This could be fixed but I doubt that it's worth cluttering the code because I don't think that `<` comparison will be used that much on records...

src/lang/builtins_bool.ml
smimram: We compare two things of the same type, so `a` and `b` should have the same methods.

src/lang/lang_types.ml
smimram: For a record to be orderable, all its fields have to be orderable. For other values, as explained above, we don't compare the fields, so we don't check for orderability (and again, they are functions in general, so they would not be orderable).

tests/language/record.liq
smimram: Argh, I thought that
```
{a = 5, b = 3} == {a = 6}
```
would not type because the two don't have the same type, but I was apparently wrong...

src/lang/builtins_bool.ml
toots: Works for me.

src/lang/builtins_bool.ml
toots: Ohh okay yes.

tests/language/record.liq
toots: That might have consequence on your assumption above?

PR #1666, opened 2021-06-03
Stdlib!

toots:
* Rename `pervasives` into `stdlib`
* Error when `stdlib.liq` cannot be found
* Make sure tests also load deprecated APIs


smimram (approved):
It's more clear as a name, go for it!

PR #1671, opened 2021-06-04
Remove active_source type.

toots:
The notion of `active_source` is mostly internal. It means that this a source (output or not) that will always be animated by its clock once it is available. This is useful for sources such as `input.http` that need to maintain an active connection, and outputs, of course.

However, this notion does not provide anything to the end user and complicates the typing system, in particular when joining sources in a list and etc.

Thus, this PR removes the `active_source` type. The only blind spot is output and warnings for top-level unused variables:
* Either we accept to not warn on any source left unused at top-level
* Or we change the type of outputs.

I suggest that we should use `unit` type for outputs, which would effectively make them leaf nodes in the streaming graph, which seems like what we always thought of them as anyways.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -94,6 +94,7 @@ Changed:
 - Implemented per-frame clock synchronization mechanism, should allow for more
   advanced flexibility when working with source synchronization while keeping
   the default safe behavior. (#1012)
+- Remove `active_source` type, make all output return `unit` type. (#1671)
 - Switch to YUV420 as internal image format, much more efficient (#848).
 - Use bigarrays for audio buffers (#950).
 - Re-implemented switch-derived operators (`fallback`, `rotate`, `random`) as
--- a/libs/io.liq
+++ b/libs/io.liq
@@ -35,10 +35,10 @@ let output.video = output.dummy
 %ifdef output.sdl
   def output.video(%argsof(output.sdl), s)
     if output.sdl.has_video() then
-      (output.sdl(%argsof(output.sdl), s):source)
+      (output.sdl(%argsof(output.sdl), s):unit)
     else
       # Avoid using SDL when there is no video output.
-      (output.dummy(fallible=fallible, s):source)
+      (output.dummy(fallible=fallible, s):unit)
     end
   end
 %endif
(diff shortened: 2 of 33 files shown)


smimram:
Didn't have time to read that yet, but I'd also be in favor of having unit as return type for outputs.

toots:
> Didn't have time to read that yet, but I'd also be in favor of having unit as return type for outputs.

Great! I think the way I did it is pretty clean. The only real consequence is that then you have to use the clock assignment operators on sources only but I think this is fine.

smimram:
> > Didn't have time to read that yet, but I'd also be in favor of having unit as return type for outputs.
> 
> Great! I think the way I did it is pretty clean. The only real consequence is that then you have to use the clock assignment operators on sources only but I think this is fine.

Can you detail what would happen if we assign a clock to an active source? Nothing or an error?

Also, maybe could we have @dbaelde point of view before proceeding?

smimram:
Also, an idea: it would be nice to indicate whether a source is self-synchronizing in the doc at least...

toots:
> Also, an idea: it would be nice to indicate whether a source is self-synchronizing in the doc at least...

Ho yeah, definitely! Added a ticket for it: https://github.com/savonet/liquidsoap/issues/1692

toots:
> > > Didn't have time to read that yet, but I'd also be in favor of having unit as return type for outputs.
> > 
> > 
> > Great! I think the way I did it is pretty clean. The only real consequence is that then you have to use the clock assignment operators on sources only but I think this is fine.
> 
> Can you detail what would happen if we assign a clock to an active source? Nothing or an error?

It's exactly the same as before.

> Also, maybe could we have @dbaelde point of view before proceeding?

I agree but we should also not let this go stale for too long.

smimram:
So, this is almost a greenlight here excepting for one point. I would leave the `~active:true` argument to `add_operator` and indicate this information in the documentation. Maybe can we think of a more general system of flags to indicate properties of the source : active / self-syncing / etc.

toots:
> So, this is almost a greenlight here excepting for one point. I would leave the `~active:true` argument to `add_operator` and indicate this information in the documentation. Maybe can we think of a more general system of flags to indicate properties of the source : active / self-syncing / etc.

That's a great idea. Would you mind if we did it in a separate PR?

PR #1698, opened 2021-06-17
Indicate properties of operators.

smimram:
It would be a good thing to indicate in the doc whether sources are
- fallible or not
- active or not
- self-syncing or not
I propose that we add this information for all operators. Since this is quite an intrusive change, I would like to first know: @toots do you agree with this proposal for implementing this?

After that we need to change all operators...

### DIFF
--- a/src/io/alsa_io.ml
+++ b/src/io/alsa_io.ml
@@ -307,8 +307,9 @@ let () =
           Some "Alsa device to use" );
         ("", Lang.source_t k, None, None);
       ] )
-    ~active:true ~return_t:k ~category:Lang.Output ~meth:Output.meth
-    ~descr:"Output the source's stream to an ALSA output device."
+    ~return_t:k ~category:Lang.Output ~meth:Output.meth
+    ~descr:"Output the source's stream to an ALSA output device." ~active:true
+    ~self_sync:true
     (fun p ->
       let e f v = f (List.assoc v p) in
       let bufferize = e Lang.to_bool "bufferize" in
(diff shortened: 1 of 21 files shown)


toots:
I agree with the proposal but with the following remarks:

* We should also add `clock_safe` and perhaps rename it.
* Fallible and active is statically `true` or `false` only
* `self_sync` should be either `true`/`false` statically or `true` when active (active here meaning when the source is being used)

I understand that some of these sources have parameters to change these values but I believe they should be "default" values, indicating the default behavior the user can expect.

We should accompany this PR with a new page in the online doc that explains what these entries mean too. I'm down to write it.

smimram:
Is it better now?

toots:
> Is it better now?

Yes!

toots:
I'm working on this. I should have mentioned that those flags really mean something on sources. Not sure how to convey how operators work..

smimram:
@toots I filled in the parameters I could, but I would need a serious review on this one: I am far from being sure that I filled everything correctly or that I did not miss something...

A way to review is
```
LIQ_FREI0R=false LIQ_LADSPA=false ./liquidsoap --no-stdlib --list-functions-md 
```
At the end of the doc of operators there is now something like
```
This function is passive, fallible, not self-syncing and clock-safe.
```

toots:
Will have a look shortly, thanks!

toots:
Sorry for the lack of response here. Still down to consider it once rebased.

PR #1700, opened 2021-06-18
Self-sync and conversions improvements

toots:
This is a rather large PR sorry!

* Make `Producer_consumer` infallible when possible, which includes inline `ffmpeg` encoder and decoder!
* Add static/dynamic annotation to `self_sync`
* Added `Child_support` to clarify logic when dealing with children sources in a clock driven by a parent source
* Added `~start:bool` to `new Clock` to allow dependent clocks without runtime thread when driven by a parent source
* Use new `self_sync` annotation to fail when underlying source is not statically `self_sync` in `Child_support`
* Use `Child_support` for all operators that control their underlying source clock (`cross`, `soundtouch`, `stretch` etc..)
* Factorized logic for `soundtouch` and `stretch` to use `Producer_consumer.consumer`, which gets complete `frame` per each clock tick.
* Add parameter for default `clock_safe` value in `Start_stop` protocol
* Unify ffmpeg's `clock_safe` with other operators, set to `false` by default
* Rewrote `cue_cut` to simplify its code
* Switch default `id` operator parameter to `string?`

### DIFF
--- a/doc/content/build.md
+++ b/doc/content/build.md
@@ -42,7 +42,7 @@ Libraries not developed by Savonet are:
 | -------------- | ------------------------- |
 | OCaml compiler | >= 4.08.0     |
 | ocaml-dtools   |  >= 0.4.2    |
-| ocaml-duppy    |  >= 0.8.1     |
+| ocaml-duppy    |  >= 0.9.1     |
 | ocaml-mm       |  >= 0.7.0        |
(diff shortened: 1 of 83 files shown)


toots:
Branch name was causing `win32` builds to fail. New PR here: https://github.com/savonet/liquidsoap/pull/1705

PR #1705, opened 2021-06-20
Self-sync and conversions improvements

toots:
This is a rather large PR sorry!

* Make `Producer_consumer` infallible when possible, which includes inline `ffmpeg` encoder and decoder!
* Add static/dynamic annotation to `self_sync`
* Added `Child_support` to clarify logic when dealing with children sources in a clock driven by a parent source
* Added `~start:bool` to `new Clock` to allow dependent clocks without runtime thread when driven by a parent source
* Use new `self_sync` annotation to fail when underlying source is not statically `self_sync` in `Child_support`
* Use `Child_support` for all operators that control their underlying source clock (`cross`, `soundtouch`, `stretch` etc..)
* Factorized logic for `soundtouch` and `stretch` to use `Producer_consumer.consumer`, which gets complete `frame` per each clock tick.
* Add parameter for default `clock_safe` value in `Start_stop` protocol
* Unify ffmpeg's `clock_safe` with other operators, set to `false` by default
* Rewrote `cue_cut` to simplify its code
* Switch default `id` operator parameter to `string?`

### DIFF
--- a/libs/hls.liq
+++ b/libs/hls.liq
@@ -6,10 +6,11 @@ let input.hls = ()
 # @param ~reload How often (in seconds) the playlist should be reloaded.
 # @param uri Playlist URI.
 # @flag experimental
-def input.hls.native(~id="",~reload=10.,uri)
+def input.hls.native(~id=null(),~reload=10.,uri)
   playlistr = ref([])
   sequence = ref(0)
   playlist_uri = ref(uri)
+  id = string.id.default(default="input.hls.native", id)
 
   def load_playlist () =
     pl = request.create(!playlist_uri)
@@ -96,7 +97,7 @@ let replaces input.hls = input.ffmpeg
 # @category Source / Input
 # @param ~id Force the value of the source ID.
 # @param uri Playlist URI.
-def input.hls(~id="", uri)
+def input.hls(~id=null(), uri)
   input.hls(id=id, uri)
 end
 
--- a/src/io/ffmpeg_io.ml
+++ b/src/io/ffmpeg_io.ml
@@ -54,7 +54,17 @@ class input ?(name = "input.ffmpeg") ~autostart ~self_sync ~poll_delay ~debug
     method is_ready =
       super#is_ready && self#mutexify (fun () -> container <> None) ()
 
-    method self_sync = self_sync && self#is_ready
+    (* This is a hack. We really want to avoid using these sources in operators
+       that control latency such as [cross], [soundtouch] and etc. It's not just
+       about pulling data at a higher rate than real-time, which would work on a
+       regular HTTP server-based media file, it's also about pulling data at a
+       lower rate than real-time, typically when dropping the pitch, in which case
+       network connection would eventually timeout.
+
+       Thus, we use [`Dynamic] here to make sure the operator fails to be used in
+       such cases, even though the value is not technically dynamic. This does not
+       impact the behavior with the controlling clock during runtime execution. *)
+    method self_sync = (`Dynamic, self_sync)
 
     method private start = self#connect
 
@@ -291,9 +301,7 @@ let register_input is_http =
     else ("input.ffmpeg", "Create a stream using ffmpeg")
   in
   Lang.add_operator name ~descr ~category:Lang.Input
-    ( List.filter
-        (fun (lbl, _, _, _) -> lbl <> "clock_safe")
-        (Start_stop.active_source_proto ~fallible_opt:`Nope)
+    ( Start_stop.active_source_proto ~clock_safe:false ~fallible_opt:`Nope
     @ ( if is_http then
         [
           ( "user_agent",
@@ -340,14 +348,7 @@ let register_input is_http =
           Some
             "Should the source control its own timing? Set to `true` if you \
              are having synchronization issues. Should be `false` for most \
-             typicaly cases." );
-        ( "clock_safe",
-          Lang.nullable_t Lang.bool_t,
-          Some Lang.null,
-          Some
-            "Should the source be in its own clock. Should be the same value \
-             as `self_sync` unless the source is mixed with other `clock_safe` \
-             sources like `input.ao`" );
+             typical cases." );
         ( "debug",
           Lang.bool_t,
           Some (Lang.bool false),
@@ -428,10 +429,7 @@ let register_input is_http =
       let debug = Lang.to_bool (List.assoc "debug" p) in
       let self_sync = Lang.to_bool (List.assoc "self_sync" p) in
       let autostart = Lang.to_bool (List.assoc "start" p) in
-      let clock_safe =
-        Lang.to_default_option ~default:self_sync Lang.to_bool
-          (List.assoc "clock_safe" p)
-      in
+      let clock_safe = Lang.to_bool (List.assoc "clock_safe" p) in
       let on_start =
         let f = List.assoc "on_start" p in
         fun _ -> ignore (Lang.apply f [])
--- a/src/lang/lang.ml
+++ b/src/lang/lang.ml
@@ -771,7 +771,7 @@ let source_methods =
     ( "self_sync",
       ([], fun_t [] bool_t),
       "Is the source currently controling its own real-time loop.",
-      fun s -> val_fun [] (fun _ -> bool s#self_sync) );
+      fun s -> val_fun [] (fun _ -> bool (snd s#self_sync)) );
     ( "is_up",
       ([], fun_t [] bool_t),
       "Indicate that the source can be asked to produce some data at any time. \
@@ -875,21 +875,18 @@ let add_operator =
         | x, y -> Stdlib.compare x y
     in
     let proto =
-      let t = T.make (T.Ground T.String) in
       ( "id",
-        t,
-        Some { pos = t.T.pos; value = Ground (String "") },
+        nullable_t string_t,
+        Some null,
         Some "Force the value of the source ID." )
       :: List.stable_sort compare proto
     in
     let f env =
       let src : < Source.source ; .. > = f env in
-      let id =
-        match (List.assoc "id" env).value with
-          | Ground (String s) -> s
-          | _ -> assert false
-      in
-      if id <> "" then src#set_id id;
+      ignore
+        (Option.map
+           (fun id -> src#set_id id)
+           (to_valued_option to_string (List.assoc "id" env)));
       let v = source (src :> Source.source) in
       _meth v (List.map (fun (name, _, _, fn) -> (name, fn src)) meth)
     in
--- a/src/lang/lang_builtins.ml
+++ b/src/lang/lang_builtins.ml
@@ -141,8 +141,8 @@ let () =
   let proto =
     [
       ( "id",
-        Lang.string_t,
-        Some (Lang.string ""),
+        Lang.nullable_t Lang.string_t,
+        Some Lang.null,
         Some
           "Identifier for the new clock. The default empty string means that \
            the identifier of the first source will be used." );
@@ -161,7 +161,7 @@ let () =
     match l with
       | [] -> Lang.unit
       | hd :: _ as sources ->
-          let id = if id = "" then (Lang.to_source hd)#id else id in
+          let id = Option.value ~default:(Lang.to_source hd)#id id in
           let sync =
             match Lang.to_string sync with
               | s when s = "auto" -> `Auto
@@ -195,7 +195,7 @@ let () =
       ] )
     Lang.unit_t
     (fun p ->
-      let id = Lang.to_string (List.assoc "id" p) in
+      let id = Lang.to_valued_option Lang.to_string (List.assoc "id" p) in
       let sync = List.assoc "sync" p in
       let l = Lang.to_list (List.assoc "" p) in
       assign id sync l)
--- a/src/operators/cross.ml
+++ b/src/operators/cross.ml
@@ -31,10 +31,13 @@ let finalise_child_clock child_clock source =
 (** [rms_width] and [minimum_length] are all in samples.
   * [cross_length] is in ticks (like #remaining estimations).
   * We are assuming a fixed audio kind -- at least for now. *)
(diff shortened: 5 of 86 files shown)


smimram:
@toots I filled in the parameters I could, but I would need a serious review on this one: I am far from being sure that I filled everything correctly or that I did not miss something...

A way to review is
```
LIQ_FREI0R=false LIQ_LADSPA=false ./liquidsoap --no-stdlib --list-functions-md 
```
At the end of the doc of operators there is now something like
```
This function is passive, fallible, not self-syncing and clock-safe.
```

libs/hls.liq
```diff
@@ -6,10 +6,11 @@ let input.hls = ()
 # @param ~reload How often (in seconds) the playlist should be reloaded.
 # @param uri Playlist URI.
 # @flag experimental
-def input.hls.native(~id="",~reload=10.,uri)
+def input.hls.native(~id=null(),~reload=10.,uri)
   playlistr = ref([])
   sequence = ref(0)
   playlist_uri = ref(uri)
+  label_id = id ?? "input.hls.native"
```
smimram, line 13: The correct way to do this is
```
id = string.id.default("input.hls.native", id)
```
If `id` is `null` it will generate an identifier of the form `input.hls.native_NNN` in the same way as usual sources do.

libs/hls.liq
```diff
@@ -18,7 +19,7 @@ def input.hls.native(~id="",~reload=10.,uri)
 
       m = string.extract(pattern="#EXT-X-MEDIA-SEQUENCE:(\\d+)",file.contents(pl))
       pl_sequence = m[1]
-      log.info(label=id,"Sequence: " ^ pl_sequence)
+      log.info(label=label_id,"Sequence: " ^ pl_sequence)
```
smimram, line 22:
```suggestion
      log.info(label=id,"Sequence: " ^ pl_sequence)
```

libs/hls.liq
```diff
@@ -45,7 +46,7 @@ def input.hls.native(~id="",~reload=10.,uri)
 
       playlistr := list.fold(add_file, !playlistr, files)
     else
-      log.severe(label=id,"Couldn't read playlist: request resolution failed.")
+      log.severe(label=label_id,"Couldn't read playlist: request resolution failed.")
```
smimram, line 49:
```suggestion
      log.severe(label=id,"Couldn't read playlist: request resolution failed.")
```

libs/hls.liq
```diff
@@ -73,7 +74,7 @@ def input.hls.native(~id="",~reload=10.,uri)
       if not (string.contains(substring="/", !playlist_uri)) then
         playlist_uri := path.dirname(request.uri(pl)) ^ "/" ^ !playlist_uri
       end
-      log(label=id,"Playlist: " ^ !playlist_uri)
+      log(label=label_id,"Playlist: " ^ !playlist_uri)
```
smimram, line 77:
```suggestion
      log(label=id,"Playlist: " ^ !playlist_uri)
```

src/io/ffmpeg_io.ml
```diff
@@ -54,7 +54,17 @@ class input ?(name = "input.ffmpeg") ~autostart ~self_sync ~poll_delay ~debug
...
+       that control latency such as [cross], [soundtouch] and etc. It's not just
+       about pulling data at a higher rate than real-time, which would work on a
+       regular HTTP server-based media file, it's also about pulling data at a
+       lower rate than real-time, typically when dropping the pitch, in which case
+       network connection would eventually timeout.
+
+       Thus, we use [`Dynamic] here to make sure the operator fails to be used in
+       such cases, even though the value is not technically dynamic. This does not
+       impact the behavior with the controlling clock during runtime execution. *)
+    method self_sync = (`Dynamic, self_sync)
```
smimram, line 67: Very good catch and idea! However, could we come up with something more explicit and clear than `Static` / `Dynamic`?

src/lang/lang.ml
```diff
@@ -771,7 +771,7 @@ let source_methods =
     ( "self_sync",
       ([], fun_t [] bool_t),
       "Is the source currently controling its own real-time loop.",
-      fun s -> val_fun [] (fun _ -> bool s#self_sync) );
+      fun s -> val_fun [] (fun _ -> bool (snd s#self_sync)) );
```
smimram, line 774: Maybe would it be better to add two fields instead of having `s#self_sync` a pair, which forces us using `snd`?

src/lang/lang_builtins.ml
```diff
@@ -141,8 +141,8 @@ let () =
   let proto =
     [
       ( "id",
-        Lang.string_t,
-        Some (Lang.string ""),
+        Lang.nullable_t Lang.string_t,
```
smimram, line 144: I was also tempted by this change, but didn't dare ;)

src/operators/cross.ml
```diff
@@ -131,47 +134,26 @@ class cross ~kind (s : source) ~cross_length ~override_duration ~rms_width
...
-       * which multiples clocks wouldn't ensure. *)
-      Clock.unify child_clock s#clock;
-      Lang.iter_sources (fun s -> Clock.unify child_clock s#clock) transition;
-
-      (* Make sure the child clock can be garbage collected, cf. cue_cut(). *)
-      Gc.finalise (finalise_child_clock child_clock) self
+      child_support#set_clock;
+      Lang.iter_sources
+        (fun s -> Clock.unify self#child_clock s#clock)
+        transition
```
smimram, line 140: This looks like a "big change", can you explain it further? Are you sure that there are no potential side effects?

src/operators/pipe.ml
```diff
@@ -236,18 +239,6 @@ class pipe ~kind ~replay_delay ~data_len ~process ~bufferize ~log_overfull ~max
...
-      source#after_output
-
-    (* See smactross.ml for details. *)
-    method private set_clock =
-      let child_clock = Clock.create_known (new Clock.clock self#id) in
-      Clock.unify self#clock
-        (Clock.create_unknown ~sources:[] ~sub_clocks:[child_clock]);
-      Clock.unify child_clock source#clock;
-      Gc.finalise (fun self -> Clock.forget self#clock child_clock) self
-
```
smimram, line 250: There is really no need for explicit clock handling anymore, really?

src/tools/child_support.ml
```diff
@@ -0,0 +1,72 @@
...
+      Gc.finalise (finalise_child_clock self#child_clock) self
+
+    method private child_tick =
+      (Clock.get self#child_clock)#end_tick;
+      needs_tick <- false
+
+    method after_output =
+      if needs_tick then self#child_tick;
+      needs_tick <- true
+  end
```
smimram, line 72: Please, please, please add comments for those tricky functions while you have al the details in mind: what are they doing? when should we use them? Ideally, we should be able to write a child-controlling source from the doc. Sorry for insisting, but I know this is the kind of details which quickly get forgotten :)

src/tools/start_stop.ml
```diff
@@ -43,24 +46,22 @@ class virtual base ~(on_start : unit -> unit) ~(on_stop : unit -> unit) =
...
             on_stop ();
             state <- `Stopped
-        | _ -> ()
+        | `Stopped, `Idle -> state <- `Stopped
+        | `Stopped, `Stopped -> ()
+        | `Idle, `Started ->
+            self#stop;
+            on_stop ();
+            state <- `Idle
+        | `Idle, `Stopped | `Idle, `Idle -> ()
```
smimram, line 64: Shouldn't we have an error (or at least a message in the logs) for `Idle -> Stopped`?

tests/language/type_errors.pl
```diff
@@ -180,4 +180,11 @@ sub correct {
         %audio.copy,
         %video.copy)');
 
+# The following is not technically checking on type errors but runtime invalid values.
```
smimram, line 183:
```suggestion
# The following is not technically checking on type errors but runtime invalid values, see #1705.
```

tests/media/test_ffmpeg_inline_encode_decode.liq
```diff
@@ -74,7 +80,9 @@ output.file(
   s
 )
 
-s = ffmpeg.decode.audio_video(s)
+s = ffmpeg.decode.audio_video(id="decoder", s)
+
+clock.assign_new(sync='none',[s])
```
smimram, line 85:
```suggestion
clock(sync='none', s)
```

tests/media/test_ffmpeg_inline_encode_decode_video.liq
```diff
@@ -73,6 +78,8 @@ output.file(
 
 s = ffmpeg.decode.video(s)
 
+clock.assign_new(sync='none',[s])
```
smimram, line 81:
```suggestion
clock(sync='none', s)
```

tests/streams/autostart.liq
```diff
@@ -0,0 +1,31 @@
...
+  test_no_autostart := false
+end
+
+def on_wanted_autostart() =
+  test_autostart := true
+end
+
+s = blank()
+
+clock.assign_new(sync="none",[s])
```
smimram, line 18:
```suggestion
clock(sync="none", s)
```

tests/streams/stretch-clock-propagation.liq
```diff
@@ -0,0 +1,21 @@
...
+
+def on_stop() =
+  test.pass()
+end
+
+output.dummy(fallible=true, on_stop=on_stop, s)
+
+s = stretch(ratio=2., s)
+
+clock.assign_new(sync="none",[s])
```
smimram, line 19:
```suggestion
clock(sync="none", s)
```

tests/streams/stretch-clock-propagation.liq
```diff
@@ -0,0 +1,21 @@
+#!../../src/liquidsoap ../../libs/stdlib.liq ../../libs/deprecations.liq
+
+# Stretch (as well as soundtouch) uses the Child_support clock control
+# mechamism. We want to make sure that it properly propagates end-of-tracks
+# to its underlying outputs/sources.
+
+%include "test.liq"
+
+s = once(blank(duration=10.))
```
smimram, line 9: No need to have longer tests :)
```suggestion
s = once(blank(duration=2.))
```

libs/hls.liq
toots: We need the name change, `id` is now `string?` but `label_id` is `string`.

tests/media/test_ffmpeg_inline_encode_decode.liq
toots: This operator is actually giving me trouble. It. returns a new source, which needs to be properly hooked-up. Any reason not to use `clock.assign_new`?

src/lang/lang.ml
toots: Two methods? That was my original approach but it would require to modify every single operator. Lot of changes, chances for mistakes while the inference logic is actually exactly the same for 90% of them..

src/lang/lang_builtins.ml
toots: Well I was on a wild run :-)

src/operators/cross.ml
toots: This is actually just a factorization. All the code that was there is now in `Child_support`. I also added some thoughts about what was done and why in `clock.ml`.

src/operators/pipe.ml
toots: Same thing, checkout this file: https://github.com/savonet/liquidsoap/blob/40057dce0da8f54eec4b6af0ca951aa3ce635323/src/tools/child_support.ml

src/tools/child_support.ml
toots: It's here: https://github.com/savonet/liquidsoap/blob/40057dce0da8f54eec4b6af0ca951aa3ce635323/src/clock.mli#L81 I'll add a pointer in this file.

src/tools/start_stop.ml
toots: No, `Idle` is an automated stop that was not ordered by the user. Asking to `Stop` from there just makes the `Stop` explicit. The difference is that the output will not restart automatically once the underlying source gets ready again.

libs/hls.liq
toots: Ok!

tests/streams/stretch-clock-propagation.liq
toots: Is it important?

toots:
Thanks for the review! I think the most important comments have been addressed. I want to think about renaming `` `Dynamic``, I think the concept is clear with this name, renaming it just for one use-case doesn't seem like a good decision.

toots:
> Thanks for the review! I think the most important comments have been addressed. I want to think about renaming `` `Dynamic ``, I think the concept is clear with this name, renaming it just for one use-case doesn't seem like a good decision.

Yeah, after thinking about it, I don't think that I want to change `` `Dynamic`` just yet. The logic with _just_ static vs. dynamic is already tricky. Took me a couple of tries to get the code right for the case of a list of sources:
```ocaml
let self_sync_type sources =
  fst
    (List.fold_left
       (fun cur s ->
         match (cur, s#self_sync) with
           | (`Static, None), (`Static, v) -> (`Static, Some v)
           | (`Static, Some v), (`Static, v') when v = v' -> (`Static, Some v)
           | _ -> (`Dynamic, None))
       (`Static, None)
       sources)
```
I'd like to keep the concepts clear for this reason.

tests/media/test_ffmpeg_inline_encode_decode.liq
smimram: Only that it is shorter :) But you can forget about it!

tests/streams/stretch-clock-propagation.liq
smimram: No.

smimram (approved):

smimram:
Seems good to me, you can merge, thanks!

toots:
Thanks!

PR #1722, opened 2021-07-03
Export settings module

toots:
This PR exports settings as a big module. Each setting from dtools is of the type:
```ruby
(unit -> 'a).{ 
  name: string, 
  description: string,
  set: ('a) ->  unit,
  ... sub settings ...
}
```
This provides type-safety for settings. 

The whole `stdlib` is converted to this new pattern. `register`, sadly, cannot be ported to this model. Instead, we define static module extension for our own settings. Because of that, it is also safer to completely remove `set` and `get` to avoid further confusion. Hopefully, the changes required for the user scripts aren't too painful.

Additionally, command-line options are cleaned with only `--conf-descr-md` remaining and renamed to `--print-settings`.

Also added in this PR are the following source and output methods:
* `log_level : unit -> int?`
* `set_log_level : int -> unit`

This allows to check and. most importantly, to set individual source level. This was always possible but never clear how to do it...

Still TODO:
* ~~Make sure exported settings doc takes all settings (built-in and defined through `stdlib`) into account~~
* ~~Update doc.~~


smimram (commented):
This looks fine to me. I only have minor concerns about the naming of some functions.

Also, how much backward compatbile is this? It would be reassuring if we could still use `set` with a warning during the transition period...

toots:
> This looks fine to me. I only have minor concerns about the naming of some functions.
> 
> Also, how much backward compatbile is this? It would be reassuring if we could still use `set` with a warning during the transition period...

Thanks for the review. Yeah, I think we can try to add a backward compatible function it'll just have to be on the ML side.

PR #1725, opened 2021-07-05
Fix early self sync check

toots:
Self synced is checked too early in `Child_support`. This fixes it by pushing the call to the `wake_up` phase.

### DIFF
--- a/src/operators/switch.ml
+++ b/src/operators/switch.ml
@@ -138,14 +138,18 @@ class virtual switch ~kind ~name ~override_meta ~transition_length
 
     method is_ready = need_eot || selected <> None || self#cached_select <> None
 
+    (* This one is tricky. We do not want to call #cached_select as
+       this requires some run-time info from underlying sources
+       (mostly ctype to be set). The only case that matters if no
+       sources are selected is to know if we are [`Static, false] for
+       any caller sus as [cross]. Since the source is not ready, we
+       can return anything so we check if any source might be not
+       self_sync in this case *)
     method self_sync =
       ( Lazy.force self_sync_type,
         match selected with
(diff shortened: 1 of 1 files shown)


codeofdusk:
```
2021/07/05 04:29:28 [clock.alsa:2
2021/07/05 04:29:28 [clock:2] Erraracters 10-16: Assertion failed!
```

toots:
> sources/request_source.ml

Do you have a full stack trace?

codeofdusk:
How do I generate one?

toots:
> How do I generate one?

```
set("log.level", 4)
```
I got it, though. Give me a minute to think about how to properly fix it. Meanwhile, I've remove the beta2 release. Again, thanks for reporting, that really helps.

toots:
@codeofdusk I think I've got it this time. Would you mind checking again? Thanks!!

src/operators/switch.ml
```diff
@@ -138,14 +138,18 @@ class virtual switch ~kind ~name ~override_meta ~transition_length
 
     method is_ready = need_eot || selected <> None || self#cached_select <> None
 
+    (* This one is tricky. We do not want to call #cached_select as
+       this requires some run-time info from underlying sources
+       (mostly ctype to be set). The only case that matters if no
+       sources are selected is to know if we are [`Static, false] for
+       any caller sus as [cross]. Since the source is not ready, we
```
toots, line 145:
```suggestion
       any caller such as [cross]. Since the source is not ready, we
```

toots:
I can confirm the fix on my end, gonna go ahead and re-generate the beta2 release assets with it, thanks for reporting.

PR #1791, opened 2021-08-03
Switch to using the number of requests when queueing new requests.

toots:
This is probably one of the most confusing aspect of liquidsoap: how requests are queued.

The current schema uses a combination of:
* `length`, to express how much media length, in seconds should be available in the queue
* `default_duration` to assign a default duration to newly queued requests when not provided. Historically, because computing requests length used to be very costly, this is in practice the assigned duration of any new requests before it starts being played.
* `conservative` to enforce (?) strict duration computation.

The result is a very confusing algorithm where it becomes pretty tricky to say: "I want one/two/3 requests in the queue at all time".

This PR changes this behavior to simply implement that: let the user say, I want one, two, 3, .. requests in the queue at all time.

This is a breaking change but I believe it should make everyone's life MUCH easier.

### DIFF
--- a/doc/content/migrating.md
+++ b/doc/content/migrating.md
@@ -115,6 +115,15 @@ s.insert_metadata([("artist", "Bob")])
 output.pulseaudio(s)
 ```
 
+### Request-based queueing
+
+Queueing for request-based sources has been simplified. The `default_duration` and `length` have been removed in favor of
+a simpler implementation. You can now pass a `prefetch` parameter which tells the source how many requests should be queued
+in advance.
+
+Should you need more advanced queueing strategy, `request.dynamic.list` and `request.dynamic` now export functions to retrieve 
+and set their own queue of requests.
+
 ### Deprecated operators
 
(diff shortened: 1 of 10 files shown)


smimram:
I agree that this is quite confusing. However, having a minimum duration could be quite handy, especially if we have short or long songs (e.g. user contributed). I would propose to
- have a `size` parameter to indicate the minimum number of songs in the queue, say 2 by default (please don't use `length` to make it clear that the parameter has changed) as you propose
- keep a `duration` parameter to in order to indicate the minimum requested duration (= the old behavior), set to 0 by default
Queueing constraint is the conjunction of these two parameters.

The `conservative` parameter indicates whether we should take currently remaining time in account when buffering. I propose that we simply drop it and always use the duration of the current file.

toots:
Thanks for the review, @smimram. I'm gonna add an optional callback to control when to queue more requests and leave whatever other queueing strategy to the end user.

smimram:
Oh, queueing in Liq side, great idea!

PR #1800, opened 2021-08-08
Detect unused arguments in functions.

### LINKED ISSUE #1249: Bug with unused variable location
The following code:
```
err = 123
ignore(err)

def f(err) =
  "gni"
end
```
Results in:
```
% ./liquidsoap /tmp/bla.liq
Line 1, char 0:
Warning 4: Unused variable err
zsh: exit 1     ./liquidsoap /tmp/bla.liq
```

Just the function alone works, without the function works as well..

smimram:
We now report arguments which are not used in functions (unless named `_`).

This fixes #1249 : the problem was that `err` argument was not used, but
reported at the wrong position.

### DIFF
--- a/.ocamlformat
+++ b/.ocamlformat
@@ -1,4 +1,4 @@
-version=0.18.0
+version=0.19.0
 profile = conventional
 break-separators = after
 space-around-lists = false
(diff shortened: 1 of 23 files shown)


smimram:
@toots it seems to be almost ready excepting for some `%argsof` which apparently generate empty variable names. Could you please have a look at it?

smimram:
I found the problem: we were detecting unused arguments in functions provided as default values. The PR should now be ready to merge.

toots (approved):
This is awesome, thank you!

PR #1810, opened 2021-08-12
Include parametrized by environment variables.

smimram:
Allow incudes of the form
```
%include getenv("MY_VAR")
```


toots:
We found an alternative solution that avoids this hack. Thanks for checking.

PR #1824, opened 2021-08-16
Add support for abstract generic json object.

### LINKED ISSUE #1823: Number representation in `json_of`
**Describe the bug**
If `float` is `infinity` or `nan` it's converted to invalid `json`, because there's no way in `json` to represent infinity.
Official json number representation scheme
Also there's json5, that allows `NaN`(not `nan`), `-NaN`, `Infinity` and `-Infinity`.

For example in [python](https://docs.python.org/3/library/json.html#basic-usage) there's a named parameter `allow_nan` that allows this.

**To Reproduce**
`print(json_of(compact = true, {a = infinity}))` -> `{"a":inf}`.
`print(json_of(compact = true, {a = sqrt(-1)}))` -> `{"a":-nan}`.
This is incorrect.

**Expected behavior**
`print(json_of(compact = true, {a = infinity}))` -> `{"a":null}`
`print(json_of(compact = true, {a = sqrt(-1)}))` -> `{"a":null}`
Maybe there should be `json5` switch. Or some options like in python.
`print(json_of(compact = true, json5 = true, {a = infinity}))` -> `{"a":Infinity}`
`print(json_of(compact = true, json5 = true, {a = sqrt(-1)}))` -> `{"a":NaN}`

**Version details**
 - OS: `ubuntu:20.04` in docker
 - Version `2.0.0-beta.3`

**Install method**
Deb package from liquidsoap releases at github

**Common issues**
N/A

toots:
This PR implements a couple of well-needed JSON improvements:
* Add optional support for `json5` style `NaN` and `Infinity` floats, revert float output to be compliant in the old case
* Add support for generic json object manipulation
* Cleanup json export to export `(string, 'a)` lists as lists
* Allow values decorated with methods to be exported as values, except for `()` (records) which as kepts exported as objects
* Renamed, with deprecations, `json_of` into `json.stringify` and `of_json` into `json.parse`

Test script for generic JSON object:
```
# j = json();;
# j.add("foo", 1);;
# j.add("bla", "baz");;
# j.add("gni", 3.14);;
# j.remove("foo");;
# json.stringify(j);;
- : string = "{ \"gni\": 3.14, \"bla\": \"baz\" }"
```

TODO:
* ~~Rename `json_of` `json.stringify` and `of_json`, `json.parse` (with deprecations)~~
* ~~Fix #1823~~
* ~~Document~~ 
* ~~Test~~

### DIFF
--- a/src/lang/builtins_json.ml
+++ b/src/lang/builtins_json.ml
@@ -22,7 +22,7 @@
 
 let log = Log.make ["lang"; "json"]
 let print_s s = Utils.escape_string (fun x -> Utils.escape_utf8 x) s
-let to_json_ref = ref (fun ~compact:_ _ -> assert false)
+let to_json_ref = ref (fun ~compact:_ ~json5:_ _ -> assert false)
 
 module JSON = Lang.MkAbstract (struct
   type content = (string, Lang.value) Hashtbl.t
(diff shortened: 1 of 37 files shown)


src/lang/builtins_json.ml
```diff
@@ -21,62 +21,73 @@
...
+        let v = String.concat "," (List.map to_json_compact l) in
+        match l with
+          | {
+              Lang.value =
+                Lang.Tuple
+                  [{ Lang.value = Lang.Ground (Lang.Ground.String _) }; _];
+            }
+            :: _ ->
+              Printf.sprintf "{%s}" v
+          | _ -> Printf.sprintf "[%s]" v)
```
toots, line 56: This should be removed now, it is unpredictable (empty `[string, 'a]` list will convert to `[]`) and is now superseded.

src/lang/builtins_json.ml
```diff
@@ -21,62 +21,73 @@
...
+              raise
+                (Lang_values.Runtime_error
+                   {
+                     Lang_values.kind = "json";
+                     msg = Some "Values with methods cannot be exported to JSON";
+                     pos =
+                       (match v.Lang_values.V.pos with
+                         | None -> []
+                         | Some pos -> [pos]);
+                   }))
```
toots, line 81: We should just export the value..

src/lang/builtins_json.ml
```diff
@@ -121,21 +132,32 @@ let rec to_json_pp f v =
...
```
toots, line 160: Same here, let's just export the value.

smimram (approved):
This is neat, great PR! I am a bit worried with backward compatibility issues but let's release and see..

PR #1825, opened 2021-08-17
Handle nullable then.

### LINKED ISSUE #1816: Type inference with `null()` in conditionals
**Describe the bug**
Nullable type inference is incorrect if I use `if-then-else-end` in wrong order

**To Reproduce**
This code sample
```
def test()
  if 0 == 0 then
    null()
  else
    1
  end
end

test()
```
will create an error
```
Error 5: this value has type
  (...) -> _?
but it should be a subtype of
  (...) -> int
```

But this code sample is ok
```
def test()
  if 0 == 0 then
    1
  else
    null()
  end
end

test()
```

The same with ternary operator.
**Expected behavior**
Allow to write `null()` in `if` in any branch.

**Version details**
 - OS: ubuntu:20.04 in docker
 - Version `2.0.0-beta.3`

**Install method**
[Liquidsoap github release deb](https://github.com/savonet/liquidsoap/releases/download/v2.0.0-beta3/liquidsoap-v2.0.0-beta3_2.0.0.beta3-ubuntu-focal-1_amd64.deb)

**Common issues**
N/A

smimram:
Add a special case in the typechecker in order to handle conditionals
where the then branch is nullable and the else branch is not. This is a
hack but a proper fix would involve deferring unification to a
constraint solver, which we are not prepared for (yet). Fixes #1816.

### DIFF
--- a/libs/ffmpeg.liq
+++ b/libs/ffmpeg.liq
@@ -50,7 +50,7 @@ def video.add_text.ffmpeg(~color=0xffffff, ~font="", ~size=18, ~x=10, ~y=10, tex
     if font == "" then
       null()
     else
-      null(font)
+      font
     end
   def mkfilter(graph)
     s = ffmpeg.filter.video.input(graph, s)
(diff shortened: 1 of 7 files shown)


toots (approved):
That looks pretty straight forward to me. It's a hack but having if/then/else as a function was already a hack to begin with...

PR #1826, opened 2021-08-17
[WIP] Proper decoder duration API

toots:


smimram:
Out of curiosity, why did you close this one?

toots:
This way way too involved. I found an easier way to do it for now here: https://github.com/savonet/liquidsoap/commit/e9adaf11caf4e8c67a0f97a27d796e0789d92131.

My original intent was to revert to a mechanism with duration + elapsed from which we get the remaining time. However, I have to admit, I'm not sure if that is the natural way to do for us. A lot of our logic relies on looking forward to remaining time.

Ideally, I'd like to cleanup this layer a bit more, though, so we can at least be exact when we have a duration and relieves the decoders of having to estimate remaining time.

PR #1848, opened 2021-08-24
Rename lang modules and split.

smimram:
Rename lang modules so that the code is nicer.

Split lang and types so that we don't have to recompile everything each time we want to change something in the typer or the language.


toots (commented):
Quick question: is it really really reasonable to have local module names as generic as `Error` and `Parser`? This is bound to clash with standard modules at some point (if it doesn't already)..

smimram:
I'd say that we would use clean module names if they are available. Most libraries do prefix module name (and in for libraries I think that this is a good idea) and we can always rename later on if we find a conflict...

toots (approved):
Nice cleanup! If some of the internal module names clashe in the future we could wrap them in in a top-level `LiquidsoapLang` module.

PR #1858, opened 2021-08-25
Delay encoders creation.

smimram:
Try to build encoder at runtime so that they can have dynamic parameters.

The bad side is that parameters are now checked at execution time: how bad is this?


src/lang/parser.mly
```diff
@@ -397,56 +380,20 @@ if_elsif:
...
+  | LPAR encoder_params RPAR { $2 }
 
-ffmpeg_opt:
-  | %prec no_app { [] }
-  | LPAR ffmpeg_list RPAR { $2 }
+encoder_param:
+  | VAR GETS expr       { $1, `Term $3 }
+  | STRING GETS expr    { $1, `Term $3 }
+  | VAR                 { "", `Term (mk ~pos:$loc (Ground (String $1))) }
+  | ENCODER encoder_opt { "", `Encoder ($1, $2) }
```
toots, line 391: Maybe also add `STRING` in case we want to do `%encoder("some non var parameter")`.

src/lang/term.ml
```diff
@@ -406,11 +418,21 @@ let rec bound_vars_pat = function
...
     | Tuple l ->
         List.fold_left (fun v a -> Vars.union v (free_vars a)) Vars.empty l
     | Null -> Vars.empty
+    | Encoder e ->
+        let rec enc (_, p) =
+          List.fold_left
+            (fun v (_, t) ->
+              match t with
+                | `Term t -> Vars.union v (free_vars t)
+                | `Encoder e -> Vars.union v (enc e))
```
toots, line 432: So, type of `%ogg` is encoder and type of `%vorbis` too and encoders can hold encoders. Gotcha!

src/lang/encoders.ml
```diff
@@ -0,0 +1,212 @@
...
+  let kind = kind_of_encoder f in
+  let audio = kind_t ~pos ~level kind.Frame.audio in
+  let video = kind_t ~pos ~level kind.Frame.video in
+  let midi = kind_t ~pos ~level kind.Frame.midi in
+  format_t ~pos ~level (frame_kind_t ~pos ~level audio video midi)
+
+exception Error
+
+(* TODO: it would be better if each encoder registered itself here than
+   having this match *)
```
toots, line 136: Another thing that can be achieved by extensible variant type. You declare a base type for encoders with whatever you need to register them, then each encoder module uses this interface to register itself with all its required bits. This would also work potentially for dynamically loaded encoders (though I'd _really_ like to avoid it).

See for instance:
```ocaml
module type AbstractDef = sig
  type content

  val name : string
  val to_json : compact:bool -> json5:bool -> content -> string
  val descr : content -> string
  val compare : content -> content -> int
end

module MkAbstract (Def : AbstractDef) = struct
  type Type.ground += Type
  type G.t += Value of Def.content

  let () =
    G.register (function
      | Value v ->
          let compare = function
            | Value v' -> Def.compare v v'
            | _ -> assert false
          in
          Some
            {
              G.descr = (fun () -> Def.descr v);
              to_json =
                (fun ~compact ~json5 () -> Def.to_json ~compact ~json5 v);
              compare;
              typ = Type;
            }
      | _ -> None);

...
```

src/lang/encoders.ml
```diff
@@ -0,0 +1,212 @@
...
+
+let make_encoder ~pos t e p =
+  let e =
+    try make_encoder_base e p
+    with Error -> raise (Term.Unsupported_format (pos, Term.print_term t))
+  in
+  (try
+     let (_ : Encoder.factory) = Encoder.get_factory e in
+     ()
+   with Not_found -> raise (Term.Unsupported_format (pos, Term.print_term t)));
```
toots, line 211: Is that gonna raise to the runtime? Should we use a catchable runtime error instead?

toots (commented):
Thanks @smimram this PR is really a game changer and makes me excited about the possibilities for the `2.1.x` branch already.

I think my most important comments are:
* Let's use extensible variant types, we won't regret it on the long run. It's a super clean and efficient pattern, using it for extending values has already proven to be super good (I have some changes adding native support for PCRE using it!)
* Let's make sure encoder errors are catchable at runtime so we can use them to drive dynamic/scriptable use of encoders.

Thanks!

src/lang/encoders.ml
smimram: This is a bit of work. Would it be ok to do this in a second PR?

src/lang/encoders.ml
smimram: I am not sure here: most of the time we will check this at startup and get an error of the form
```
Error 12: Unsupported format: %bla().
You must be missing an optional dependency.
```
I think this is more clear than a catchable error, isn't it?

src/lang/term.ml
smimram: Yes!

smimram:
Encoders now register themselves. Are we greenlighted? :)

toots:
Looks like something still isn't `%ifdef`'d:
```
At ./libs/externals.liq, line 31, char 15-42:
Error 5: this value has no method ffmpeg
```
https://github.com/savonet/liquidsoap/runs/3514290926?check_suite_focus=true

PR #1870, opened 2021-09-02
Add support for union types

smimram:
We add union types. For instance,
```
def f(x) =
  match x
    case x : string do
      ignore (x^"")
      4
    case y : int do y+2
  end
end
```
has type
```
f : (string | int) -> int
```
which means that it can be used both with `string` and `int` arguments.

I use this to add support for polymorphic variants. For instance, we can write:
```
# Two ground types
def f(x) =
  match x
    case x : string do
      ignore (x^"")
      4
    case y : int do
      y+2
  end
end

# Constructors
def g(x) =
  match x
    case `a do "a"
    case `b do "b"
    case `c do "c"
  end
end

# Constructors with a catchall
def h(x) =
  match x
    case `a do "a"
    case `b do "b"
    case `c do "c"
    case _ do "?"
  end
end

def run() =
  # Simple test
  x = match 2
    case x do 4
  end

  _ = f("x")
  _ = f(2)

  _ = g(`a)
  _ = g(`b)
  _ = g(`c)

  _ = h(`a)
  _ = h(`b)
  _ = h(`c)
  _ = h("x")
end
```
NB : we really need #1905 for this to be useful.


smimram:
Some possible extensions:
- add support for matching one string (`case "bla" do`): this requires ensuring that we have a proper catchall
- add support for arguments to constructors

PR #1876, opened 2021-09-06
Remove variance information from constructed types.

smimram:
It seems that the variance information in constructed types is never used. I propose that we simply remove that?

### DIFF
--- a/src/lang/typing.ml
+++ b/src/lang/typing.ml
@@ -91,7 +91,7 @@ let copy_with (subst : Subst.t) t =
     match t.descr with
       | EVar v -> ( try Subst.value subst v with Not_found -> t)
       | Constr c ->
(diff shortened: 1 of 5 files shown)


toots (approved):
If it passes the tests I'm all about it.

src/lang/typing.ml
```diff
@@ -156,20 +156,20 @@ let rec ( <: ) a b =
     | Constr c1, Constr c2 when c1.name = c2.name ->
         let rec aux pre p1 p2 =
           match (p1, p2) with
-            | (v, h1) :: t1, (_, h2) :: t2 ->
+            | h1 :: t1, h2 :: t2 ->
                 begin
                   try (* TODO use variance info *)
```
smimram, line 161: We actually should use the variance here...

PR #1877, opened 2021-09-06
Convert more nullable parameters.

toots:


src/operators/dyn_op.ml
```diff
@@ -138,11 +139,11 @@ let () =
           "Whether the source is infallible or not (be careful when setting \
            this, it will not be checked by the typing system)." );
       ( "resurection_time",
-        Lang.float_t,
-        Some (Lang.float 1.),
```
smimram, line 142: Do you really want to change the default value here?

toots: Good catch thks

src/operators/dyn_op.ml
```diff
@@ -138,11 +139,11 @@ let () =
...
+        Lang.nullable_t Lang.float_t,
+        Some Lang.null,
```
toots, line 143:
```suggestion
        Some (Lang.float 1.),
```

PR #1881, opened 2021-09-07
Abstract regexp API

toots:
This PR creates an abstract regexp type and adds builtin syntax for it.

Regexp can be created through a regular operator:
```ruby
r = regexp(?flags=[...], "...")
```
Or through a syntactic sugar:
```ruby
r = r/.../g
```

String methods are attached to a regexp:
```ruby
Methods:

 * exec : (string) -> [int * string]
     Extract substrings from a string. Returns a list of (index,value). If
     the list does not have a pair associated to some index, it means that
     the corresponding pattern was not found.

 * replace : (((string) -> string), string) -> string
     Replace all substrings matched by a pattern by another string returned
     by a function.

 * split : (string) -> [string]
     Split a string on the given regular expression.

 * test : (string) -> bool
     Match a string with the expressionn.
```
Those methods should behave the same was as they did with the previous `string.*` functions.

One of the bonus of the syntactic sugar is that we can now avoid the escaping madness when writing regular expressions, for instance:
```ruby
r = r/[\r]?\n/
```
is equivalent to:
```ruby
r = regexp("[\\r]?\\n")
```


smimram:
This is very cute, thanks! Is the syntax `r/.../` standard? What are other languages doing?

smimram (approved):

toots:
> This is very cute, thanks! Is the syntax `r/.../` standard? What are other languages doing?

Thanks. This was inspired by javascript, which does `/.../`. Unfortunately, I was not able to find a disambiguation for the lexer w.r.t. the division symbol. Have no idea how javascript is doing it.

I think `r/../` is actually still cool and kinda helps pointing to what this could mean.

libs/externals.liq
```diff
@@ -145,7 +145,7 @@ def enable_external_flac_decoder() =
         # If the value is not an int, this returns 0 and we are ok :)
         int_of_string(channels)
       else
-        if string.match(pattern="flac",fname) then
+        if r/flag/.test(fname) then
```
toots, line 148:
```suggestion
        if r/flac/.test(fname) then
```

PR #1905, opened 2021-09-12
Increasing types

smimram:
Testing a variant of subtyping where substituted type variables carry variance information and can subsequently be changed.

The typical use case is
```
if ... then 1 else null() end
```
Here the `if` takes two branches of type `'a` and returns `'a`. When we look at the first branch, we determine that `'a` must be `int` and therefore we raise a type error when we look at the second branch because we don't have `'b? <: 'a`. With this proposal, we record, when we substitute `'a` by `int`, that any greater type would have been ok too, and therefore allow ourselves changing, `int` to `int?` (which is ok since `int <: int?`).

It should also solve the issue of lists with heterogeneous (but compatible) elements (e.g. the elements have some methods in common but not all).


src/lang/type.ml
```diff
@@ -237,8 +252,9 @@ let repr ?(filter_out = fun _ -> false) ?(generalized = []) t : repr =
   let evar level i c =
     let constr_symbols, c = split_constr c in
     if !debug then (
-      let v = Printf.sprintf "?%s%d" constr_symbols i in
-      let v = if debug_levels then Printf.sprintf "%s[%d]" v level else v in
+      (* let v = Printf.sprintf "?%s%d" constr_symbols i in *)
```
toots, line 255: nit: remove

src/lang/type.ml
```diff
@@ -275,7 +291,8 @@ let repr ?(filter_out = fun _ -> false) ?(generalized = []) t : repr =
         | EVar (i, c) ->
             if List.exists (fun (j, _) -> j = i) g then uvar g t.level (i, c)
             else evar t.level i c
-        | Link t -> repr g t)
+        | Link (Covariant, t) when !debug -> `Misc ("[>", repr g t, "]")
```
toots, line 294: Do you think it'd be worthwhile to make this a setting? As the language gets more complex, we might need this info from users and a settings would make it easier to turn on.

src/lang/typing.ml
```diff
@@ -165,97 +169,183 @@ let rec occur_check a b =
...
+              constraints;
             constraints
-      | _ -> assert false (* only EVars are bindable *)
-    end;
+        | _ -> assert false
+      (* only EVars are bindable *)
+    in
+
+    (* We do not want to check the constraints when we increase the types (for
+       now). *)
```
toots, line 255: Constraint was always hacky I think. As the typing system gets richer, we should be able to get rid of them. Typically, `internal` could become: `pcm | yuva420p | midi | none`

src/lang/typing.ml
```diff
@@ -277,8 +367,30 @@ exception Error of (repr * repr)
...
+  if !debug || !debug_subtyping then
+    Printf.eprintf "\n%s <: %s\n%!" (print a) (print b);
+  match (a.descr, b.descr) with
+    | _, Link (Covariant, b') ->
+        (* When the variable is covariant, we take the opportunity here to correct
+           bad choices. For instance, if we took int, but then have a 'a?, we
+           change our mind and use int? instead. *)
+        let b'' = try sup ~pos:b'.pos a b' with Incompatible -> b' in
+        (* Printf.printf "the sup of %s and %s is %s\n%!" (Type.print a) *)
+        (* (Type.print b') (Type.print b''); *)
```
toots, line 379: nit: remove

src/lang/typing.ml
```diff
@@ -277,8 +367,30 @@ exception Error of (repr * repr)
...
+        (try b' <: b''
+         with _ ->
+           failwith
+             (Printf.sprintf "sup did to increase: %s !< %s" (Type.print b')
+                (Type.print b'')));
+        if b'' != b' then
+          (* Printf.printf "%s becomes %s\n%!" (Type.print b) (Type.print b''); *)
```
toots, line 386: nit: remove

src/lang/typing.ml
```diff
@@ -277,8 +367,30 @@ exception Error of (repr * repr)
...
+        (* When the variable is covariant, we take the opportunity here to correct
+           bad choices. For instance, if we took int, but then have a 'a?, we
+           change our mind and use int? instead. *)
+        let b'' = try sup ~pos:b'.pos a b' with Incompatible -> b' in
+        (* Printf.printf "the sup of %s and %s is %s\n%!" (Type.print a) *)
+        (* (Type.print b') (Type.print b''); *)
+        (try b' <: b''
+         with _ ->
+           failwith
+             (Printf.sprintf "sup did to increase: %s !< %s" (Type.print b')
```
toots, line 383: ?

src/lang/typechecking.ml
```diff
@@ -131,12 +131,7 @@ let rec check ?(print_toplevel = false) ~throw ~level ~(env : Typing.env) e =
         e.t >: t
     | List l ->
         List.iter (fun x -> check ~level ~env x) l;
-        let t =
-          List.fold_left
-            (fun t e -> Typing.sup ~pos:e.t.Type.pos ~level t e.t)
-            (Type.fresh_evar ~level ~pos)
-            l
-        in
+        let t = Type.fresh_evar ~level ~pos in
```
toots, line 134: This is real nice.

src/lang/type.ml
```diff
@@ -459,6 +476,11 @@ let print_repr f t =
     | `Range_Ellipsis ->
         Format.fprintf f "...";
         vars
+    | `Misc (a, b, c) ->
```
toots, line 479: I would suggest a more explicit name. It seems that this is only used for debugging so perhaps `` `Debug``

src/lang/typing.ml
```diff
@@ -165,97 +169,183 @@ let rec occur_check a b =
...
+                    match (demeth b).descr with
+                      | Ground g ->
+                          if g <> Int && g <> Float then
+                            raise (Unsatisfied_constraint (Num, b))
+                      | EVar (j, c) ->
+                          if List.mem Num c then ()
+                          else b.descr <- EVar (j, Num :: c)
+                      | _ -> raise (Unsatisfied_constraint (Num, b))))
+              constraints;
             constraints
```
toots, line 249: This a bit weird, you're doing: 
```ocaml
let constraints = 
  List.iter fn constraints;
  constraints
in
```
I feel like this could be rewritten.

src/lang/typing.ml
```diff
@@ -165,97 +169,183 @@ let rec occur_check a b =
...
-      | _ -> assert false (* only EVars are bindable *)
-    end;
+        | _ -> assert false
+      (* only EVars are bindable *)
+    in
+
+    (* We do not want to check the constraints when we increase the types (for
+       now). *)
+    let variance = if constraints <> [] then Invariant else variance in
```
toots, line 256: When are types `Invariant`? `int?` should be invariant, `g?` where `g` is ground. Anything ending with `?` not `'a`? Anything with constraints?

src/lang/typing.ml
```diff
@@ -165,97 +169,183 @@ let rec occur_check a b =
...
 
 (** {1 Subtype checking/inference} *)
 
+exception Incompatible
+
+(** Approximated supremum of two types. We grow the second argument so that it
+    has a chance be be greater than the first. No binding is performed by this
+    function so that it should always be followed by a subtyping. *)
+let rec sup ~pos a b =
+  (* Printf.printf "sup %s // %s\n%!" (Type.print a) (Type.print b); *)
```
toots, line 275: nit: remove

src/lang/typing.ml
```diff
@@ -165,97 +169,183 @@ let rec occur_check a b =
...
+    | _, Nullable b -> mk (Nullable (sup a b))
+    | List a, List b -> mk (List (sup a b))
+    | Arrow (p, a), Arrow (q, b) ->
+        if List.length p <> List.length q then raise Incompatible;
+        mk (Arrow (q, sup a b))
+    | Tuple l, Tuple m ->
+        if List.length l <> List.length m then raise Incompatible;
+        mk (Tuple (List.map2 sup l m))
+    | Ground g, Ground g' ->
+        (* It might happen that we compare abstract values. *)
```
toots, line 302: Indeed, this can happen with the extensive abstract type thanks for thinking of that.

toots (commented):
This looks great and a definite step in the right direction.

I have some questions to understand the implementation:
* What is the theoretical partial order on types now? Are the `Invariant` types the types that same as local maximum?
* Could we add some basic tests of type `sup` and inference? It feels like we might want to test specific tricky examples to make sure that the implementation does what it's supposed to do..

src/lang/type.ml
smimram: This is very very technical . For now, I only see this as useful for implementors of the type system.

src/lang/typing.ml
smimram: Yes, that's one of my mid-term goals with those changes!

src/lang/typing.ml
smimram: I didn't think of it, it actually bite me ;)

src/lang/typing.ml
smimram: excepting that you need the match to extract the `constraints`!

smimram:
Let me try to explain better why you need the variance information. Basically, a `'a` variable can be replaced by anything, so we have all freedom there. When you perform a unification `int <: 'a`, we replace `'a` by `int` but we keep the information that anything greater than `int` would also do (such as `int?`). However, when the comparison is in the other direction `'a <: int`, we also replace `'a` by `int`, but this time, we cannot increase `'a` after since it might not be smaller than `int` afterward. In this case, we mark the variance as `Invariant` so that we know that we don't have the right to change it. We could try to mark it as `Contravariant` and use this information, but I don't want to think too hard for now (e.g. what should we do if we have a covariant link to a contravariant link to something?...). So either we allow increasing, or we don't.

Concerning the tests, there are already quite a few of them:
```
  _ = if true then {a = 4, b = 5} else {a = 4} end
  _ = if true then {a = 4} else {a = 4, b = 5} end
```
```
 def f() = if true then null() else 1 end end
  def f() = if true then 1 else null() end end
  _ = [null(), 3]
  _ = [3, null()]
```
etc. And with our "large" codebase, it's already quite tricky to have stdlib passing ;)

src/lang/typechecking.ml
smimram: :)

PR #1922, opened 2021-09-14
Switch to Re for regular expressions

toots:
The library seems to be well maintainer, under the `ocaml` banner and has no external dependencies. Since `Pcre` is hard dependency for u, it would make sense to use a native library.


toots:
First runs looks pretty bad, closing :-)

PR #1934, opened 2021-09-15
Runtime modules!

toots:
This is fun!
```shell
% cat /tmp/module.liq
# foo is exported and defined locally
let export foo = 123

print(foo)

# gni is exported but not defined locally
export gni = { bla = 123.455 }
```

```ruby
# let import x = "/tmp/module.liq";;
x : {foo : int, gni : {bla : float}} = {foo = 123, gni = {bla = 123.455}}

# let import {foo, gni} = "/tmp/module.liq";;
gni : {bla : float} = {bla = 123.455}
foo : int = 123
```

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -208,7 +208,7 @@ synth = synth/keyboard.ml synth/synth_op.ml \
 	$(if $(W_SDL),synth/keyboard_sdl.ml)
 
 builtins = lang/lang_builtins.ml \
-           lang/builtins_ref.ml \
+           lang/builtins_ref.ml lang/builtins_importer.ml \
            lang/builtins_bool.ml lang/builtins_list.ml \
            lang/builtins_string.ml lang/builtins_regexp.ml lang/builtins_json.ml \
            lang/builtins_request.ml lang/builtins_error.ml \
--- a/src/lang/builtins_importer.ml
+++ b/src/lang/builtins_importer.ml
@@ -22,6 +22,11 @@
 
 module TypeValue = Lang.MkAbstractValue (Parser_helper.TypeTerm)
 
+let throw exn =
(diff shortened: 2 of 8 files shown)


smimram:
Small comment: why do we force the `let` binding when importing? I'd rather have
```
x = import("/tmp/module.liq")
```
so that we could even do
```
(import("/tmp/module.liq")).myfun()
```

smimram:
Also, I don't understand why we should declare the type of the imported module...

toots:
> Small comment: why do we force the `let` binding when importing? I'd rather have
> 
> ```
> x = import("/tmp/module.liq")
> ```
> 
> so that we could even do
> 
> ```
> (import("/tmp/module.liq")).myfun()
> ```

That is because of the way the implementation works. This:
```ruby
let import x = "/path/to/module.liq"

print(x.foo + 1)
```
is translated into:
```ruby
type ty = 'a
let (x:ty) = _internal_module_importer_(ty, "/path/to/module.liq")

print(x.foo + 1)
```
Let's sat that `/path/to/module.liq` contains:
```ruby
let export foo = 123

export bar = "gni"
```
Then this is translated into:
```ruby
_exports_ = {}
let _exports_.foo = 123
foo = _exports_.foo

let _exports_.bar = "gni"

_exports_
```

At runtime, we parse `/path/to/module.liq` and typecheck it against `ty`. If the script is later using the exported module with incorrect method types, we can then raise a type error at runtime.

This works because we have a special syntactic construct that informs the parser to generate a run-time reference to the the inferred type to be use later when parsing the module's content.

This construct could be anything but I think using `let import` makes sense because:
* We are already doing `let replaces `
* `let` is nice to signify. that there is a variable definition happening
Same for `export`..

PR #1943, opened 2021-09-17
Inductive types with strings

smimram:
This is a variant of #1870, where we use strings insteads of constructors. This means full backward compatibility (we write `mode="randomize"` and not ``mode=`randomize``) and simpler syntax, but I cannot manage to make it fully work for now (the stdlib is not going through...)


toots:
Why strings and not constructors?

PR #1946, opened 2021-09-17
Allow "any" as maximal kind

smimram:
This should allow using incompatible encoders as in the situation proposed in #1915.


toots:
Thanks for working on this. Give me a minute to review this. I'd like to make sure that we're not rushing to a solution here.

smimram:
Yes, of course. I wanted to make sure that this was easily doable from a type system point of view, but I am a bit worried concerning the functionning of operators, although I don't see anything bad coming...

toots (changes requested):
Let's hold on this one until we see how the general work on typing progresses. A general solution would be better than a ad-hoc fix. Meanwhile, I have explained to the original reported how to do it with the current type system.

PR #1957, opened 2021-09-20
Nicer implementation of type variables

smimram:
This cleans up the current implementation of type variables while keeping the same functionalities :
- we have a clean definition of the type `var` of variables
- levels are attached to existential variables only (this is the only place where they are used)
- the level computation is fixed (it was not handled correctly)
- substitution is done in a reference attached to a variable instead of having to preserve sharing in the surrounding `descr`
- we now use `max_int` instead of `-1` as default level, which behaves correctly wrt `min`
- many various small improvements


toots (approved):
Thanks!!

PR #1959, opened 2021-09-21
Add usual cons notation for lists.

smimram:
This allows for
```
1::2::[3,4]
```
I think we use lists enough now to justify introducing this notation.

### DIFF
--- a/libs/audio.liq
+++ b/libs/audio.liq
@@ -264,8 +264,8 @@ def replaces dtmf(~duration=0.1, ~delay=0.05, dtmf)
         (0., 0.)
       end
     s = add([sine(row, duration=duration), sine(col, duration=duration)])
-    l := list.add(blank(duration=delay), !l)
-    l := list.add(s, !l)
+    l := blank(duration=delay) :: !l
+    l := s :: !l
   end
   l = list.rev(!l)
   sequence(l)
--- a/libs/http.liq
+++ b/libs/http.liq
@@ -21,7 +21,7 @@ def http.response(~protocol="HTTP/1.1",
                  ("Connection", "close")])
   headers =
     if content_type != "" then
-      list.add(("Content-Type", content_type), headers)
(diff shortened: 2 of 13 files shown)


toots:
We already have
```
[x, y, ...z]
```
and this matches common notations from other languages.

I'm not against adding more constructs but we should use the one we have already..

smimram:
Coming from OCaml it looks much more natural for me to write
```
l = x :: l
```
than
```
l = [x; l]
```
Can you point me to the syntax in other languages?

toots:
> Coming from OCaml it looks much more natural for me to write
> 
> ```
> l = x :: l
> ```
> 
> than
> 
> ```
> l = [x; l]
> ```
> 
> Can you point me to the syntax in other languages?

Javascript has `...`:
Javascript: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax

Ruby has `*` for `...`:
Ruby: https://www.honeybadger.io/blog/ruby-splat-array-manipulation-destructuring/

Python has `...`:
https://how.wtf/spread-operator-in-python.html

PHP has `...`:
https://wiki.php.net/rfc/spread_operator_for_array

Java has a `...` notation for function variable arguments that refer back to an array:
https://www.geeksforgeeks.org/variable-arguments-varargs-in-java/

The `...` notation is easier to adopt than `::` (see [here](https://github.com/savonet/liquidsoap/discussions/1915#discussioncomment-1361991) for instance) but also is symmetric with the restructuring, which makes the language more regular, intuitive:

```ruby
let [x, y, ...z] = l
let l = [x, y, ...z]

# Only difference is when building we can do:
let l = [...x, y, ...z]
```

Also, the `...` notation is extensive: we could (and should) implement it for records:
```ruby
let x = { foo: 123, ... y }
```
And could even do tuples the same way:
```ruby
let x = (123, ...y)
```

PR #1976, opened 2021-09-24
Runtime eval!

toots:
This PR adds runtime, type-safe `eval` let patterns!

The implementation uses a runtime type variable that is used to type-check the parsed term at runtime to make sure that it matches the expected type.

This could be used to write runtime hot loader script. See below for details.

### Examples
```ruby
let eval x = "{foo = 123, gni = \"aabbcc\"}"

let eval x = "output.dummy(id='bla', blank())"

# Eval with patterns
let eval {foo, gni = [x, y]} = "{foo = 123, gni = [1,2]}"
```

### Runtime hot loader

The idea for a runtime hot loader would be to pass a string containing a code that:
* Initializes a new stream
* Returns a callback to shutdown the stream

This could be used to implement a long-needed feature from users, live reload. 

Here's a quick prototype:
```ruby
loaded_streams = ref([])

def start_stream(x) =
  let eval (stream_name, shutdown_callback) = x
  loaded_streams := (stream_name, shutdown_callback)::!loaded_streams
end

def stop_stream(name)
  if list.assoc.mem(name, !loaded_streams) then
    shutdown_callback = list.assoc(name, !loaded_streams)
    shutdown_callback()
    loaded_streams := list.assoc.remove(name, !loaded_streams)
  end
end
```

Then you would call the `start_stream` function with, for instance: `"s = output.ao(sine()); {source.shutdown(s)}"`.  This could be connected to a telnet function or a HTTP callback for instance.

### Implementation details

The implementation also refactors the `let` decorators to remove them from the list of keywords and make them a parameter of the `LET` token. This should allow for more flexibility for future decorators (I have some in store!)

### DIFF
--- a/libs/error.liq
+++ b/libs/error.liq
@@ -1,7 +1,7 @@
 let error.not_found = error.register("not_found")
 let error.assertion = error.register("assertion")
 let error.invalid = error.register("invalid")
-let error.eval = errpr.register("eval")
+let error.eval = error.register("eval")
 
 # Ensure that a condition is satisfied (raise `error.assertion` exception
 # otherwise).
--- a/libs/error.liq
+++ b/libs/error.liq
@@ -1,8 +1,7 @@
 let error.not_found = error.register("not_found")
-
 let error.assertion = error.register("assertion")
-
 let error.invalid = error.register("invalid")
+let error.eval = errpr.register("eval")
 
 # Ensure that a condition is satisfied (raise `error.assertion` exception
 # otherwise).
--- a/src/lang/parser_helper.ml
+++ b/src/lang/parser_helper.ml
@@ -25,6 +25,14 @@
 open Term
 open Ground
 
+type let_decoration = [ `None | `Recursive | `Replaces | `Eval ]
+
+let string_of_let_decoration = function
+  | `None -> ""
+  | `Recursive -> "rec"
+  | `Replaces -> "replaces"
+  | `Eval -> "eval"
+
 let gen_args_of ~only ~except ~pos get_args name =
   match Environment.get_builtin name with
     | Some ((_, t), Value.{ value = Fun (args, _, _, _) })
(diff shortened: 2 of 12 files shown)


libs/error.liq
```diff
@@ -1,8 +1,7 @@
 let error.not_found = error.register("not_found")
-
 let error.assertion = error.register("assertion")
-
 let error.invalid = error.register("invalid")
+let error.eval = errpr.register("eval")
```
smimram, line 4: Did you try to test this? ;)
```suggestion
let error.eval = error.register("eval")
```

src/lang/runtime.mli
```diff
@@ -46,4 +46,4 @@ val from_string : ?parse_only:bool -> lib:bool -> string -> unit
 val interactive : unit -> unit
 
 (** Evaluate a string *)
```
smimram, line 48:
```suggestion
(** Evaluate a string. The result is checked to have the given type. *)
```

smimram (changes requested):
A small request: can you update the emacs mode in order to color `eval` (and other `let` qualifiers)?

A deeper problem. I am not convinced that this is actually type safe. Let me try this:
```
def f(s)
  let eval x = s
  x
end

x = f("").bla
```
Oh
```
Error 13: Internal error: invoked method bla not found
```
sorry ;)

src/lang/parser_helper.ml
```diff
@@ -168,16 +176,51 @@ let mk_fun ~pos args body =
...
+      | PVar l when l <> [] -> List.hd (List.rev l)
+      | _ -> assert false
+  in
   let bound = List.map (fun (_, x, _, _) -> x) args in
   let bound = name :: bound in
   let fv = Term.free_vars ~bound body in
   mk ~pos (RFun (name, fv, args, body))
 
+let mk_eval ~pos (doc, pat, def, body) =
+  let ty = Type.var () in
```
smimram, line 191: Didn't try it (I thought of it after my counter-example) but I think that
```suggestion
  let ty = Type.var ~level:(-1) () in
```
might work since it will prevent type generalizations...

src/lang/parser_helper.ml
```diff
@@ -188,7 +188,7 @@ let mk_rec_fun ~pos pat args body =
   mk ~pos (RFun (name, fv, args, body))
 
 let mk_eval ~pos (doc, pat, def, body) =
-  let ty = Type.var () in
+  let ty = Type.var ~level:(-1) () in
```
smimram, line 191: Also it's a good idea to put a position information for errors here if you can:
```suggestion
  let ty = Type.var ~level:(-1) ~pos () in
```

toots:
This is looking good, will merge in a little while unless there's any objection to it!

PR #1988, opened 2021-09-28
Playlist cleanup

toots:
This PR:

* Merges `playlist.relodable` with `playlist`. Now that we have methods, it doesn't really make sense to have two different operators
* Adds optional `empty_queue` argument to `reload` function, set to `true` by default, to make sure we flush the internal playlist queue when calling a reload, which is what most users would expect to happen.


smimram:
It would be nice to have a "deprecated" implementation.

toots:
> It would be nice to have a "deprecated" implementation. Apart from this it's all good to me!

Sure thing. However, the current implementation was already not following the documentation since it was only returning a source, not a pair `(reload, source)` as stated. I'll just default to `playlist` for it.

PR #1989, opened 2021-09-29
Json type decorators && record/modules type annotation

toots:
This is a preliminary PR in anticipation for the new json parser/renderer. It adds JSON description annotation to the types, which will be used to direct the json parser and renderer to handle two specific cases:

**Json object with arbitrary string as names**

This is handled via the new `json_name` on methods. The corresponding syntax is:
```ruby
# This maps to {foo : int}
(x : { foo("arbitrary name") : int })
```

**Json objects as associative lists**

This is handled via the new top-level `json_repr` on types. The corresponding syntax is:
```ruby
# This maps to [string * int ]
(x : {|int|})
```

**Record type annotation**

Looks like we did not have record type annotation before.. 😅


smimram:
This is very cool, thanks!

A few comments.
- The idea of using records for methods parameters is a huge improvement in terms of readability.
- Adding `json_repr` to every type is a bit heavy I think. Couldn't we add a new type for associative lists (via type constructors for instance) with a subtyping to `[string * a]` lists?
- I don't like the syntax `{ foo("arbitrary name") : int }`. Maybe more `{ "arbitrary name" as foo : int }`?
- I am not a huge fan of ading a completely new syntax for assocative list (`{|int|}`). I am not sure I understand what is wrong with using `[string * int]` (sorry I know that you already explained to me). Otherwise I'd go for something like `associative(int)`? (which would be compatible with my above proposal of implementing that with constructors...

PR #2011, opened 2021-10-09
Reimplement JSON parser, factor out json renderer.

toots:
This PR is another step toward improving json parsing and rendering.

Here, we implement our own json parser/lexer, based on the specs from https://www.json.org/json-en.html. The parser is tested against the test suite from https://github.com/nst/JSONTestSuite.

We also implement parsing/rendering of JSON5 values. The parser is tested against the test suite from: https://github.com/json5/json5-tests

We also factor out the logic for JSON rendering, which will prove useful with the upcoming type-driven parser/renderer.

There is a breaking change: a runtime error is raised when trying to convert liquidsoap values that cannot be represented as JSON.

This is so we can have full iso representation between what we can parse and render in JSON so that we can always guarantee that you can parse back any value that was exported as JSON.

#### References:
* https://www.json.org/json-en.html
* http://seriot.ch/projects/parsing_json.html
* https://github.com/nst/JSONTestSuite.
* https://json5.org/
* https://spec.json5.org/
* https://github.com/json5/json5-tests


toots:
> Didn't dive yet into it, but can you explain why yojson was not enough?

The JSON specification is small enough to be implemented on our end. On the long term, I expect this to be beneficial by giving us more flexibility to e.g. implement the JSON5 superset spec, accepting certain ad-hoc values if we see a need for it and avoid external dependencies.

smimram (approved):
Looks fine to me thanks! (I trust you on the details)

PR #2022, opened 2021-10-17
Split out repr and rename print to to_string.

smimram:
More splitting of `Type` :
- we move representation-related operations to `Repr`
- we change `print` to the more standard `string_of`

This PR is mostly harmless and its purpose is only to make sure that we avoid conflicts with other current PRs.


toots (approved):
Great thanks! Let's merge this one then I'll fix https://github.com/savonet/liquidsoap/pull/1908 and we should merge it too.

PR #2066, opened 2021-11-20
Update to default port 8000 instead 8080

rrolla:
By default it uses port 8000 instead of 8080

From logs
```bash
2021/11/20 22:57:14 [server.harbor:3] Website should be ready at <http://localhost:8000/telnet>.
2021/11/20 22:57:14 [interactive.harbor:3] Website should be ready at <http://localhost:8000/interactive>.
```


toots:
Thanks for this!

PR #2076, opened 2021-11-23
Undeprecate `json.stringify` and `json.parse`

smimram:
They are quite useful and cannot be found in the doc currently. We can still advise power users to use the `let` syntax.


src/lang/builtins_json.ml
```diff
@@ -329,8 +329,7 @@ let () =
 
 let () =
   Lang.add_builtin "json.stringify" ~category:`String
-    ~flags:[`Hidden; `Deprecated]
-    ~descr:"Deprecated. Please use `let json.stringify ... instead."
+    ~descr:"Convert a value to a json string."
```
smimram, line 332:
```suggestion
    ~descr:"Convert a value to a JSON string."
```

toots (commented):
I'm fine with moving `json.stringify` back into the main library but using `let json.stringify` underneath so we have only one implementation:
```ruby
def json.stringify(s) =
  let json.stringify s = x
  s
end
```

Do you have a strong case for keeping `json.parse`? It is ugly and hard to understand, there's not way you can use this one is a one-liner. I think it needs to go.

smimram:
Well, the function `json.parse` does not show in the API otherwise, and the `let` syntax can be scary for newcomers.

For simple cases, it does the job:
```
  song     = process.read("./next-song-json")
  song     = json.parse(default=[("","")], song)
```
Would you be ok if we kept it for now?

toots:
This is closed?

smimram:
Yes in a8e26e946e036f8e472d351a5c31f0f31c202a4b.

PR #2077, opened 2021-11-23
Undeprecate string.* regexp-related functions.

smimram:
While I like the `regexp.function` syntax, it is a bit exotic and difficult to find in the doc. I propose that we keep the old functions for now, it does not cost anything and we can more easily find string-related functions.

### DIFF
--- a/libs/string.liq
+++ b/libs/string.liq
@@ -1,3 +1,38 @@
+# Match a string with an expression. Perl compatible regular expressions are
+# recognized. Hence, special characters should be escaped.
+# @category String
+def string.match(~pattern, s) =
+  deprecated("string.match", "r/_/.test(_)")
+  regexp(pattern).test(s)
+end
+
+# Extract substrings from a string. Perl compatible regular expressions are
+# recognized. Hence, special characters should be escaped. Returns a list of
+# (index,value). If the list does not have a pair associated to some index, it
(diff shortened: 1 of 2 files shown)


libs/string.liq
```diff
@@ -1,3 +1,38 @@
+# Match a string with an expression. Perl compatible regular expressions are
+# recognized. Hence, special characters should be escaped.
+# @category String
+def string.match(~pattern, s) =
+  deprecated("string.match", "r/_/.test(_)")
```
smimram, line 5:
```suggestion
```

libs/string.liq
```diff
@@ -1,3 +1,38 @@
...
+# @param s String whose substrings should be replaced.
+def string.replace(~pattern, f, s) =
+  regexp(flags=["g"], pattern).replace(f, s)
+end
+
+# Split a string at "separator". Perl compatible regular expressions are
+# recognized. Hence, special characters should be escaped.
+# @category String
+def string.split(~separator, s) =
+  deprecated("string.split", "r/_/.split(_)")
```
smimram, line 32:
```suggestion
```

toots (approved):
Fine by me. Could you also add a reference to the `/.../` syntax in the function's doc? That would help spreading awareness about them.

PR #2085, opened 2021-12-02
Switch to Re for pcre.

toots:


smimram:
So, does it work in the end?

glondu:
I gave it a try. I didn't succeed, but I've pushed it to my fork nonetheless for reference: https://github.com/glondu/liquidsoap/commit/23ee74272267251ef2dd600146dff4184701dd1e. See the commit message for more information on why I failed.

toots:
Thanks. I'l resume work as well. It would be nice to remove a required C dependency.

toots:
@glondu the absence of support for named group is gonna be a deal breaker for us, I'm afraid. We use that to match urls in the HTTP handles `/path/:id`. It's a pretty big missing feature in my opinion.

I'll have a look at migrating `ocaml-pcre` to pcre2.

toots:
Reported here: https://github.com/ocaml/ocaml-re/issues/221

glondu:
The named group feature seemed easy enough to implement in re, so I did it in https://github.com/ocaml/ocaml-re/pull/223.

However, there is still an issue with:

    Pcre.split ~rex:(Pcre.regexp "#{(.*?)}") "#{artist} - #{title}"

which is used in liquidsoap, for which I didn't find a documented semantics (and behaves differently in pcre and re).

toots:
> The named group feature seemed easy enough to implement in re, so I did it in [ocaml/ocaml-re#223](https://github.com/ocaml/ocaml-re/pull/223).
> 
> However, there is still an issue with:
> 
> ```
> Pcre.split ~rex:(Pcre.regexp "#{(.*?)}") "#{artist} - #{title}"
> ```
> 
> which is used in liquidsoap, for which I didn't find a documented semantics (and behaves differently in pcre and re).

Thanks for doing that! We have a workaround for this one in the Re switch PR: https://github.com/savonet/liquidsoap/pull/2085

glondu:
To reduce divergence with pcre, I've added support for DOTALL in the Re.Pcre module: https://github.com/ocaml/ocaml-re/pull/225.

toots:
This is great @glondu thanks for your work. I'm gonna resume the effort here soon.

toots:
Happy to report that this PR is now ready for merge! I'm gonna let it sit for a little while to have time to add more tests. I would like to make sure that we don't run into some backward compatibility issues in some untouched corners.

@glondu are you aware of some corner cases with `Re` compared to `Pcre` that we should testing? I have reported a bug when splitting using an empty regexp and added a workaround for splitting the empty string as well.

toots:
I've also noticed that splitting `"#{x}"` used to produce an initial empty string (`["", x, (maybe: "")]` which `Re` does not do (it returns `[Delim "#{x}", Match x]`. I don't think that this should be a big issue though.

PR #2086, opened 2021-12-03
Show code excerpts on errors

### LINKED ISSUE #2063: Feature Request: Show Relevant Code in Error Messages

Due to the nature of Liquidsoap's error messages referring to line x and columns x-y of a given configuration file, we're often left referring back and forth to the actual generated configuration file to see what line is causing the problem. Given that the error printout already knows what line/columns are causing the problem, it seems like it would be helpful to actually print those lines themselves into the error log instead of just directing us to them in the file. When diagnosing issues, that would save a _huge_ amount of time.

smimram:
We display the relevant portion of code when raising an error.

Should fix #2063.

### DIFF
--- a/src/lang/repr.ml
+++ b/src/lang/repr.ml
@@ -27,7 +27,6 @@ let show_record_schemes = ref true
 
 open Type
 
-let string_of_single_pos = Runtime_error.print_single_pos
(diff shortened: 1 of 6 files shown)


toots (approved):
Looks great! Do you have screenshots of some examples?

PR #2087, opened 2021-12-05
Only load external plugins when needed.

### LINKED ISSUE #2019: Allow to control which dynamic plugins get loaded
Some of them have triggered segfault in the past and they slow down start time in some cases.

smimram:
This PR makes external plugins loaded only when needed. This means all
- LADSPA
- DSSI
- LV2
- frei0r
plugins. They are loaded when some keyword is used for a variable (eg `ladspa` for LADSPA plugings. This is hackish but should bug effective. Fixes #2019.

BTW: we gain startup time
```
$ time liquidsoap

real	0m0,983s
user	0m0,879s
sys	0m0,104s
```
vs
```
$ time ./liquidsoap

real	0m0,356s
user	0m0,311s
sys	0m0,045s

```

### DIFF
--- a/src/lang/parser.mly
+++ b/src/lang/parser.mly
@@ -144,7 +144,7 @@ expr:
   | BOOL                             { mk ~pos:$loc (Ground (Bool $1)) }
   | FLOAT                            { mk ~pos:$loc (Ground (Float  $1)) }
   | STRING                           { mk ~pos:$loc (Ground (String $1)) }
-  | VAR                              { mk ~pos:$loc (Var $1) }
+  | VAR                              { External_plugins.trigger $1; mk ~pos:$loc (Var $1) }
   | varlist                          { mk_list ~pos:$loc $1 }
   | GET expr                         { mk ~pos:$loc (App (mk ~pos:$loc($1) (Invoke (mk ~pos:$loc($1) (Var "ref"), "get")), ["", $2])) }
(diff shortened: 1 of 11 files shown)


smimram:
PS: we still have 20% startup time in FFmpeg filters registration but I don't know how to easily detect the use of those.

src/lang/parser.mly
```diff
@@ -144,7 +144,7 @@ expr:
   | BOOL                             { mk ~pos:$loc (Ground (Bool $1)) }
   | FLOAT                            { mk ~pos:$loc (Ground (Float  $1)) }
   | STRING                           { mk ~pos:$loc (Ground (String $1)) }
-  | VAR                              { mk ~pos:$loc (Var $1) }
```
toots, line 147: Can this and the one below be moved to the `mk` level?

toots (commented):
It's a great idea and I actually like the implementation. Could we more the trigger closer to term evaluation so that it becomes more visible when we work on terms?

src/lang/parser.mly
smimram: Good idea, I have just implemented this!

smimram:
The FFmpeg filters are now loaded on demand too. With
```
$ time ./test.liq
Registering FFmpeg filters.

real	0m0,457s
user	0m0,427s
sys	0m0,030s
```
and without
```
$ time ./test.liq

real	0m0,265s
user	0m0,218s
sys	0m0,047s
```

PR #2174, opened 2022-01-26
Swscale segfaulting.

smimram:
The test makes swscale (presumably) segfault.


toots:
Looks like the issue was from a faulty commit in `ocaml-ffmpeg` and is fixed now.

Issue #2172, opened 2022-01-26
Replay gain is not ignored for `opus` by default.


vitoyucepi:
**Describe the bug**
If `opus` file contains replay gain metadata tag, then it will use it even if `enable_replaygain_metadata()` not used.
I checked some other file formats and there's no such bug for them.

<details>
<summary>Logs</summary>

```
liquidsoap_1  | 2022/01/26 02:03:18 >>> LOG START
liquidsoap_1  | 2022/01/26 02:03:16 [main:3] Liquidsoap 2.0.2
liquidsoap_1  | 2022/01/26 02:03:16 [main:3] Using: bytes=[distributed with OCaml 4.02 or above] pcre=7.4.6 sedlex=2.3 menhirLib=20210419 curl=0.9.1 dtools=0.4.4 duppy=0.9.2 cry=0.6.5 mm=0.7.3 xmlplaylist=0.1.5 lastfm=0.3.3 ogg=0.7.0 ogg.decoder=0.7.0 vorbis=0.8.0 vorbis.decoder=0.8.0 opus=0.2.1 opus.decoder=0.2.1 speex=0.4.0 speex.decoder=0.4.0 mad=0.5.1 flac=0.3.0 flac.ogg=0.3.0 flac.decoder=0.3.0 dynlink=[distributed with Ocaml] lame=0.3.5 shine=0.2.2 frei0r=0.1.2 fdkaac=0.3.2 theora=0.4.0 theora.decoder=0.4.0 ffmpeg=1.1.1 bjack=0.1.6 alsa=0.3.0 ao=0.2.3 samplerate=0.1.6 taglib=0.3.7 ssl=0.5.9 magic=0.7.3 camomile=1.0.2 inotify=2.3 yojson=1.7.0 faad=0.5.0 soundtouch=0.1.9 portaudio=0.2.3 pulseaudio=0.1.4 ladspa=0.2.0 dssi=0.1.3 tsdl=v0.9.8 tsdl_ttf=0 tsdl-image=0.3.2 camlimages=4.2.6 cohttp-lwt-unix=2.5.5 prometheus-app=1.1 srt.constants=0.2.1 srt.types=0.2.1 srt.stubs=0.2.1 srt.stubs.locked=0.2.1 srt=0.2.1 lo=0.2.0 gd=1.0a5
liquidsoap_1  | 2022/01/26 02:03:16 [clock:3] Using builtin (low-precision) implementation for latency control
[...]

PR #2177, opened 2022-01-27
Use R128 gain as replaygain for opus.

smimram:
Let's discuss my fix for #2172 here.

Replaygain and EBU R128 standards are different ones. However, opus only uses the second one, and you'd expect a unified way of handling both. This is done by default, but I have added an argument to `replaygain` to disable that.

### DIFF
--- a/libs/audio.liq
+++ b/libs/audio.liq
@@ -181,6 +181,23 @@ end
 # @category Source / Sound Processing
 # @param s Source to be amplified.
 def replaygain(s)
+  s = insert_metadata(s)
+  # Normalize opus gain
+  def f(m)
+    def f(m)
+      let (k, v) = m
+      if k == "r128_track_gain" then
+        v = int_of_string(v)
+        v = v + 5 # normalize to -18 dB as usual replaygain instead of -23 dB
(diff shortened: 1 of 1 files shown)


toots:
How about making this part of the regular `amplify` operator and deprecating `replaygain` altogether? There's no reason for this duplication of functions, we can just make `amplify` have `replaygain` and `ebu128` parameters, deprecated `replaygain` and simplify the API (and people's mind..)

smimram:
I find it a bit messy:
- what should be the behavior if the user passes both the arguments `replaygain=true` and a variable getter for the volume?
- we want to be able to write `s = amplify(replaygain=true, s)` which means that the volume should have a default value (say 1.), which means that in `s = amplify(s)` the typing system will not notice that we forgot to write the volume

toots:
> I find it a bit messy:
> 
> * what should be the behavior if the user passes both the arguments `replaygain=true` and a variable getter for the volume?
> * we want to be able to write `s = amplify(replaygain=true, s)` which means that the volume should have a default value (say 1.), which means that in `s = amplify(s)` the typing system will not notice that we forgot to write the volume

Can't we just catch those at runtime and error when initializing the operator?

PR #2202, opened 2022-01-31
camlimages as text renderer.

smimram:


libs/video.liq
```diff
@@ -305,11 +305,16 @@ let replaces video.add_text.best = video.add_text.gstreamer
...
+let video.add_text.best = video.add_text.camlimages
+%endif
+
 %ifdef video.add_text.ffmpeg
 let video.add_text.best = video.add_text.ffmpeg
 %endif
 
-# Display a text using the first available operator in: SDL, GStreamer, gd or native.
+# Display a text using the first available operator in: FFmpeg, camlimages, SDL,
+# GStreamer, gd or native.
```
toots, line 317: Perhaps we should make that a settings?

smimram: Well, this is just a reasonable default anyway, you can use `video.add_text.blablabla` if you want to use a specific implementation.

toots: Yeah but this makes you vulnerable to our arbitrary changes of order like what just happened here.

PR #2204, opened 2022-02-01
Remove partial application support.

smimram:


libs/http.liq
```diff
@@ -163,7 +163,10 @@ def harbor.http.static.base(serve,~content_type,~basepath,~headers,~browse,direc
   end
 
   def register(method)
-    serve(method=method,"^#{basepath}",handler(method=method))
+    def handler(~protocol,~data,~headers,uri)
+      handler(method=method, protocol=protocol, data=data, headers=headers, uri)
+    end
```
toots, line 168: Maybe a good opportunity to cleanup and generalize the use of `%argsof`?

src/lang/lang.mli
```diff
@@ -62,11 +62,10 @@ and in_value = Value.in_value =
...
   | Ref of value ref
-  | Fun of (string * string * value option) list * env * lazy_env * Term.t
-  (* A function with given arguments (argument label, argument variable,
-     default value), parameters already passed to the function, closure and
-     value. *)
-  | FFI of (string * string * value option) list * env * (env -> value)
+  | Fun of (string * string * value option) list * lazy_env * Term.t
+  (* A function with given arguments (argument label, argument variable, default
+     value), closure and value. *)
+  | FFI of (string * string * value option) list * (env -> value)
```
toots, line 68: Maybe a good opportunity to switch those to records? 🙂

src/lang/evaluation.ml
```diff
@@ -290,27 +295,24 @@ and apply f l =
     | _ :: l -> pos l
     | [] -> f.Value.pos
   in
-  (* Position of the whole application. *)
   let pos = pos l in
-  let pos_f = f.Value.pos in
-  let mk ~pos v = { pos; Value.value = v } in
+  *)
+  (* NB: the above is more precise (we get the arguments), but I don't think
+     this is worth the price: we compute it for every application... *)
```
toots, line 301: I'd either remover it or make it optionally actionable. As comment, the code is bound to become obsolete..

toots (commented):
This looks good to me. What kind of wrapper can we provide to make life easier to the few who might want to still do partial applications? `%argsof` is cool but it's a hack. If we could expand maybe we could define something useful?

toots (commented):
The PR is great, I support this simplification 100%, thanks for doing it.

If you don't mind a coule of things that would be good to have before merging:
* A description and rational for the changes in the PR description for future reference
* A note in the migration notes in the doc
* A quick review of the doc if possible to make sure our examples do not use partial applications anymore.
* Any trick you think could be added to make user's experience better when actually trying to apply some parameters in advance?

src/lang/lang.mli
smimram: I'd like to keep the diff minimal for now.

libs/http.liq
smimram: I will think of something with records, but for next time!

Issue #1066, opened 2019-12-23
Reimplement Video Effects


dhannyz:
These effects were lost during the YUV420 change.

https://github.com/savonet/liquidsoap/blob/dece14d0cec8b4eaf821daaca307954fc3fe0756/src/operators/video_effects.ml#L82-L229

PR #2207, opened 2022-02-01
Use canvas for images.

smimram:


smimram:
I think that most functionality is back without noticeable regression (in particular perfs don't seem to be affected). I'd say we can start merging. In a second time, we should add support for decoding at various sizes so that we can tile efficiently...

This of course requires the [corresponding branch](https://github.com/savonet/ocaml-mm/pull/13) in mm.

liquidsoap.opam
```diff
@@ -21,7 +21,7 @@ depends: [
   "dtools" {>= "0.4.4"}
   "duppy" {>= "0.9.1"}
   "menhir" {>= "20180703"}
-  "mm" {>= "0.7.4"}
+  "mm" {>= "0.7.5"}
```
toots, line 24:
```suggestion
  "mm" {>= "0.8.0"}
```

src/operators/add.ml
```diff
@@ -212,7 +213,7 @@ let () =
          ~kind ~renorm ~power
          (List.map2 (fun w s -> (w, s)) weights sources)
          (fun _ -> ())
-         (fun _ buf tmp -> Video.Image.add tmp buf)
+         (fun _ buf tmp -> Video.Canvas.Image.add tmp buf)
```
toots, line 216: Do you plan on changing this do use canvas?

src/operators/add.ml
```diff
@@ -299,3 +301,4 @@ let () =
          (List.map2 (fun w s -> (w, s)) weights sources)
          video_init video_loop
         :> Source.source))
+*)
```
toots, line 304: Let's delete dead code or else make it compile.

src/operators/frei0r_op.ml
```diff
@@ -134,9 +134,10 @@ class frei0r_mixer ~kind ~name bgra instance params (source : source) source2 =
             let rgb' = Content.Video.get_data rgb' in
             for i = offset to offset + length - 1 do
               (* TODO: we could try to be more efficient than converting to/from RGBA32 and swap colors... *)
-              let img = Video.get rgb i in
+              let img = Video.Canvas.render rgb i in
               let img = Image.YUV420.to_RGBA32 img in
-              let img' = Video.get rgb' i in
+              let img' = Video.Canvas.get rgb' i in
```
toots, line 139: What's the difference between `render` and `get`?

src/operators/frei0r_op.ml
```diff
@@ -146,7 +147,7 @@ class frei0r_mixer ~kind ~name bgra instance params (source : source) source2 =
               Frei0r.update2 instance t src src' dst;
               if bgra then Image.RGBA32.swap_rb img;
               let img = Image.YUV420.of_RGBA32 img in
-              Video.set rgb i img;
+              Video.Canvas.put rgb i img;
```
toots, line 150: What's the order of operations here? I'd expect the data to be an array of canvas but this API seems a bit confusing to me?

toots (commented):
This looks pretty good to me, exciting to make progress on that!

I just had a couple of questions regarding the implementation.

I'm guessing, for now, you're just aiming at updating the API without changing the underlying logic.

src/operators/add.ml
smimram: Thanks for spotting this, it was a mistake...

src/operators/add.ml
smimram: Restored the code :)

src/operators/frei0r_op.ml
smimram: `get` gives you a canvas (= a formal sum of images) whereas `render` computes an image from the canvas (it computes the result of the formal sum of images).

src/operators/frei0r_op.ml
smimram: Essentially
```
Video.Canvas.put rgb i img
```
is the same as
```
rgb.(i) <- img
```
excepting that we cast the `img` (which is a proper image) into a canvas.

src/decoder/decoder_utils.ml
```diff
@@ -84,30 +84,29 @@ let channels_converter dst =
...
-        let x = (dst_width - scl_width) / 2 in
-        let y = (dst_height - scl_height) / 2 in
-        Image.YUV420.add img2 ~x ~y img3;
-        img3))
+      let img2 = Video.Canvas.Image.make img2 in
+      let x = (dst_width - scl_width) / 2 in
+      let y = (dst_height - scl_height) / 2 in
+      img2
+      |> Video.Canvas.Image.translate x y
+      |> Video.Canvas.Image.viewport dst_width dst_height)
```
toots, line 109: Not a blocker but do you think that, eventually, we could keep the original decoded image and do the scaling on the final output render? This would make it possible to optimize and only write once each pixel on the final image..

src/operators/add.ml
```diff
@@ -283,9 +289,13 @@ let () =
               (x + ((w - w') / 2), y, w', h)))
           else (x, y, w, h)
         in
+        let tmp = Video.Canvas.Image.render tmp in
```
toots, line 292: Same q here: can we combine canvas abstractly and do the render only once at the end?

src/operators/video_effects.ml
```diff
@@ -208,14 +225,14 @@ let () =
...
     ~return_t
     ~category:`Video
     ~descr:"Blur opacity of video."
     (fun p ->
        let src = Lang.to_source (Lang.assoc "" 1 p) in
-         new effect ~kind Image.Effect.Alpha.blur src)
+         new effect ~kind src Image.YUV420.Effect.Alpha.blur)
 
 let () =
   Lang.add_operator "video.lomo"
```
toots, line 238: Is that dead code?

src/operators/video_fade.ml
```diff
@@ -168,46 +168,52 @@ let proto =
...
           let h' = ifm h t in
-          let tmp = Video.Image.create w' h' in
-          Video.Image.scale img tmp;
+          let tmp = Video.Canvas.Image.create w' h' in
+          Image.YUV420.scale img tmp;
           Video.Image.fill_alpha img 0;
           let x = (w - w') / 2 in
           let y = (h - h') / 2 in
           Video.Image.add ~x ~y tmp img
+    *)
```
toots, line 205: Can you drop a ticket for this?

src/operators/video_fade.ml
```diff
@@ -216,7 +222,7 @@ let rec transition_of_string p transition =
             "slide_up";
             "slide_down";
             "fade";
-            "grow";
+            (* "grow"; *)
```
toots, line 225: Same, can you drop a ticket to restore this one?

toots (approved):
Looks great! Only have comments about rendering later in the pipeline and filling tickets for code that will need to be restored. Thanks for the effort!

src/operators/video_effects.ml
smimram: Yes, they should be restored at some point, but it's not easy on YUV representation (#1066).

src/operators/video_fade.ml
smimram: There's already one: #1066.

src/operators/video_fade.ml
smimram: #1066, should be the place too.

src/decoder/decoder_utils.ml
smimram: Yes! The canvas type in ocaml-mm was create with this in mind (although, I don't want to do this right now...)

src/operators/video_fade.ml
smimram: Actually just fixed it.

PR #2210, opened 2022-02-02
Add an IRC client.

smimram:
We add `irc.channel` which dumps the contents of an IRC channel into a string getter. This is useful to show the log of the IRC into video with something like
```
s = blank()
s = video.add_text.native(speed=0, irc.channel(),s)
output.video(s)
```


toots:
Could this be turned into a wrapper around an external command? This pulls a lot of dependencies for something that is quite peripheral and could be done externally in the scripting library..

PR #2226, opened 2022-02-08
Change text to a generator.

smimram:
We want to have a generator which renders the text as basic operations and `add_text` implemented with `add` in the Liquidsoap language. Thanks to the recent canvas addition, this should be as efficient as the previous implementation!


smimram:
I apparently mistakenly closed the wrong bug.

PR #2227, opened 2022-02-08
Add examples sections in documentation.

### LINKED ISSUE #1221: Add usage examples to the API reference
It would be extremely helpful for newcomers if the [API reference](https://www.liquidsoap.info/doc-1.4.2/reference.html) would contain examples how to use the functions.

For newcomers (like me) the descriptions of the functions are often hard to understand. I spent several hours to understand the parameter descriptions (e.g. `(pattern : string, ((string) -> string), string) ->
string`). In fact I still only understand the simple ones. I am coming from PHP background where such nomenclatures are not used. I can imagine that people without programming knowledge will have even harder times to understand that.

Examples would help a lot.

For example it took me a while to understand how to use _string.replace_. Although it is such a simple function. I had to google it and found somebody who has used this function in a code snippet in the mailinglist.

It would be great if the reference could look like this:
> string.replace
> 
> Replace substrings in a string. Will replace all substrings matched in the pattern by the string returned by the replace function.
> 
> Type:
> (pattern : string, ((string) -> string), string) -> string
>
> Arguments:
>     pattern (of type string)
>     (unlabeled) (of type (string) -> string)
>     (unlabeled) (of type string)
>
> **Examples:**
>      str = "Hello world" 
>      str = string.replace(pattern="world", (fun(_) -> "Frank"), str) 
>      print(str) # would print "Hello Frank"

What do you think?

smimram:
Fixes #1221.

### DIFF
--- a/src/builtins/builtins_string.ml
+++ b/src/builtins/builtins_string.ml
@@ -59,6 +59,7 @@ let () =
     ~descr:
       "Retrieve a character in a string. Raises `error.not_found` if character \
        does not exist."
+    ~examples:["c = string.nth(\"abcde\", 2)\nprint(c) # should print \"c\""]
     [
       ("", Lang.string_t, None, Some "String to look into.");
       ("", Lang.int_t, None, Some "Index of the character.");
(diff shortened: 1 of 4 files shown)


mcfiredrill:
:tada: :yellow_heart:

PR #2237, opened 2022-02-11
Remove clutter from current Makefile.

smimram:
Much simplified Makefile system, which should now be more readable and modifiable...


smimram:
I know that CI is not passing right now, and I cannot understand that. It seems that we have a leftover `src/operators/video_text.ml`, which was supposed to be removed in #2226...

Makefile
```diff
@@ -55,18 +61,19 @@ endif
...
+	$(MAKE) -C doc $@
 api-doc-install:
-	$(V)echo Installing developer documentation...
+	$(V)echo "Installing developer documentation..."
 	$(V)$(INSTALL) -d $(datadir)/doc/$(DISTDIR)/api
 	$(V)for doc in $(wildcard autodoc/liquidsoap/*.html autodoc/liquidsoap/*.css) ;\
 	do $(INSTALL_DATA) $$doc $(datadir)/doc/$(DISTDIR)/api ; \
 	done
 
 # user and group are defined in Makefile.defs, written by configure.
```
toots, line 74:
```suggestion
.PHONY: install dist print-tarball-filename tarball
```

Makefile
```diff
@@ -55,18 +65,21 @@ endif
...
-
-install-local: doc-install
+.PHONY: install dist print-tarball-filename tarball
+install: doc-install
```
toots, line 81: I still think that doc install should be optional. Otherwise, everyone compiling liquidsoap will need to have `pandoc` installed to be able to install the build.

toots (approved):
This is good to go. I don't think doc should be installed by default, tho. Thanks for the hard work!

PR #2281, opened 2022-03-09
Remove positions in values.

smimram:
It is not very reasonable to propagate positions in values at runtime. This PR removes those. We only add back positions when exceptions are raised at function application. The locations are a bit less precise (we get the whole function application instead of the precise argument), but I think that this is still reasonable (and more efficient).

### DIFF
--- a/.merlin
+++ b/.merlin
@@ -1,4 +1,4 @@
 S src/**
 B src/**
 B +threads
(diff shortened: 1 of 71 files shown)


smimram:
Resulting error messages are really too badly located. Closing for now.

PR #2305, opened 2022-03-21
Add `unnan` operator to remove nan values.

smimram:


toots (commented):
The next release will have `clip` also clear out `NaN` values. Do we need a separate operator just for them?

toots:
> Ah I missed that sorry. Somehow, I don't think that clip should really clear out nan, should it? But maybe am I too picky there...

From a performance perspective, it makes sense to do all at once. More generally, I'm thinking clip is making sure all values are within the `[-1,1]` range, which includes `NaN`.

toots:
Done!

PR #2343, opened 2022-04-20
Enhance ffmpeg copy encoder/decoder

toots:
This PR adds some enhancement to the handling of keyframes when using the ffmpeg copy encoder, following issues raised in https://github.com/savonet/liquidsoap/discussions/2331#discussioncomment-2560399.

Ideally, we would like to be able to switch sources and tracks based on keyframe boundaries from encoded content to make sure that content is decodable immediately when concatenating different streams.

However, being a streaming model, liquidsoap makes switching decisions _before_ pulling content from a source and so would require some kind of look ahead mechanism to know if a keyframe is coming, which is pretty complex to implement, specially when sources are potentially shared between different operators.

This PR adds some ways to mitigate this issue at the `%ffmpeg` encoding level by adding 3 different options when switching content:
1. Ignore all keyframes issues.
2. Replay the latest known keyframe when switching to a new stream and the first packet is not a keyframe
3. Wait until at least one keyframe has been passed to start passing encoded packets from a new stream.

All 3 options apply to the cases where the `%ffmpeg` encoder has streams created using `%audio.copy` or `%video.copy`. In this situation, let's consider the case where the video content goes from one encoded stream to a new one.

* With option 1 we start passing encoded data right away. Content is immediately added but playback might get stuck until a new keyframe is passed.
* With option 2, we replay the latest  known keyframe for the new stream. This can happen if, e.g. the stream was stopped by a `switch` going to a different source and we are getting back to it. With this option, playback should be able to resume right away but there might still be decoding glitches due to missing frames between the latest keyframe and the current frame.
* With option 3, we discard any new packet until a keyframe is passed. This means that playback will be paused until it can be  resumed properly with no decoding glitches.

Option 3. is implemented globally when possible, i.e. in case of a video track with keyframes and an audio track with no keyframes, the audio track will discard packets until a video keyframe has been passed.

It is worth noting that some audio encoders may also have keyframes.

All 3 options can be set via the `%audio.copy` and `%video.copy` encoder:
* `%audio,copy(ignore_keyframe)` and `%video.copy(ignore_keyframe)`
* `%audio,copy(replay_keyframe)` and `%video.copy(replay_keyframe)`
* `%audio,copy(wait_for_keyframe)` and `%video.copy(wait_for_keyframe)`

Option 3, which seems what most user would expect, is the default so `%audio.copy` and `%video.copy` are also equivalent to, resp., `%audio,copy(wait_for_keyframe)` and `%video.copy(wait_for_keyframe)`

Lastly, this PR cleans up the logic around keyframes. While it appears that packets from format that do not have keyframes are currently marked as keyframes, it is preferable to look at wether the format actually has keyframes and. This is indicated by `AV_CODEC_PROP_INTRA_ONLY` in ffmpeg, meaning decoder only rely on the current content and no previous content to be able to decode it.

In this case, we should ignore all keyframe flags from encoded packets and proceed accordingly, which this PR also implements.

A simple test script is:
```ruby
log.level.set(4)

s1 = single("/tmp/s1.mp4")
s2 = single("/tmp/s2.mp4")

s = switch(track_sensitive=false, [
  ({file.exists("/tmp/s1")}, s1),
  ({true}, s2)
])

# Same as: enc = %ffmpeg(format="flv", %audio.copy, %video.copy(wait_for_keyframe))`
enc = %ffmpeg(format="flv", %audio.copy, %video.copy)
# Or:
# enc = %ffmpeg(format="flv", %audio.copy, %video.copy(ignore_keyframe))`
# enc = %ffmpeg(format="flv", %audio.copy, %video.copy(replay_keyframe))`

output.file(enc, "/tmp/out.flv", s)
```

### DIFF
--- a/liquidsoap.opam
+++ b/liquidsoap.opam
@@ -86,8 +86,8 @@ conflicts: [
   "dssi" {< "0.1.3"}
   "faad" {< "0.5.0"}
   "fdkaac" {< "0.3.1"}
-  "ffmpeg" {< "1.1.2"}
-  "ffmpeg-avutil" {< "1.1.2"}
+  "ffmpeg" {< "1.1.4"}
+  "ffmpeg-avutil" {< "1.1.4"}
   "flac" {< "0.3.0"}
   "frei0r" {< "0.1.0"}
   "gstreamer" {< "0.3.1"}
--- a/src/decoder/ffmpeg_copy_decoder.ml
+++ b/src/decoder/ffmpeg_copy_decoder.ml
@@ -69,10 +69,7 @@ let mk_audio_decoder ~stream_idx ~format container =
        Ffmpeg_copy_content.(Audio.lift_params (Some params)));
   let stream_time_base = Av.get_time_base stream in
   let lift_data = Ffmpeg_copy_content.Audio.lift_data in
-  let latest_keyframe =
-    Ffmpeg_copy_content.latest_keyframe
-      Avcodec.Audio.(descriptor (get_params_id params))
-  in
+  let latest_keyframe = Ffmpeg_copy_content.latest_keyframe params in
   ( idx,
     stream,
     mk_decoder ~stream_idx ~lift_data ~stream_time_base ~put_data:G.put_audio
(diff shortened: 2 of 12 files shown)


TheRealMattLear:
This is a huge achievement, great work @toots. I'm struggling to install the branch, tried with both ocaml 4.12.0 and 4.14.0 but i'm getting ```Your request can't be satisfied:   - liquidsoap >= 2.0.4 is in conflict with ffmpeg < 1.1.4```

PR #2345, opened 2022-04-23
Liquidsoap v2.0.4

toots:


CHANGES.md
```diff
@@ -1,4 +1,4 @@
-2.0.4 (04-28-2022)
+2.0.4 (23-04-2022)
```
vitoyucepi, line 1: I think it will be better if the date is in ISO8601 format.

toots: It would be better indeed. We've been keeping historical date formats in the changelog but it's time to change them.

PR #2385, opened 2022-05-06
Compile Liquidsoap to javascript in order to make an online version

smimram:


rrolla:
This looks interesting 👁️

toots:
This is in main now and can be checked out here: https://www.liquidsoap.info/try/. Still some work to do to polish it but it's there!

PR #2391, opened 2022-05-07
Enhance ffmpeg documentation

toots:


rrolla (commented):
@toots @smimram Question where is a link in https://www.liquidsoap.info/doc-2.0.4/ to this Documentation index page? https://www.liquidsoap.info/doc-2.0.4/documentation.html I can open it when just manually enter in url.

PR #2403, opened 2022-05-16
Restore mix.

### LINKED ISSUE #2401: Bring back `mix` operator
From slack:
> Hi :wave: I'm currently looking into updating our Liquidsoap setup from 1.4 to 2.0. We're using the mix function ([link](https://www.liquidsoap.info/doc-1.4.4/reference.html#mix)) to have a mixer between several inputs to be controlled via telnet. It seems this is gone in 2.0? Is there a replacement? How are people dynamically switching between several inputs at runtime in 2.0?

smimram:
Fixes #2401.


smimram:
Superseeded by #2404.

PR #2409, opened 2022-05-20
Switch to dune!

toots:


doc/dune.inc
```diff
@@ -0,0 +1,1039 @@
+(rule
```
smimram, line 1: This one looks uselessy very heavy. Couldn't we simply use a makefile with a rule `dune exec ...src/bin/liquidsoap.exe ...` which will force the compilation of liquidsoap anyway? And it's even worse for the gneration of files with pandoc below. `make` is still useful sometimes :)

toots: This file is auto-generated via `dune build @gendune --auto-promote`

toots:
`dune` support still needs some tweaking but tests are passing and we've confirmed the build mechanism to be able to produce valid system-wide binaries for at least the debian packages. This is good to go with follow-up changes to be coming soon. Exciting!

PR #2419, opened 2022-05-27
Documentation - update Beets examples to Liquidsoap 2.x

martinkirch:
I removed the playlist generator example because it's a weird approach, whereas `request.dynamic` with replaygain does wonders :ok_hand: 

not sure I'm targeting the right branch ; this is valid for Liquidsoap 2.x


toots:
Thanks for your contribution! Merging to `rolling-release-v2.0.x` to come out with `2.0.6` and cherry-picking to `rolling-release-v2.1.x` and `main`.

PR #2420, opened 2022-05-27
log output if process.read[.lines] does not success

martinkirch:
can be disabled with a new parameter, `log_errors`.

I propose this change because trivial errors (say, a typo in the command name) are completely silenced by `process.read`, which can cascade to weird behaviours (like a requests loop if `process.read` is called in `request.dynamic`). This will hopefully ease debugging.


toots:
This is an excellent suggestion, thank you so much for contributing it and sorry it took me so long to get back to you.

PR #2421, opened 2022-05-27
Update index.md

negotiamini:
Made some grammatical changes for easier readbility.


toots:
Thank you so much! Don't hesitate to send more like this!

negotiamini:
Thanks will do!

PR #2463, opened 2022-06-18
Make request pool a weak hash

toots:
I'm not sure why requests are structured the way they are. This is some of the oldest part of the code. In particular, it's not clear why we need a grace period and, hence, an explicit destroy.

This PR replaces the requests pool hash with a weak hash, leaving to the GC the task of destroying them without having to make an explicit call. The call to `destroy` is left and can still be done to speed up request cleanup but it is made idempotent so that it can be called by the user ahead of the GC.

### DIFF
--- a/.github/scripts/build-win32.sh
+++ b/.github/scripts/build-win32.sh
@@ -37,7 +37,7 @@ export CC=""
 eval `opam config env`
 opam repository set-url default https://github.com/ocaml/opam-repository.git
 cd /home/opam/opam-cross-windows/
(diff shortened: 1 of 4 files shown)


smimram:
Thanks this is much cleaner! I think that the idea behind the grace period is that we could reuse requests and avoid downloading twice the same file if we know that it will be used frequently (e.g. a jingle). But this clearly complexifies things and relying on the GC to automatically handle that is a much cleaner strategy (and prevents requests leaks due to forgetting `request.destroy`...)

PR #2470, opened 2022-06-21
Preserve tracks in soundtouch.

### LINKED ISSUE #2467: Track markers on soundtouch and/or stretch
If I use soundtouch or stretch before a cross function I get no crossfade and no response to on_track method (non dependent on cross being present).

**Describe the solution you'd like**
I'd like to be able to use the above situation, to have cross see the tracks and trigger fades on a stretched / soundtouch source

**Describe alternatives you've considered**
#2466 

I could also achieve the result I'm after with being able to extend an 'effect' from inside a source transition (see #2466), but I'm thinking this is not in any way the intended behavior , so I'll mention it no more.

smimram:
Can be tested with
```
s = chop(every=1., sine())
s = soundtouch(s)
s = source.on_track(s, fun(_) -> print("***** new track #{time.up()}"))
output(s)
```
Fixes #2467.

### DIFF
--- a/src/core/clock.ml
+++ b/src/core/clock.ml
@@ -349,14 +349,17 @@ let rec active_params c =
     | _ when Atomic.get global_stop -> raise Has_stopped
     | _ -> raise Invalid_state
 
-and _tick ~clock x =
+and _activate_pending_sources ~clock x =
   Queue.flush clock.pending_activations (fun s ->
       check_stopped ();
       s#wake_up;
       match s#source_type with
         | `Active _ -> WeakQueue.push x.active_sources s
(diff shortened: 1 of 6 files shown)


toots:
Unfortunately, this breaks the tests. The streaming API is pretty fragile with frame breaks vs. end of tracks. I'd rather fix this the right way, i.e. finally switch to immutable content array and make breaks first-class track marks. This branch is the first step: https://github.com/savonet/liquidsoap/pull/2469

toots:
Just push a redo of this one with the test included. Thanks for starting the work here. Turned out that the operator was broken during the recent clock changes but track boundaries was fixed by the switch to immutable content!

PR #2571, opened 2022-08-22
Generic content type

toots:
This PR switches internal types used to represent source and format content from the current ad-hoc content triple to a generic record, making it possible to represent any type of content in streams. This will be used to support arbitrary stream content (multiple audio tracks, subtitle track, track marks etc).

Technically, the PR is rather simple now that most of the fixes and tweaks required for it have been extracted. It simply:
* Replaces `Frame_type` returned types from a constructed type triplet to a record
* Updates the different APIs
* Updates operator's declared types

Notable quirks and notes:
* Sources may now be fully typechecked as `source('a)`, for instance `output.dummy(blank())`. In this case, content type is instantiated using the default audio/video (defaults to `{audio=pcm(stereo)}`).
* Some implicit video stuff have to be explicit now, typically for `fade.in` which is changed to audio-only with this PR. 
* I have noticed some imperfect type errors report like `value is of type source(...) but should of type source(...)`. This is most likely a current limitation on record-related type errors and could/should be addressed in a separate PR.
* Most of the code change is strictly about types with a couple tweaks also required for decoders.
* The typing for `drop_*` operator is imperfect:
```
% liquidsoap -h drop_audio
Drop all audio content of a stream.

Type: (?id : string?, source(audio='a, 'b)) -> source('b)
```
This could be wrong if `'b` is later found to also have `audio` content. We have discussed using a constraint for that. I have also some more specific plans for source track muxing/demuxing that would completely remove the need for these operators
* `None` content is totally removed except in the `Generator` module where it is defined and used to avoid changing the current generator implementation and logic. The plan is to completely rewrite generators using immutable content once that lands.


src/core/lang.mli
```diff
@@ -220,10 +197,8 @@ val error_t : t
 val source_t : ?methods:bool -> t -> t
 val of_source_t : t -> t
 val format_t : t -> t
-val kind_t : Frame.kind -> t
-val kind_none_t : t
-val frame_t : t Frame.Fields.t -> t
-val of_frame_t : t -> t Frame.Fields.t
+val frame_t : t -> t Frame.Fields.t -> t
```
smimram, line 200: Please add detailed comment for this one. In particular, explain that the first `t` should either be a variable or `unit_t`. It might be more natural here and in related places to have the first argument a `t` option so that we can use `None` instead of `unit_t` for closed records?

src/core/lang.mli
```diff
@@ -220,10 +197,8 @@ val error_t : t
...
+val internal_t : unit -> t
```
smimram, line 201: Add a comment here. Does this correspond to the default frame type with internal content types?

src/core/stream/frame.mli
```diff
@@ -26,42 +26,45 @@
...
+(* Default fields *)
 val audio_field : field
 val video_field : field
 val midi_field : field
+val audio_field_n : int -> field
+val video_field_n : int -> field
+
+(* Find fields *)
 
 (* The following raise [Not_found] when the field does not exist. *)
```
smimram, line 44: Outdated comment:
```suggestion
```
As a general remark, it's nice to move away from `Not_found` exceptions and use options instead!

tests/language/typing.liq
```diff
@@ -152,10 +152,10 @@ def f() =
   end
   ignore(sum_eq)
 
-  (noise():source(audio=pcm,video=canvas,midi=none))
```
smimram, line 155: Is `midi=none` still supported? I think we should for best backward compat...

src/libs/fades.liq
```diff
@@ -68,15 +68,6 @@ def mkfade(~type="lin",~start=0.,~stop=1.,~duration=3.,~on_done={()},s) =
...
 
-# Scale source during fading.
-# @category Source / Fade
-# @flag hidden
-def fade.scale(~id="fade.scale", x, s)
-  s = amplify(id=id, override=null(), x, s)
-  s = video.opacity(id=id, x, s)
-  s
-end
-
```
smimram, line 79: Please leave `fade.scale` (but simply comment the line about `video.opacity`) and continue using it below. We ought to find a way to restore the support for video (and I think that I have ideas on this) and this will make things easier.

smimram (dismissed):
Thanks for this work this will bring great improvements and modularity! I have only read the main files, but it seems decent. You can merge once the minor remarks have been addressed (mostly requests for comments ;))

src/core/lang.mli
toots: I will add a comment. I decided to leave it explicit to make the code more readable. When optional, you always have to wonder what is the default, universal or unit? Or something else (internal constraint for instance). Ideally, I'd like see an operator type and be able to figure out the whole deal:
```ocaml
let return_t =
  Lang.frame_t
    (Lang.univ_t ())
    (Frame.fields ~audio:(Format_type.audio ()) ())
in
```
Vs.:
```ocaml
let return_t =
  Lang.frame_t
    Lang.unit_t
    (Frame.fields ~audio:(Format_type.audio ()) ())
in
```
This was a common mistake when proof reading inputs such as `input.alsa` type (where it should be `unit_t`)

src/core/stream/frame.mli
toots: Agreed!

tests/language/typing.liq
toots: In this specific case I believe that it is implicit:
```
(s:source)
```
Is an open type `source('a)`

But:
```
(s:source(audio=pcm, video=canvas)
```
is a closed type: `source(audio=pcm, video=canvas)`

I believe that this solves most of the use case in the most natural way. In the future, we could consider:
* Adding a `without <field>` to records to explicitly prevent specific fields. This would also be useful for muxers/demuxers and it's pretty likely we'll get to it.
* Refine type annotation to allow open type with specific fields, i.e. `source(audio=pcm, video=canvas, 'a)`. I think we should get to this one when we need it.

src/libs/fades.liq
toots: I will do it for sake of brevity of this PR changes but I've never been a fan of doing implicit stuff with video. Most users would also be surprised by this I think. Things done on video should be explicit and be noted in the type as well.

PR #2580, opened 2022-08-27
Remove kind in favor of run-time type check

toots:
This PR removes the `kind` layer that was used to check sources content at runtime and replaces it with a type annotation.

In essence, this brings back runtime types but only for sources. The benefit of this are:
* Single unique implementation of content type check
* No confusion between the two layers and less potential bugs
* Less paperwork to do when instantiating operators

This will also make updating the types to support variable frame content much, much easier!

The new code does:
* Initiate all sources with a `#frame_type` type, set as: `audio='a, video='b, midi='c`
* Runs a typecheck to track the source type as assigned through the typechecker:
```ocaml
let eval_check ~env:_ ~tm v =
  let open Liquidsoap_lang.Term in
  if Lang_source.V.is_value v then
    Typing.(tm.t <: Lang.source_t (Lang_source.V.of_value v)#frame_type)
```
* Runs a typecheck against the generalized operator return type:
```ocaml
      let v = source (src :> Source.source) in
      let generalized, return_t = Typing.generalize ~level:0 return_t in
      let return_t = Typing.instantiate ~level:0 ~generalized return_t in
      Typing.((V.of_value v)#frame_type <: return_t);
```
* Operators using `Kind.unify` are now using `Typing.(<:)`
* Source and output classes do not need to declare their runtime kind anymore as this is handled implicitly and systematically!

As a side note, it looks like the check we had to make sure runtime kind and typecheck matched was never working as it was invoked too early and a the wrong place. Typically, it would be invoked at this line:
```
s = blank()
```
where we did not know yet if the source would be used in a `mono` or `stereo` encoder, so its type is simply `source(audio='a, ...)`.


smimram:
The diff is quite heavy but the resulting code is much simpler and cleaner (in particular, no more kind type vs kind format...). Go for it!! :)

toots:
I've done some tests with muxers and ffmpeg filters, the trickiest part and it seemed to work fine. We also have extensive tests for encoding/decoding/filters and it did catch issues like not passing the right number of channels from the encoders. I think that this is safe to merge. Thanks for the review!

PR #2603, opened 2022-09-07
Use atomic for ref.

toots:
This PR swaps our implementation of `ref` to use the new `Atomic` API which helps remove concurrency issues when running liquidsoap code inside a duppy task via `thread.run`.

It also removes `thread.mutexify` which becomes almost outdated. Technically, reads might have concurrency issues but this should be rare and will also be fixed with OCaml 5. It is not providing a backward compatible deprecation as this function was pretty arcane and the replacement solution is simple, just remove it 😅

### DIFF
--- a/src/core/builtins/builtins_thread.ml
+++ b/src/core/builtins/builtins_thread.ml
@@ -126,31 +126,3 @@ let () =
       in
       Stack.push handler error_handlers;
       Lang.unit)
-
-let () =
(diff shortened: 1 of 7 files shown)


smimram (approved):
Go for it!

toots:
Thanks. The failing test is `time.liq` which seems unrelated. I'm merging this and will follow-up in `main`.

PR #2613, opened 2022-09-12
Add nullable default to conversion functions

toots:
This PR adds nullable default to `{int,float,bool}_of_string` conversion functions, raise an exception if conversion fails and no default is given.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -49,6 +49,8 @@ Changed:
 - Added `--list-functions-json` option.
 - Removed built-in use of `strftime` conversions in output filenames, replaced
   by an explicit call to `time.string` (#2593)
+- Added nullable default to `{int,float,bool}_of_string` conversion functions, raise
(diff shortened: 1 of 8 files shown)


smimram:
I am not so sure we should deprecate the `string.*` functions: the new syntax is nice but a bit puzzling for newcomers. Couldn't we officially support both versions?

toots:
> I am not so sure we should deprecate the `string.*` functions: the new syntax is nice but a bit puzzling for newcomers. Couldn't we officially support both versions?

Ok. In this case, should be make the `string.*` functions operate on `string` only? I'm trying to avoid having implicit regexp and this can be a surprising side-effect.

toots:
PR updated to only add nullable default to `{int,float,bool}_of_string` conversion functions, raise an exception if conversion fails and no default is given. Should be straight forward now, will merge once it passes CI.

PR #2614, opened 2022-09-12
Remove let json.stringify.

toots:
This PR removes `let string.stringify` from the language. It was not a good solution as the type checker was not able to generalize properly the construct inside a function. The only advanced use of it was to use type annotations to convert e.g. a `[string * string]` to a JSON object but you can do that explicitly via the abstract `json()` object API.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -30,7 +30,7 @@ Changed:
 - Add `sorted` option to `file.ls`.
 - Add `buffer_length` method to `input.external.rawaudio` and
   `input.external.wav` (#2612).
-- Removed confusing `let json.stringify` in favor of `json.stringify()`
+- Removed confusing `let json.stringify` in favor of `json.stringify()`.
 
 Fixed:
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -30,6 +30,7 @@ Changed:
 - Add `sorted` option to `file.ls`.
 - Add `buffer_length` method to `input.external.rawaudio` and
   `input.external.wav` (#2612).
+- Removed confusing `let json.stringify` in favor of `json.stringify()`
 
 Fixed:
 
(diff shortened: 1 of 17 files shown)


CHANGES.md
```diff
@@ -28,6 +28,7 @@ Changed:
   new tracks. Use `persist_overrides` to revert to previous behavior
   (`persist_override` for `cross`/`crossfade`) (#2488).
 - Add `sorted` option to `file.ls`.
+- Removed confusing `let json.stringify` in favor of `json.stringify()`
```
smimram, line 31:
```suggestion
- Removed confusing `let json.stringify` in favor of `json.stringify()`.
```

doc/content/json.md
```diff
@@ -260,19 +260,8 @@ If a `json5` variable is in scope, you can also simply use `let json.parse[json5
...
-```liquidsoap
-let json.stringify s = (x : {foo: int})
-```
-
-The `json.stringify` syntax supports the following arguments:
-
-* `json5`: allow `json5` representations, in particular infinite and `NaN` floating point numbers. Default: `false`
-* `compact`: output a compact value instead of a human-readable value. Default: `false`.
+Exporting JSON values can be done using the `json.stringify` function. Please note that not
+all values are exportable as JSON, in which case the function will raise an exception.
```
smimram, line 264: A small example please :)

something like
```
r = {artist = "Bla", title = "Blo"}
print(json.stringify(r))
```

src/lang/builtins_json.ml
```diff
@@ -372,10 +305,9 @@ let () =
       Lang.meth (JsonValue.to_value v) meth)
 
 let () =
-  Lang.add_builtin "_internal_json_renderer_" ~category:`String ~flags:[`Hidden]
-    ~descr:"Internal JSON renderer."
+  Lang.add_builtin "json.stringify" ~category:`String
+    ~descr:"Convert a value to JSON when possible."
```
smimram, line 309: The "when possible" is a bit scary: explain " (values like functions cannot be converted to JSON)".

smimram (approved):
Apart from small suggested improvements, this is good to go.

PR #2618, opened 2022-09-14
Relax video compatibility.

toots:


TheRealMattLear:
Testing confirms:

- Transition across single() and input.rtmp() between different resolutions of the same codec worked PERFECT! :tada:
- Transition with different audio bitrate also works perfectly :tada:
- Transition between a stream with different audio channels results in incompatible format error.
- Transition between audio samplerates results in incompatible format. 

Would love to test the last two 🙂

TheRealMattLear:
Latest Build (d5401d00) seems to lockup when transitioning back from live rtmp to vod with same audio profile but differing video resolution (rtmp provided by nginx-rtmp). Original build (3c08a26d) is not experiencing this problem. CTRL+C is unable to exit the process and kill -9 is required to terminate.

I'm seeing this line spammed repetitively once the live rtmp connects:
```
[NULL @ 0x7f53a8217380] ct_type:0 pic_struct:0
```

Then the following on disconnection right before the hang up:
```
[rtmp @ 0x7f53a800a5c0] Deleting stream...
```

PR #2629, opened 2022-09-18
Native replaygain computation

### LINKED ISSUE #1002: In house replaygain extraction
Now that ocaml-ffmpeg is more mature, we should add support for replaygain filter and use the library to extract replaygain instead of relying on the external script `extract-replaygain`.

smimram:
Should fix #1002.

### DIFF
--- a/src/core/builtins/builtins_source.ml
+++ b/src/core/builtins/builtins_source.ml
@@ -220,3 +220,29 @@ let () =
       log#info "Source dumped.";
       fo#leave s;
       Lang.unit)
+
+let () =
+  let log = Log.make ["source"; "drop"] in
+  Lang.add_builtin "source.drop" ~category:`Liquidsoap
+    ~descr:"Animate the source as fast as possible, dropping its output."
+    ~flags:[`Experimental]
+    [("", Lang.source_t (Lang.univ_t ()), None, Some "Source to encode")]
+    Lang.unit_t
+    (fun p ->
+      let s = List.assoc "" p |> Lang.to_source in
+      let o =
+        new Output.dummy
+          ~infallible:false
+          ~on_start:(fun () -> ())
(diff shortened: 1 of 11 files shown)


src/core/builtins/builtins_source.ml
```diff
@@ -220,3 +220,29 @@ let () =
...
+      in
+      o#get_ready [s];
+      log#info "Start dropping source.";
+      while s#is_ready do
+        o#output;
+        o#after_output
+      done;
+      log#info "Source dropped.";
+      o#leave s;
+      Lang.unit)
```
toots, line 248: Can you just do a `output.dummy` with an unsynced clock?

toots (commented):
Thanks for working on this! 

Have you considered implementing an operator working at the request level? That would feel more natural there I think..

src/core/builtins/builtins_source.ml
smimram: nope: we want to perform some action after the whole source has been played, which is not  easy to achieve with unsynced clocks...

src/libs/protocols.liq
```diff
@@ -346,14 +324,11 @@ def ffmpeg_protocol(~rlog,~maxtime,arg) =
     list.fold(f,[],lines)
   end
 
-  def replaygain_filter(file) =
+  def replaygain_filter(fname) =
     if replaygain then
-      # The extraction program
-      extract_replaygain = settings.protocol.replaygain.path()
-      delay = maxtime - time()
-      ret = exec_replaygain(delay=delay,extract_replaygain=extract_replaygain,file)
```
toots, line 354: Does `exec_replaygain` also need to be removed?

smimram: `extract_replaygain` is removed yes!

PR #2640, opened 2022-09-22
Add http.post.file.

### LINKED ISSUE #2617: Support multipart/form-data with http.{post.put}
I'll like to avoid use CURL to post form-data audio to external API, and replace them with native `http.post`
```
uploader = process.read.lines("curl -i -X POST https://url.com/ -F \"file=@#{fname}\"")
```
With something more native like:
```
res = http.post(data="file=@file.read(fname)",file_url(path.basename(fname))) 
```

smimram:
Fixes #2617.

### DIFF
--- a/src/core/builtins/builtins_http.ml
+++ b/src/core/builtins/builtins_http.ml
@@ -105,7 +105,10 @@ let add_http_request ~stream_body ~descr ~request name =
       let headers =
         List.map (fun (x, y) -> (Lang.to_string x, Lang.to_string y)) headers
       in
-      let timeout = Lang.to_int (List.assoc "timeout_ms" p) in
+      let timeout =
+        List.assoc "timeout_ms" p |> Lang.to_option |> Option.map Lang.to_int
+        |> Option.value ~default:10000
+      in
       let http_version =
         Option.map Lang.to_string (Lang.to_option (List.assoc "http_version" p))
       in
--- a/src/core/builtins/builtins_http.ml
+++ b/src/core/builtins/builtins_http.ml
@@ -106,8 +106,7 @@ let add_http_request ~stream_body ~descr ~request name =
         List.map (fun (x, y) -> (Lang.to_string x, Lang.to_string y)) headers
       in
       let timeout =
-        List.assoc "timeout_ms" p |> Lang.to_option |> Option.map Lang.to_int
-        |> Option.value ~default:10000
+        Lang.to_valued_option Lang.to_int (List.assoc "timeout_ms" p)
       in
       let http_version =
         Option.map Lang.to_string (Lang.to_option (List.assoc "http_version" p))
--- a/src/libs/http.liq
+++ b/src/libs/http.liq
@@ -395,7 +395,7 @@ def http.post.file(~name="file", ~content_type=null(), ~headers=[], ~boundary=nu
   end
 
   if null.defined(file) and null.defined(contents) then
-    error.raise(error.http, "Only one of: `content` or `file` must be defined!")
+    error.raise(error.http, "Only one of: `contents` or `file` must be defined!")
   end
 
   # Massage parameters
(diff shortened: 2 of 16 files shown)


src/core/builtins/builtins_http.ml
```diff
@@ -105,7 +105,10 @@ let add_http_request ~stream_body ~descr ~request name =
       let headers =
         List.map (fun (x, y) -> (Lang.to_string x, Lang.to_string y)) headers
       in
-      let timeout = Lang.to_int (List.assoc "timeout_ms" p) in
+      let timeout =
+        List.assoc "timeout_ms" p |> Lang.to_option |> Option.map Lang.to_int
+        |> Option.value ~default:10000
```
toots, line 110:
```suggestion
        Option.value ~default:1000 (Lang.to_valued_option Lang.to_int (List.assoc "timeout_ms" p))
```

toots: Actually this is not good. We want `null()` for timeout to mean "no timeout set". Will fix.

toots:
Added to this PR:

### `http.port.multipart_form_data`

A utility to generate `multipart/form` data. This could be re-used for posting forms data but was mostly intended to make the resulting data testable

### `file.mime` cleanup

The `file.mime` API was confusing, it was changed to follow the way we do it with `file.metadata`:
  * `file.mime.libmagic` is the libmagic implementation, present when compiled
  * `file.mime.cli` calls the `file` CLI
  * `file.mime` combines the two, using `libmagic` if present and `file` otherwise.
  * `null` is returned when no result can be found. Worth noting: there shouldn't be any benefit in trying `file` after `libmagic` since the two use the same library.

### `string.getter` convenience functions:
* `string.getter.flush`:  Flush all values from a string getter and returns the concatenated result. If the getter is constant, return the constant string. Otherwise, call the getter repeatedly until it returns an empty string and return the concatenated result
* `string.getter.concat`: Combine a list of string getters `[g1, ..., ]` and return a single getter `g` such that: `string.getter.flush(separator=s, g) = string.concat(separator=s, list.filter(fun (s) -> s != "", [string.getter.flush(g1), ...]))`

src/libs/http.liq
```diff
@@ -327,3 +327,97 @@ def replaces harbor.http.static(~transport=http.transport.unix,~port=8000,~path=
...
+# @param ~timeout Timeout in ms.
+# @param ~redirect Follow reidrections.
+# @param url URL to post to.
+def http.post.file(~name="file", ~content_type=null(), ~headers=[], ~boundary=null(), ~filename=null(), ~file=null(), ~contents=null(), ~timeout=null(), ~redirect=true, url)
+  if not null.defined(filename) and not null.defined(file) then
+    error.raise(error.http, "At least one of: `file` or `filename` must be defined!")
+  end
+
+  if null.defined(file) and null.defined(contents) then
+    error.raise(error.http, "Only one of: `content` or `file` must be defined!")
```
toots, line 398:
```suggestion
    error.raise(error.http, "Only one of: `contents` or `file` must be defined!")
```

src/libs/string.liq
```diff
@@ -204,3 +204,82 @@ def string.data_uri.decode(s) =
     null()
   end
 end
+
+let string.getter = ()
+
+# Flush all values from a string getter and returns
```
toots, line 210:
```suggestion
# Flush all values from a string getter and return
```

src/libs/string.liq
```diff
@@ -204,3 +204,82 @@ def string.data_uri.decode(s) =
...
+      chunk = getter.get(gen)
+      if chunk == "" then string.concat(separator=separator, data) else
+        f([...data, chunk])
+      end
+    end
+    f([])
+  end
+end
+
+# Combine a list of string getters `[g1, ..., ]`
```
toots, line 230:
```suggestion
# Combine a list of string getters `[g1, ...]`
```

toots (approved):

PR #2643, opened 2022-09-25
Various `http.post` fixes

toots:
* Fix `http.{post,put}` data type.
* Rename `http.post.multipart_form_data` into `http.multipart_form_data`
* Add `http.put.file`
* Test, fix curl headers logic.


rrolla:
@toots this fix can be related to my question? https://github.com/savonet/liquidsoap/discussions/2068

PR #2457, opened 2022-06-16
Format code and fix spelling errors

jooola:
To be the one who will annoy on each commit, I propose to add a pre-commit config to format the code and check for spelling errors. This follows #1654.

There is also a CI job that enforce that.

The PR is quiet big, hope you don't mind.

PR #2646, opened 2022-09-27
Add pre-commit setup

jooola:
Since there is no way to easily review #2457, I will break this huge PR in multiple smaller ones, by adding tools incrementally.

This is the first part: add a basic pre-commit, and clean the files (mostly fixing end-of-file-lines, mixed-line-ending, trailing-whitespace)


tests/core/test.srt
```diff
@@ -18,4 +18,3 @@ I guess I was wrong.
 5
 00:03:20,476 --> 00:03:22,671
 There was no danger at all.
-
```
toots, line 21: Looks like this is causing a test failure. Might need to adjust either here or in the test.

toots (commented):
This is pretty awesome! Could you also add `dune build @fmt --auto-promote` to the `pre-commit` hooks? This automatically re-formats the ocaml and dune files.

I love precommit so I'm down to merge this anytime as long as this doesn't slow down the `git commit` too much. I'll leave some time for @smimram to review as well.

jooola:
> This is pretty awesome! Could you also add dune build @fmt --auto-promote to the pre-commit hooks? This automatically re-formats the ocaml and dune files.

I'll add extra features in future PR, otherwise we will end up with  #2457 which is impossible to review.

> I love precommit so I'm down to merge this anytime as long as this doesn't slow down the git commit too much. I'll leave some time for @smimram to review as well.

Pre-commit doesn't add much overhead, as it only checks files that are about to be committed, not the whole repository. That said, if we add a hook that does requires some time, we can only tweak that tools to go faster.

tests/core/test.srt
jooola: I am not sure I understand how to fix this:
```
2022-09-27T14:53:27.4605815Z File "tests/core/dune.inc", line 329, characters 0-145:
2022-09-27T14:53:27.4606506Z 329 | (rule
2022-09-27T14:53:27.4606873Z 330 |  (alias runtest)
2022-09-27T14:53:27.4666046Z 331 |  (package liquidsoap)
2022-09-27T14:53:27.4666332Z 332 |  (deps
2022-09-27T14:53:27.4666589Z 333 |   (:test_srt ./test.srt)
2022-09-27T14:53:27.4666880Z 334 |   (:parsesrt parsesrt.exe))
2022-09-27T14:53:27.4667187Z 335 |  (action (run %{parsesrt} %{test_srt})))
2022-09-27T14:53:27.4667537Z (cd _build/default/tests/core && ./parsesrt.exe test.srt)
2022-09-27T14:53:27.4667892Z Fatal error: exception Stdlib.Queue.Empty
2022-09-27T14:53:27.4668561Z Raised at Stdlib__Queue.peek in file "queue.ml", line 60, characters 11-22
2022-09-27T14:53:27.4669189Z Called from Srt_parser.parse in file "src/core/tools/srt_parser.ml", line 86, characters 12-24
2022-09-27T14:53:27.4669798Z Called from Dune__exe__Parsesrt in file "tests/core/parsesrt.ml", line 6, characters 10-33
```

From what I see above I understand that the problem comes from the parser not from a snapshot not being equal ? Shouldn't the srt_parser be able to work around this ?

jooola: It should probably check for 1 blank line and maybe the EOF ?

toots: Let me see if I can fix this one.

toots: Should be fixed in `main` now.

toots:
We had some issues in the CI that should be fixed now if you want to rebase again `main`. I'm happy to merge this as soon as it passes the CI.

jooola:
Done, I re based onto the main branch.

jooola:
CI isn't fork friendly yet, I'll open another PR for that

PR #2652, opened 2022-09-28
Add support for completion in emacs.

smimram:


scripts/dune
```diff
@@ -1,6 +1,31 @@
...
+   (progn
+    (run ./gen_emacs_completion.exe)
+    (cat ./liquidsoap-completion.el.in)))))
+
 (install
  (section share_root)
  (package liquidsoap)
  (files
-  (liquidsoap-mode.el as emacs/site-lisp/liquidsoap-mode.el)
   (bash-completion as bash_completion/completions/liquidsoap)))
```
toots, line 24: Shouldn't this be moved to `liquidsoap_mode` as well?

toots (approved):

PR #2653, opened 2022-09-28
Don't upload CI artifacts on forks

jooola:


toots:
Would you mind factoring out some kind of higher-level variable like `is_fork` so we can re-use it accros actions? This is bound to happen more I imagine.

jooola:
This is more a quick fix, I was hoping to have less false positives on my side. The refactor will happen when I start improve the CI.

toots:
Cool with me.

PR #2655, opened 2022-09-29
Add codespell pre-commit hook

jooola:
This is the second part of #2457: Add codespell pre-commit hook

### DIFF
--- a/.codespellignore
+++ b/.codespellignore
@@ -1,0 +1,13 @@
+als
+ans
+fo
+hda
+mot
+nd
+ot
+ro
+ser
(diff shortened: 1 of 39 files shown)


toots (approved):

jooola:
I would prefer doing one step at the time and not mix everything together.

Feel free to add the ocamlformat hook, I was intended to add it at the end but please don't mix all the commits or I feel changes are gonna be lost in the middle.

jooola:
I already had some work for ocamlformat locally btw...

PR #2663, opened 2022-10-03
Add pre-commit install instructions

jooola:


.pre-commit-config.yaml
```diff
@@ -20,7 +23,7 @@ repos:
       - id: trailing-whitespace
 
   - repo: https://github.com/savonet/pre-commit-ocamlformat
-    rev: v1
```
toots, line 23: Any reason to revert this? Commit tags are more user-friendly.

.pre-commit-config.yaml
```diff
@@ -1,6 +1,9 @@
 ---
 # See https://pre-commit.com for more information
 # See https://pre-commit.com/hooks.html for more hooks
+#
+# Run the following command to set up the pre-commit git hook scripts:
+# $ pre-commit install
```
toots, line 6: Ultimately, we should start writing a contributor/developper's guide. But this is good for now.

toots (commented):
This looks good. Just one question about the change in of `rev` for the `ocamlformat` check.

.pre-commit-config.yaml
```diff
@@ -23,7 +23,7 @@ repos:
       - id: trailing-whitespace
 
   - repo: https://github.com/savonet/pre-commit-ocamlformat
-    rev: v1
+    rev: d120e7d5a
```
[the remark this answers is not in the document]
jooola: I think the v1 is understood as v1.x.x, using a  tag like v1.0.0 should remove the warning.

toots:
I forked the github action repository and added `pre-commit` setup and opam repository entries for our internal packages over at: https://github.com/savonet/liquidsoap-dev-tools. I think that this could act at the new `liquidsoap-full` for all builds.

PR #2668, opened 2022-10-04
Add shfmt (docker) pre-commit hook

jooola:
Next iteration on the pre-commit setup: This adds a local shfmt pre-commit hook using the https://hub.docker.com/r/mvdan/shfmt docker image.

This puts a hard dependency on docker for the devs, but on the other hand don't force everyone to install the multiple tools required to make pre-commit work (I will add another one (shellcheck) in a future PR).

This was also the way I intended to setup ocamlformat, as it reduces a lot the setup time (think faster CI) when running pre-commit. (Related to https://togithub.com/ocaml-ppx/ocamlformat/pull/2168)

### DIFF
--- a/.github/alpine/liquidsoap.pre-install
+++ b/.github/alpine/liquidsoap.pre-install
@@ -1,7 +1,7 @@
 #!/bin/sh
 
-addgroup -S liquidsoap 2>/dev/null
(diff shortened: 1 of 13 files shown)


toots:
I'm fine with adding `docker` to the dev toolbox. We use it extensively all over the place.

toots (approved):

Issue #2654, opened 2022-09-29
No stdlib in emacs completion


smimram:
I just added support for stdlib in the completion in a79d5820673ab7c485e12f965070fc65da256933. This works when I compile locally:
```
$ dune exec ./gen_emacs_completion.exe|grep interactive
#("interactive.bool" 0 1 (:type "(?description : string, ?osc : string, string, bool) -> () -> bool" :description "Read a boolean from an interactive input."))
#("interactive.float" 0 1 (:type "(?min : float, ?max : float, ?step : float, ?description : string,  ?unit : string, ?osc : string, string, 'a) -> () -> 'a" :description "Read a float from an interactive input."))
...
```
but when I install with opam, the stdlib is not taken in account:
```
$ cat liquidsoap-completion.el |grep interactive
  (interactive (list 'interactive))
    (interactive (company-begin-backend 'company-liquidsoap-backend))
```
@toots any idea about this? I guess it's an intricate problem about the way the generator tries to find the standard library when compiling...

toots:
Dependencies strike again I believe, when it's generated via `dune build -p liquidsoap-mode` then `liquidsoap` is not available and the `stdlib` isn't. We need to make `liquidsoap-mode` depend explicitly on something that installs the libs, either:
1. The full `liquidsoap` package, in which case the package cannot depend on `liquidsoap-mode` itself.
2. A new `liquidsoap-stdlib` package.

I can see advantage on both side. We eventually want to do stdlib releases only so perhaps we should go that route?

smimram:
The 2. seems to be the cleanest I'd say. It's not very urgent though.

PR #2669, opened 2022-10-04
Split libs into their own files

toots:
Fixes: #2654


smimram (approved):
Thanks, go for it!

toots:
@smimram can you confirm that this fixes your issue?

smimram:
Yes it does!
image

PR #2682, opened 2022-10-08
Use operator type for content negotiation.

toots:
This PR scraps out the remains of a input/output mechanism implemented to negotiate sources content at runtime and replaces it with a mechanism to do the same using the declared types.

Now that we use the type system internally for the runtime content negotiation via `source#frame_type`, we should be able to leverage operator's types to make sure that we declare and resolve types in a consistent way during type checking and at runtime time.

The idea of this change is that, e.g. if we have an operator of type:
```
(source(audio=pcm('a)) -> source(audio=pcm('a), video=canvas)
```
Or, for an encoded output:
```
(format('a), source('a)) -> unit
```
We have enough information to apply these constraints at runtime as well. In concrete terms, this means that, when we apply the operator at runtime we:
* Instantiate a copy of the operator's type with fresh universal variables (so as to not change the type of the operator itself) but connected in the same way (for instance `source('a)` and `format('a)` above)
* We look at the operator's arguments and, each time we have a source or a format, we unify the runtime source and format `frame_type` with the type instantiated above.
* We do the same for the returned value.

Here's a simple example with an encoded output. First, the liq scipt:
```
output.file(%mp3(mono), noise())
```
Operator type is: 
```
(format('a), source('a)) -> unit
```
When `output.file` is called at runtime:
* We instantiate a fresh copy of this type
* We calculate the type of `%mp3(mono)`, which is `format(audio=pcm(mono))`
* We unify `format('a)` with this type so now, `'a` becomes `audio=pcm(mono)`
* We unify `noise#frame_type` with `'a` so now, our noise source will produce mono audio pcm data.

This mechanism is generalized to all possible arguments by recursively parsing down all of the operator's arguments in the [check_content](https://github.com/savonet/liquidsoap/pull/2682/files#diff-475b136494f5913b425f6ba66fc00e1eac74619693a532eb57245892dabfa5a5R254) function.


smimram:
I don't understand something from the description: how do you deal with operators changing the frame contents? e.g. `audio_to_stereo`

src/core/lang_source.ml
```diff
@@ -282,15 +331,44 @@ let add_operator =
       :: List.stable_sort compare proto
     in
     let f env =
+      let return_t, proto =
```
smimram, line 334:
```suggestion
      (* Create a fresh instatiation of the return type and the type of arguments. *)
      let return_t, proto =
```

src/core/lang_source.ml
```diff
@@ -282,15 +331,44 @@ let add_operator =
...
+        let generalized, t =
+          Typing.generalize ~level:(-1)
```
smimram, line 336:
```suggestion
          (* TODO: level -1 generalization is abusive, but it should be a good enough approximation for now *)
          Typing.generalize ~level:(-1)
```

src/core/lang_source.ml
```diff
@@ -282,15 +331,44 @@ let add_operator =
...
+          Hashtbl.replace positions name (pos + 1);
+          pos
+        in
+        List.iter
+          (fun (name, typ) ->
+            let v =
+              try Some (assoc name (position name) env) with Not_found -> None
+            in
+            match v with None -> () | Some v -> check_content v typ)
+          proto
```
smimram, line 362: Wouldn't it be easier to look for the nth argument in the environment (where n is the position in the proto, without taking the label in account) rather than bookkeeping this hastable for arguments with same label? I.e. use `List.iteri` instead of `List.iter`.

smimram (changes requested):
Looks mostly ok to me. I am not 100% sure that we won't have corner-cases, but it should be good enough for now :)

src/core/lang_source.ml
toots: Yeah I overthought this.

toots:
> I don't understand something from the description: how do you deal with operators changing the frame contents? e.g. `audio_to_stereo`

Type is:
```
(source(audio=pcm('a)) -> source(audio=pcm(stereo))
```
This is enforced at runtime so calling `audio_to_stereo(s)` asks the source `s` to produce any pcm content and the returned source to produce `pcm(stetreo)` content.

PR #2685, opened 2022-10-11
Remove switch operator, implement simple fallback core operator, use …

toots:
…cross to define top-level fallback operator, define switch using it.


smimram:
I think the best abstraction would be a `select` operator such that
```
select(f, l)
```
with
- `f` of type `unit -> int`
- `l` of type `source list`
plays the `f()`th source.

We can implement fallback over this.

toots:
> I think the best abstraction would be a `select` operator such that
> 
> ```
> select(f, l)
> ```
> 
> with
> 
> * `f` of type `unit -> int`
> * `l` of type `source list`
>   plays the `f()`th source.
> 
> We can implement fallback over this.

Can you explain more in details why? I'm not sure the benefit is immediate to me. 

There's a complex interaction between `source.is_ready` and switch predicates that makes `switch` implementation tricky. When starting with a `fallback` based on `source.is_ready`, the low-level implementation is pretty straight forward. Then we can build on top of that using `source.available` to implement the switch, still following a strict logic based on source availability..

PR #2696, opened 2022-10-19
Make camomile optional

smimram:


toots (changes requested):
This looks great but the various paths need to be brought back:
* On standalone builds (windows platform), the `camomile/` files directory is located in the same directory as the `liquidsoap` binary.
* On packaged builds, the camomile files are installed at `/usr/share/liquidsoap/camomile`

PR #2700, opened 2022-10-21
Deprecate `string_of`.

smimram:


toots (commented):
Nice cleanup! Couple of questions:
* Can we have some kind of way to force a type to string as well? I've seen cases where because a function was only using its argument via `#{argument}`, it was typed as `'a`
* Why the silence ignore? I really don't like those.
* Should we implement our default charset conversion using pure ocaml libraries? There are a couple:
  * https://github.com/mirage/uuuu for ISO to unicode
  * https://github.com/mirage/coin for KOR to unicode
  * https://github.com/mirage/rosetta a mix of all available but seems lagging in dev? Might be the one tho, it's still more recent than camomille.
  * There still isn't unicode to other encodings which I believe would still be a useful optional functionality for people with weird encoding needs (e.g. shoutcast)

smimram:
Thanks!

- We can force a type to string with `(x : string)`.
- If we don't support a particular charset I think that it is best not to fail, because this is not that much of an issue. That being said I will add a warning in the logs.
- Concerning other libs, I see that you have addressed that in #2701 :)

Issue #411, opened 2017-03-23
Live harbor - Bad song title encoding only harbor


desavil:
Oryginal (broadcasting by SHOUTcast DSP): `start ę€ół.śążźćń end` (Polish characters and Euro symbol)
Send to SHOUTcast 2 via input.harbor/output.shoutcast: `start êó³.¹¿æñ end`

Filename: `start ę€ół.śążźćń end.mp3`

Liquidsoap log:
`2017/03/23 14:37:56 [input(dot)harbor_5820:3] New metadata chunk ? -- start êó³.¹¿æñ end.`

If I load this file in the liquidsoap playlist, there is no problem with encoding only if I broadcasts by harbor.

Screens:
![Good](https://i.snag.gy/8cYEGH.jpg?nocache=1490276861539)
![Bad](https://i.snag.gy/NkfyvE.jpg)

toots:
You should look at this parameter in `input.harbor`:
```
 * icy_metadata_charset : string (default: "")
     ICY (shoutcast) metadata charset. Guessed if empty. Default for
     shoutcast is ISO-8859-1. Set to that value if all your clients send
     metadata using this charset and automatic detection is not working for
     you.
```
Try setting it to `"latin1"` or  `"utf8"`

desavil:
I've checked this before.
`"latin1"` - `start êó³.¹¿æñ end`
`"utf8"` - `start êó³.¹¿æñ end`

desavil:
Some solution? I use the newest (compiled) liquidsoap and the problem is still there.

desavil:
Not working - https://i.snag.gy/Hi7xMO.jpg :(

> 2018/08/18 21:46:57 [input(dot)harbor_6446:3] New metadata chunk ? -- start ę�ół.�šż�ćń end.
> 2018/08/18 21:47:05 [camomile:3] Failed to convert "start \196\153\194\128\195\179\197\130.\194\156\197\161\197\188\194\159\196\135\197\132 end" from auto(UTF-8,ISO-8859-1) to Latin-1 (CamomileLibrary__UChar.Out_of_range)!
> 2018/08/18 21:47:05 [camomile:3] Failed to convert "start \196\153\194\128\195\179\197\130.\194\156\197\161\197\188\194\159\196\135\197\132 end" from auto(UTF-8,ISO-8859-1) to Latin-1 (CamomileLibrary__UChar.Out_of_range)!

toots:
[...]

PR #2701, opened 2022-10-21
Use rosetta as default charset conversion library

toots:
This PRs expand the native charset conversion support to include conversions from the formats supported by `rosetta` and by OCaml version >= `4.14`.

It also adds tests based on the `camomile` tests dataset.

The introduction of `conf-liquidsoap-charset` is because of: https://github.com/ocaml/dune/issues/6284


tests/core/charset_test.ml
```diff
@@ -0,0 +1,118 @@
...
+    (fun enc ->
+      let enc_str = Charset.to_string enc in
+      match (fetch (encoded_file enc), fetch (utf8_file enc)) with
+        | Some encoded, Some utf8 ->
+            if
+              List.mem enc Charset.can_decode
+              && List.mem `UTF_8 Charset.can_encode
+            then (
+              Printf.printf "%s\n%!"
+                [%string "Testing conversion from %{enc_str} to UTF8"];
```
smimram, line 74: I don't really see the point of adding a dependency on `ppx_string` which is not standard here instead of simply using `printf`:
```suggestion
              Printf.printf "Testing conversion from %s to UTF8\n%!" enc_str
```
It makes the code longer and less readable...

(of course this comment also applies on other prints in the file...)

src/core/dune
```diff
@@ -22,10 +22,10 @@
   liquidsoap_lang
   menhirLib
   (select
-    charset.ml
+    charset_impl.ml
     from
-    (camomile -> charset.camomile.ml)
-    (-> charset.native.ml)
+    (camomile -> charset_impl.camomile.ml)
+    (rosetta -> charset_impl.rosetta.ml)
```
smimram, line 28: Why not keep the "native" implementation? I agree that we should ask for rosetta or camomil in the opam file, but we could still keep the native implementation around to still be able to compile with dune if we have a problem with rosetta, or want to have a small binary etc?
```suggestion
    (camomile -> charset_impl.camomile.ml)
    (rosetta -> charset_impl.rosetta.ml)
    (-> charset_impl.native.ml)
```

smimram (approved):
Thanks for this work, this is much more complete than my first implementation.

Aparat from the questions in the review, I have a general one: why not simply drop camomile support in favor of rosetta instead of keeping both, which requires introducing a new opam package... ?

toots:
Thanks for the review. I'm not sure that I want to go this route now.. 😅

Moving forward, we want to make sure that everything is encoded with the de-facto `UTF-8` standard. However, and particularly for a project like us, the ability to decode and encode in outdated character encoding is particularly important.

In the past, we've had reports about issues with legacy software. See for instance: https://github.com/savonet/liquidsoap/issues/411. 

`output.shoutcast` still to that day sends metadata in `latin1` encoding for legacy `icy` protocol. The original reason was that icy metadata in mp3 streams have historically been in this encoding and many encoders will expect it. Reading more (see: https://stackoverflow.com/questions/31044773/liquidsoap-tag-encoding) it looks like we should default back to `UTF-8` now.

However, in this case we do want to give the user a chance to encoding in `latin1` if it's still needed.

For ocaml dependencies with external dependencies, it makes sense to make them optional but a package with pure ocaml code should be made mandatory if it's important enough and I believe that our capacity to encode and decode to more than `ISO` and `unicode` encodings is.

What we should really consider here is a friendly, community supported fork of `camomile` with support for `ocaml` 5 merged into it.

PR #2706, opened 2022-10-24
Allow optional methods

toots:
This PR introduces optional record methods. 

The operational part is:
```ruby
x?.foo
```
This means that the method `foo` will be invoked on `x` but, if `foo` is not found,`null()` will be returned instead.

Optional methods are denoted using a similar `?:` syntax:

```ruby
(x:{foo?:int})
```

Notes:
* We allow `{foo:int?} <: {foo?:int}` and `{foo?:int?} <: {foo?:int}` and prohibit `{foo?:int} <: {foo:int?}` so that if `x.foo` is typed `int?`, this means that `x` does indeed have a `foo` method.
* Both `()` and `{foo:int}` can be unified in lists with `{foo?:int}`
* ~~The idea of adding different values than `null()` for the method's value when it does not exist on the caller has been explored.~~
 
~~It sounds like a great idea b/c it would bridge the gap between function arguments and records. However, it is not clear how to type a method with a default value.  Let's consider this:~~
```
y = x.foo ?? 123
```
~~Or any equivalent syntactic declaration for "let's find method `foo` or else return `123`". It really seems that this shouldn't be reflected in `x`'s type except from the fact that `x` may not have the method `foo`.~~

Well, this has been done. See below for details!

* A great follow-up would be a type annotation for "any type without method `foo`". In this case, it should be possible to define an unreachable type, for instance named `never` and do:
```
'a.{foo?: never}
```
This could be used to define a syntactic sugar to remove a method from a type:
```
f: 'a -> 'a.{foo?: never}
```

### DIFF
--- a/doc/content/language.md
+++ b/doc/content/language.md
@@ -987,7 +987,7 @@ The type of this function is:
 f : (int, 'a.{add? : int}) -> int = <fun>
 ```
 
-Which denotes that the `options` argument can be any value that may or may not have
+which denotes that the `options` argument can be any value that may or may not have
 a `add` field. However, if this field is present, it must be of type `int`.
 
 2. Using the `x?.foo` syntax
@@ -998,7 +998,7 @@ otherwise.
 The `?.` syntax can be chained and works with functions, which make it a very convenient
 way to drill deep inside nested records:
 
-```ruby
+```liquidsoap
 x?.fn(123, "aabb")?.field
 ```
 
--- a/doc/content/language.md
+++ b/doc/content/language.md
@@ -961,6 +961,47 @@ will print
 
 (note that the string is modified but not the fields `duration` and `bpm`).
 
+### Optional fields
+
+During the execution of your script, it can be useful to allow functions
+to receive records that may or may not have a specific field. This can
+be used, for instance, to model optional arguments.
+
+This can be achieved in two ways:
+
+1. Using the `x.foo ?? default` syntax
+
+Here's an example:
+
+```liquidsoap
+# This functions adds 1 to x unless options has a
+# add field in which case it adds this value
+def f(x, options) =
+  x + (options.add ?? 1)
+end
+```
+
+The type of this function is:
+
+```liquidsoap
+f : (int, 'a.{add? : int}) -> int = <fun>
+```
+
+Which denotes that the `options` argument can be any value that may or may not have
+a `add` field. However, if this field is present, it must be of type `int`.
+
+2. Using the `x?.foo` syntax
+
+Given a variable `x`, `x?.foo` returns the field value `foo`, if present, or `null`
+otherwise.
+
+The `?.` syntax can be chained and works with functions, which make it a very convenient
+way to drill deep inside nested records:
+
+```ruby
+x?.fn(123, "aabb")?.field
+```
+
 ## Patterns
 
 As explained earlier, you can use several constructions to extract data from structured values such
--- a/src/lang/parser.mly
+++ b/src/lang/parser.mly
@@ -146,7 +146,7 @@ open Parser_helper
 %type <Term.pattern> meth_pattern
 %type <Parser_helper.meth_pattern_el> meth_pattern_el
 %type <Parser_helper.meth_pattern_el list> meth_pattern_list
-%type <string * Type.t * string option> meth_ty
+%type <Parser_helper.meth_ty_opt> meth_ty
 %type <Term.t option> opt
 %type <string> optvar
 %type <Term.pattern> pattern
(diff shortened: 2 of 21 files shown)


doc/content/language.md
```diff
@@ -961,6 +961,31 @@ will print
...
+be used, for instance, to model optional arguments.
+
+Liquidsoap supports that functionality using the `?.` syntax. Here's an example:
+
+```liquidsoap
+# This functions adds 1 to x unless options has a
+# add field in which case it adds this value
+def f(x, options) =
+  x + (options?.add ?? 1)
+end
```
smimram, line 977: I am not so fond of this `.?` syntax and moreover it seems to me that providing a default value is more primitive than returning `null` if we have nothing: most of the time we want to provide a default value. I therefore propose  
to add a new syntax `x.field ??? value` which would be parsed as such and would be nicer.

```suggestion
def f(x, options) =
  x + (options.add ??? 1)
end
```

You can always implement your `x?.field` as `x.field???null()` if needed.

On the long run, I think we will be able to drop most `null` values and leave the `??` syntax for records with optional fields (we can leave this for after, I don't expect that the syntax will be widely used soon).

doc/content/language.md
```diff
@@ -961,6 +961,31 @@ will print
...
+```
+
+The type of this function is:
+
+```liquidsoap
+f : (int, 'a.{add? : int?}) -> int = <fun>
```
smimram, line 983: I don't think that the field is nullable here:
```suggestion
f : (int, 'a.{add? : int}) -> int = <fun>
```

src/lang/parser.mly
```diff
@@ -146,7 +146,7 @@ open Parser_helper
 %type <Term.pattern> meth_pattern
 %type <Parser_helper.meth_pattern_el> meth_pattern_el
 %type <Parser_helper.meth_pattern_el list> meth_pattern_list
-%type <string * Type.t * string option> meth_ty
+%type <string * Type.t * bool * string option> meth_ty
```
smimram, line 149: A 4-uple: maybe should we start using a record here?

src/lang/typing.ml
```diff
@@ -316,10 +314,22 @@ and bind ?(variance = `Invariant) a b =
...
 (** Ensure that the type for the method [l] in [a] is a subtype of the one for the same method in [b]. *)
 and unify_meth a b l =
-  let { meth = l; scheme = s1; json_name = json_name1 } = get_meth l a in
-  let { scheme = s2; json_name = json_name2 } = get_meth l b in
+  let { optional = optional1; meth = l; scheme = s1; json_name = json_name1 } =
+    get_meth l a
+  in
+  let { optional = optional2; scheme = s2; json_name = json_name2 } =
+    get_meth l b
+  in
```
smimram, line 322: This would be more symmetric and safer:
```suggestion
  let { optional = optional1; meth = meth1; scheme = s1; json_name = json_name1 } =
    get_meth l a
  in
  let { optional = optional2; meth = meth2; scheme = s2; json_name = json_name2 } =
    get_meth l b
  in
  assert (meth1 = l && meth2 = l);
```

smimram (commented):
This is a very welcome addition, thanks!

I am not approving yet, because I'd really prefer my alternative syntax `x.l ??? default`.

doc/content/language.md
toots: Nah that's actually the expected type, unfortunately. in `x?.foo ?? 123`, `x.foo` can either:
* Have no method
* Have a method that returns an `int`
* Have a method that returns `null`

toots:
Thanks for the review. I've applied most of the comments.

Regarding the `?.` construct, it is widely used and known from javascript/typescript and ruby. Having something familiar helps with the language's adoption.

Also, the PR has now been updated to allow chaining `.` and `?.` calls (see below for details). This makes it much easier to dig through highly nested values, for instance:

```
let json.parse x = data
nested_optional_value = x?.foo.gni?.bla
```

However, I was also able to implement your `???` suggestion! It did not need a new `???` operator, we can catch `x.foo ?? default` during parsing.

The new implementation works as follows:
* `Invoke` now takes a `default` value instead of a `optional` boolean
* At parsing, `x.foo ?? default` becomes: `Invoke(x, "foo", default = Some default)`
* At parsing, `x?.foo` becomes: `Invoke(x, "foo", default = Some null)`
* At typecheck of `Invoke(x, "foo", default)`
  * `x.foo` is typechecked
  * if `default` is present, it is typechecked as a subtype of `x.foo`: `default <: typeof x.foo`
* At runtime of `Invoke(x, "foo", default)`
  * If method `"foo"` is found on `x` and default is `None`, return `x.foo`
  * If method  `"foo"` is found on `x` with a default and `x.foo` is `null`, return `default`
  * If method `"foo"` is not found with a default, return `default`

This is pretty cool. In particular, we now type:
```ruby
def f(x) =
     (x.foo ?? 2) + 1
  end
```
as: `f : ('a.{foo? : int}) -> int`

But, also, still accept `f({foo = null()})`

The reason for accepting `{foo = null()}` as type `{foo?: int}` is that:
* Operationally speaking, this is equivalent
* This allows some very nice symmetry in record spreads and default values, for instance:
```ruby
def f(x) =
  x.foo ?? 1
end
# f : ('a.{foo? : int}) -> int

def g(x = null()) =
  f({foo = x})
end
# g : (?int?) -> int
```
This makes is possible to re-package `null` values passed e.g. as default arguments to a function as record with optional fields. At scale, this is a very powerful and readable pattern.

Lastly, nested `.` and `?.` calls are supported as follow:
* When parsing an invoke of `"bla"` on `x.foo` with default `default`.
  * If `x.foo` has a non-null default, a `bla` method with value `value` is added to it
  * If `x.foo` has no default or a null default, default is set to `Any.{bla: default}`.
  * `Never` is a new term introduced to be able to craft defaults with methods where the base value is never reache. This makes it possible to have this type:
```
# fn = fun (x) -> x?.foo.gni;;
fn : ('c.{foo? : 'b.{gni : 'a?}}) -> 'a? = <fun>
```
The default value for `x?.foo` is `Never.{gni = null}` and we can call the function as:
```
# fn({foo = 123.{gni = "aabb"}});;
- : string? = "aabb"
```

toots:
I forgot to say: there are still limitations for chained calls with function application with optional values
```
x?.foo.gni(1,2,3).gno
```
This is due to some difficulty in properly not propagating the `nullable` types onto the default values.

I believe the current implementation is type safe so we can start this way and refine if we find a good way to make it work.

This call still works:
```
# fn = fun (x) -> x.foo.gni(1,2,3).gno.gni;;
fn :
  ('e.{foo : 'd.{gni : (int, int, int) -> 'c.{gno : 'b.{gni : 'a}}}}) -> 'a =
  <fun>
```

doc/content/language.md
```diff
@@ -961,6 +961,47 @@ will print
...
+end
+```
+
+The type of this function is:
+
+```liquidsoap
+f : (int, 'a.{add? : int}) -> int = <fun>
+```
+
+Which denotes that the `options` argument can be any value that may or may not have
```
smimram, line 990:
```suggestion
which denotes that the `options` argument can be any value that may or may not have
```

doc/content/language.md
```diff
@@ -961,6 +961,47 @@ will print
...
+
+2. Using the `x?.foo` syntax
+
+Given a variable `x`, `x?.foo` returns the field value `foo`, if present, or `null`
+otherwise.
+
+The `?.` syntax can be chained and works with functions, which make it a very convenient
+way to drill deep inside nested records:
+
+```ruby
```
smimram, line 1001:
```suggestion
```liquidsoap
```

src/lang/typechecking.ml
```diff
@@ -220,14 +223,30 @@ let rec check ?(print_toplevel = false) ~throw ~level ~(env : Typing.env) e =
...
+        let vars, typ = aux a.t in
+        let typ =
+          match default with
+            | None -> typ
+            | Some v ->
+                check ~level ~env v;
+                (* We want to make sure that: x?.foo types as: { foo?: 'a } *)
+                let typ =
+                  match (Type.deref v.t).descr with
+                    | Type.Nullable _ -> mk Type.(Nullable typ)
```
smimram, line 243: If we have a universal variable here, isn't there a chance that we build a "double nullable" here? (like `int??`) I don't think we want that... For instance, we would have `'a?` and then `'a` gets substituted by `int?`. Or maybe should this be solved by a "not nullable" constraint on universal variables? I don't think we have to solve this *right now* but it would be nice to have a comment or keep a BR open so that we don't entirely forget about it.

tests/language/json.liq
```diff
@@ -208,25 +199,45 @@ def f() =
...
+
+  data = '{
+    "foo": {
+      "gni": {
+        "bla": 123
+      }
+    }
+  }'
+  let json.parse x = data
+  test.equals(x?.foo.gni?.bla, null(123))
```
smimram, line 230: Do we really need the cast here?
```suggestion
  test.equals(x?.foo.gni?.bla, 123)
```

smimram (approved):
Apart from minor things (see the comments), I think that this is now good to go!

tests/language/json.liq
toots: Not for the value test but for the type test I think it's good to have, to make sure that this returning `int?`.

src/lang/typechecking.ml
toots: You might still want the double `null` I think, for instance if the default value is expected to be a nullable thing:
```ruby
# def f(x) =
    v = x?.foo
    if (null.defined(v)) then
      null.get(v) ?? 1
    else
      2
    end
  end;;
f : ('a.{foo? : int?}) -> int = <fun>
```

toots: Ha yeah no that doesn't work.. 😅

PR #2711, opened 2022-10-26
Error enhancements

toots:
This PR introduces several enhancements to runtime error handling:
* Runtime errors are now marked as `private` so that the dedicated functions have to called to manipulate them
* Raising errors is done solely through `Runtime_error.raise` (for code internal to the language) and `Lang.raise_error` for code using the language
* Building on this, we can now register error listeners that are called whenever an error is raised. These can be used to monitor and report errors, for instance using `Sentry`.
* Error positions are enforced across the board. In particular, a new mechanism is introduced wherein a special `_pos_` environment variable is passed to functions are run-time, allowing them to grab the location of their execution in the code and use it when raising a runtime error.
* Errors full trace is exported in the liquidsoap side:
```ruby
- : error
    .{
      kind : string,
      message : string,
      trace : [
               {
                 position_end :
                 {
                   character_offset : int,
                   filename : string,
                   line_number : int
                 },
                 position_start :
                 {
                   character_offset : int,
                   filename : string,
                   line_number : int
                 },
                 to_string : (?prefix : string) -> string
               }]
```
This means that the user should now be able to grab the file/line/character of each error start/stop position, for reporting and potentially editor's highlight and etc.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -28,6 +28,9 @@ Changed:
 - Added support for a Javascript build an interpreter.
 - Removed support for `%define` variables, superseded by support for actual
   variables in encoders.
+- Errors now report proper stack trace via their `trace` method, making it
+  possible to programmatically point to file, line and character offsets
+  of each step in the error call trace (#2712)
 - Reimplemented `harbor` http handler API to be more flexible. Added a new
   node/express-like registration and middleware API (#2599).
 - Switched default persistence for cross and fade-related overrides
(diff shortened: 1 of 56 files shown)


src/lang/runtime_error.ml
```diff
@@ -36,7 +36,13 @@ let () =
...
 
-let error ?bt ?(pos = []) ?(message = "") kind =
+let listeners = Atomic.make []
+let on_error fn = Atomic.set listeners (fn :: Atomic.get listeners)
+let make ?(message = "") ~pos kind = { kind; msg = message; pos }
+
+let raise ?bt ?(message = "") ~pos kind =
+  let err = { kind; msg = message; pos } in
+  List.iter (fun fn -> fn err) (Atomic.get listeners);
   let e = Runtime_error { kind; msg = message; pos } in
```
toots, line 46:
```suggestion
  let e = Runtime_error err in
```

smimram:
Very nice, thanks!

PR #2723, opened 2022-11-04
Rewrite generators

toots:
The PR rewrites the generators to be ready to support multiple channels in sources.

**Naming note:** OCaml has a `Buffer` module so, in Liquidsoap, we call buffer `Generators`. When clear, the two words may be interchangeable.

**Some context:** "switch to immutable" content is a plan that's been in the work for a while. The old content framework was using frames of fixed content length that was copied/overwritten on each streaming cycle with break marks marking how much had been filled. This has proven complicated for many reasons, in particular a semantic confusion between partial fill and end of tracks and also very specific requirements on breaks.

Instead, we would like to switch to a model where:
* Content are immutable chunks of data with `offset` and `length`
* Getting a smaller slice of a piece of content just means adjusting `offset` and `length`
* Filling up a frame means concatenating immutable content chunks.

Key takeways:

#### The full switch to immutable content should happen with OCaml 5

We are planning a deep rewrite of the streaming model and loop using OCaml 5 effects. This will require a lot of internal changes. With what was learned here, it seems that we should just do both at the same time. Changing to immutable content  already requires touching all these parts and, with all the quirks we have accumulated over the years b/c of the break/track mark situation, we would have to work around that a lot just to remove it when changing the streaming model.

#### Breaks as both track marks and filling marks is definitely a first-class footgun.

I cannot count the number of times I tripped on this doing those changes. Fortunately, we have a lof of tests and internal assertions to catch it.

#### The whole concept of `PTS` (Presentation TimeStamp) is totally removed from our internals

This concept was introduced early in the `2.0.x` cycle with the hope of mimicking what `ffmpeg` does but there are striking differences from our point of view:
* We do not do any consistent muxing/demuxing
* We allow source switch dynamically

Therefore, it should be possible to:
* Take a `audio/video` source
* Split `audio` and `video`
* Initialize another `audio2` source
* Delay and add a flanger effect to the `audio` source, creating a new `audio3` source
* Create a source with the one `video` track, alternating between the 3 audio sources

In a situation like this, there is no way to add a consistent notion of `PTS` that will support what the user is trying to do.

`FPS` as a concept still exist for everything `ffmpeg`. In particular, we do have a mechanism that took us a long time to perfect to keep consistent `PTS` accros streams when outputting to `ffmpeg`. This mechanism will indeed guarantee that audio and video keep in sync when demuxing and remuxing encoded tracks, which is what most people would expect.

#### Generators use the new immutable content

With no PTS, generators become really much more simpler. All they do is queue up chunks of immutable content. The buffered length is the smallest of all queued content. This becomes a little more hairy with discrete content, more on this later.

With the changes, generators are now initialized with a `content_type`, to which metadata and track marks (the future replacement to breaks) are added.

Frame are now also generators, initialized with content of the frame's size. This creates a little bit of a conundrum but allow for one single implementation for everything holding content. Eventually, with the switch to immutable content, frame as a concept will disappear and sources will just be holding a buffer of the content that they have produced during the current streaming cycle.

To this purpose, sources now hold a default `buffer`, initialized with the source's `content_type`, making it simpler to store and retrieve data for a given source.

#### Timed content is still imperfect

The general idea for immutable content is that content would be _appended_ to the buffer and then, when enough content has been added for each type of content, we slice off one chunk to output.

This makes it a little tricky for timed content, though. Imagine a situation where bot `audio` and `video` content have been produced but no metadata was generated. 

In this case, we may have some audio and video queued but no metadata:
```
----------
| audio  |
----------
---------------
| video       |
---------------
-
| metadata
-
```

In such a situation, we may not be able to output a content chunk both `audio`/`video` and `metadata` because the metadata content is still empty.

In the future, we will want to make sure that each streaming cycle does generate a metadata content chunk, possibly with no metadata in it. However, this was too much of a change for this PR.

Therefore, timed content have been changed to allow _infinite_ content chunks. These content chunks have infinite length. One can add metadata/track marks at any point in the future and it's possible to either get a finite slice of them or truncate then when advancing the buffer:

```
----------
| audio  |
----------
---------------
| video       |
---------------
-----------------------------------------------------
| metadata                                          ...................
-----------------------------------------------------
```

While this is neat, this is also a foot gun which should be removed once we switch to immutable content.

With all being said, let's review, comment, merge and move on to the fun part: adding support for generic track muxing, demuxing and multi track support!

### DIFF
--- a/src/core/stream/generator.mli
+++ b/src/core/stream/generator.mli
@@ -20,305 +20,84 @@
 
  *****************************************************************************)
 
-(** Operations on generators, which are like FIFO for multimedia data. They are
-    efficiently handling small chunks of data (as found in frames) with minimal copy,
-    and also metadata.
-
-    Data coming from frames is copied when added to the buffer because frame content
-    are reused during each streaming loop. However, data is only assigned exiting the
-    buffer. This means that we expect the buffer to feed a single source on the way
-    out. *)
-
-(** Raised when trying to feed a generator with data of incorrect type (wrong
-    number of audio channels, etc.). *)
-exception Incorrect_stream_type
-
-type content = Content.data Frame.Fields.t
-
-val content : Frame.t -> content
-
-module NoneContent : sig
-  include Content.Content
-
-  val lift_data : length:int -> unit -> Content.data
-end
-
-(** Signature for a generator. *)
-module type S = sig
-  type t
-
-  (** Length of the generator in ticks. *)
-  val length : t -> int
-
-  (** Duration of data (in ticks) before the next break, -1 if there's none. *)
-  val remaining : t -> int
-
-  (** Clear the generator. *)
-  val clear : t -> unit
-
-  (** Fill a frame from the generator. *)
-  val fill : t -> Frame.t -> unit
-
-  (** Forget a given duration (in ticks) of the generator. *)
-  val remove : t -> int -> unit
-
-  (** Add metadata. *)
-  val add_metadata : ?pos:int -> t -> Frame.metadata -> unit
-end
-
-(** Content-agnostic generator. *)
-module Generator : sig
-  (** A generator. *)
-  type 'a t
-
-  (** Create a generator. *)
-  val create : unit -> 'a t
-
-  (** Empty a generator. *)
-  val clear : 'a t -> unit
-
-  (** Length of data contained in a generator (in ticks). *)
-  val length : 'a t -> int
-
-  (** Remove given amount (in ticks) of data. *)
-  val remove : 'a t -> int -> unit
-
-  (** [put data ofs len] adds [len] of data [data] starting from [ofs]. Data is
-      not copied. *)
-  val put : 'a t -> 'a -> int -> int -> unit
-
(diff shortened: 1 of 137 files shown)


src/core/stream/generator.mli
```diff
@@ -20,305 +20,33 @@
...
+val put : t -> Frame_base.field -> Content.data -> unit
+val get : ?length:int -> t -> Content.data Frame_base.Fields.t
+val peek : t -> Content.data Frame_base.Fields.t
+val add_metadata : ?pos:int -> t -> Frame_base.metadata -> unit
+val add_track_mark : ?pos:int -> t -> unit
+
+val feed :
+  ?offset:int -> ?length:int -> ?fields:Frame_base.field list -> t -> t -> unit
+
+val fill : t -> t -> unit
```
smimram, line 52: Please, please, please: **comments** here!

smimram (approved):
Looks fine to me thanks! Althgouh I cannot claim to have had a very detailed look at it (I mostly trust you on this one).

PR #2737, opened 2022-11-13
Record spread and `never` type

toots:
This PR does two things:

#### Add an impossible `never` type when removing record methods

Example:
```ruby
def f(x) =
 let y.{foo, gni} = x
 y
end

f : ('c.{foo : 'b, gni : 'a}) -> 'c.{foo? : never, gni? : never} = <fun>
```

This makes sure that the returned `y` value never has any `foo` or `gni` method, which is the runtime behavior. 

Before that, the following was a correct function:
```ruby
def f(x) =
  let y.{foo} = x
  y.foo + 1
end

f : ('b.{foo : 'a}) -> int = <fun>
```

#### Add syntactic sugar for record and record pattern spread

```ruby
# This:
let y.{foo, gni} = x

# Can be written as:
let {foo, gni, ...y} = x

# This:
y = x.{foo = 123, gni = "aabb"}

# Can be written as:
y = {foo = 123, gni = "aabb", ...x}
```

These spread patterns are much more common nowadays and will help programmers being more confident in learning and understanding the language.

Ideally, we would like the spread to only apply to records (unit values with methods) and not other values, for instance:
```ruby
x = 1.{foo = 123}
let { ...y } = x
y : { foo = 123}
```
However, I don't think that this is currently possible with the existing type system because we need to keep the root type universal in case this becomes wrapped into a function:
```ruby
def f(x) =
  let {foo, gni, ...y } = x
   y
end

f : ('c.{foo : 'b, gni : 'a}) -> 'c.{foo? : never, gni? : never} = <fun>
```

There isn't a way currently to both say that `y` should be a record _and_ that it should retain all the other methods from `x`. This might be possible with e.g. a constraint but I didn't want to enter that territory yet as I believe non-unit values with methods are still pretty advanced so why not let the advanced programmer know what they are doing?

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -18,6 +18,8 @@ New:
   - `file.replaygain` to compute the replaygain of a file
 - Added support for ImageLib to decode images.
 - Added support for completion in emacs based on company (#2652).
+- Added syntactic sugar for record spread: `let {foo, gni, ..y} = x`
+  and `y = { foo = 123, gni = "aabb", ...x}` (#2737)
 
 Changed:
 
--- a/doc/content/language.md
+++ b/doc/content/language.md
@@ -879,6 +879,26 @@ the duration with
 print("The duration of the song is #{song.duration} seconds")
 ```
 
+Records can be re-used using _spreads_:
+
+```liquidsoap
+song = { filename = "song.mp3", duration = 257., bpm = 132. }
+
+# This is a fresh value with all the fields from `song` and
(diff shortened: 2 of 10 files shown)


toots:
> I think that this is good to go.
> 
> Small question: do we want to introduce a syntax to hide fields? Currently, this can be done by matching, but this is a bit heavy...

Thank! It'd be nice to have a syntax for it. This could be done is a different PR.

PR #2739, opened 2022-11-15
Decode any number of audio and video track with ffmpeg.

toots:
This PR adds support for encoding and decoding multi-track files using `%ffmpeg`. It is written in anticipation of the soon-to-come support for multi track at the language level.

On the encoding side, `%ffmpeg` encoders can now take any number of named tracks, for instance:
```
%ffmpeg(
  format="mvk",
  %audio_en(codec="aac", ...),
  %audio_fr(codec="aac", ...),
  %video(codec="libx264", ...)
)
```

Internally, this will be converted into a format of type: `format(audio_en=pcm, audio_fr=pcm, video=canvas)` which, in turn will inform the various operators that we are expecting content of this type.

Because of this, we need to know if a given track is audio or video (and later subtitle). The proposed convention is, in order of preference:
* If the track has `audio_content` or `video_content` as parameter (for instance `%foo(audio_content, ...)`) then it is considered, resp., `audio` or `video`.
* If the track name has `audio` or `video` in it (for instance `%dolby_audio_fr`) then it is considered, resp., `audio` or `video`
* If the track codec is hardcoded (for instance (`%foo(codec="aac", ...)`) then the codec is used to detect the content.

This also has the benefit of keeping everything backward compatible with existing code.

The decoder is also updated to support multitrack. This provides a limited support for multi-track end-to-end decoding and encoding that will be expanded later with the support of language-level track manipulation. Here's a simple end-to-end example:

```
# This a movie with two audio tracks (EN and FR) in a mkv container
s = single("/path/to/movie.mkv")

# We re-encode audio tracks to aac and keep video and mux in a mp4 container:
encoder = %ffmpeg(
  format="mp4",
  %audio(codec="aac",...),
  %audio_2(codec="aac", ...),
  %video.copy
)

output.file("/path/to/movie.mp4", encoder, s)
```

Technical notes:
* The ffmpeg copy audio and video content is merged into a single copy content. This makes sense since, with copy content, we are passing everything verbatim so we shouldn't even care about audio vs. video. This allows to simplify the encoder logic (no need for the above contention) but comes at a slightly annoying complication due to the fact that encoded content are types audio/video at the library level).
* ~~The convention for default tracks names is: `%audio` for first audio track, then `%audio_1`, `%audio_2` etc. While this is pretty intuitive for programmers, this might not be as easy for people used to `1-based` counting where the first object starts at `1`. Might change that later.~~
* Inline encoding and decoding operators are not heavily tested against those changes. It is anticipated that they will be reworked next to encode at the track level, getting rid of the multiple `audio`, `video` and `audio_video` variants. They shall be tested more intensively by then.
* There needs to be a way to ensure predictability and stability of a simple decoder/encoder with multi-track so that the same order of tracks is guaranteed in the output file compare to the input files. Since we don't have much to work on those at the moment, it seems that this is something to re-visit with concrete examples when the whole multi-track framework is in place so that this doesn't get over-engineered right now. The two possibilities that I have in mind are number indexes or string tags.

### DIFF
--- a/src/core/decoder/decoder.ml
+++ b/src/core/decoder/decoder.ml
@@ -61,8 +61,8 @@ type fps = Decoder_utils.fps = { num : int; den : int }
     - Implicit content drop *)
 type buffer = {
   generator : Generator.t;
-  put_pcm : samplerate:int -> Content.Audio.data -> unit;
-  put_yuva420p : fps:fps -> Content.Video.data -> unit;
+  put_pcm : ?field:Frame.field -> samplerate:int -> Content.Audio.data -> unit;
+  put_yuva420p : ?field:Frame.field -> fps:fps -> Content.Video.data -> unit;
 }
 
 type decoder = {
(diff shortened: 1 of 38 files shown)


src/core/decoder/decoder.ml
```diff
@@ -419,65 +419,94 @@ let get_stream_decoder ~ctype mime =
...
+            let data, _, _ = resampler ~samplerate data 0 (Audio.length data) in
+            let data = (get_channel_converter ()) data in
+            let data = Content.Audio.lift_data data in
+            (*
+Printf.printf "put pcm for %s\n%!" (Frame.Fields.string_of_field field);
+Printf.printf "Buffer content_type: %s\n%!" (Frame.string_of_content_type (Generator.content_type generator));
+Printf.printf "Buffer len: %d\n%!" (Generator.length generator);
+Printf.printf "Buffer buffered len: %d\n%!" (Generator.buffered_length generator);
+Frame.Fields.iter (fun field c -> Printf.printf "%s len: %d\n%!" (Frame.Fields.string_of_field field) (Content.length c)) (Generator.peek generator);
+*)
```
smimram, line 460: Remove this debug, unless you really think that it could be useful in the future:
```suggestion
```

smimram (approved):
Looks fine to me (although I did not particularly try to understand the details...)

PR #2748, opened 2022-11-22
Use glob for stdlib install.

toots:


smimram:
This is much better, thanks! Do you think, we could also use this to remove `dune.inc` in tests?

toots:
> This is much better, thanks! Do you think, we could also use this to remove `dune.inc` in tests?

Not really, no. There's no support for a `each` or `map` operation on those it seems.

toots:
This PR would require rebuilding all docker images.. I think it unfortunately has too limited benefits vs. cost at the moment. Maybe after the `2.2.x` release?

smimram:
No need to rush as long as it gets merged at some point!

smimram (approved):

toots:
This is done now,

PR #2752, opened 2022-11-24
Abstract module names

toots:
This is a redo of https://github.com/savonet/liquidsoap/pull/2479 making module names abstract so that inter-dependencies in module/operator registration is reflected in the OCaml modules dependencies.

This time, let's try to merge quickly before it starts conflicting.. 🙂


src/core/lang.mli
```diff
@@ -108,24 +108,26 @@ val add_builtin :
   ?flags:Doc.Value.flag list ->
   ?meth:(string * Liquidsoap_lang.Type.scheme * string * value) list ->
   ?examples:string list ->
+  ?base:module_name ->
   string ->
   proto ->
   t ->
   (env -> value) ->
-  unit
+  module_name
```
smimram, line 116: I don't really see the point in returning the module name here. Could you explain?

toots: This how you track dependencies. Module names are abstract types so, if you want to, say, register module `foo.bar`, you have to:
```ocaml
let foo = Lang.add_module "foo"
let foo_bar = Lang.add_module ~base:foo "bar
```
Same thing with operators:
```ocaml
let blank = Lang.add_operator "blank" ...

let blank_detect = Lang.add_operator ~base:blank "detect"
```

This way, if you have to extend a given record, you need to fetch its base/root from the module that defines it, thus creating the module dependencies required to order them at compilation time.

toots:
This should be ready for final review/merge. I've confirmed that the API hasn't changed by generating a diff of the JSON functions dump.

PR #2764, opened 2022-11-30
Add support for generic track manipulation

toots:
This PR enables generic source tracks manipulation! 🎉

The new APIs are:

### Tracks getter:

* `source.tracks` returns the tracks of a given source:
```
% liquidsoap -h source.tracks
Return the tracks of a given source.

Type: (source('a)) -> 'a

Category: Source / Track processing

Arguments:
 * (unlabeled) : source('a)

Methods:
 * track_marks : track_marks
 * metadata : metadata
```
This can be a bit confusing because it is actually very simple! It takes a source of type `source('a)` and returns a record of type: `'a.{track_marks: track_marks, metadata: metadata}`, that is a record of the source's media tracks along with two tracks representing track marks and metadata.

For instance:
```
# source.tracks(sine());;
- : {
  audio : pcm(stereo),
  metadata : metadata,
  track_marks : track_marks
} = {
  audio = track(source=sine_0,field=audio),
  track_marks = track(source=sine_0,field=track_marks),
  metadata = track(source=sine_0,field=metadata)
}
```

**Note:** I initially had this as a source method but it proved too burdensome, both on the runtime (methods are calculated at the value's evaluation where the full source content type might not yet been known) and in the types, where it was polluting all source operators types..

### Tracks muxer:

Tracks can be recombined into a new source using the `source()` muxer, including with or without track marks, for instance:
```
# Two audio sine:
s1 = sine(440.)
s2 = sine(880.)

# A video clip:
s3 = single("path/to/video/file.mp4")

# recombine them all, using track marks and metadata from the video file:
let { video, track_marks, metadata } = source.tracks(s3)
s = source({
  audio = source.tracks(s1).audio,
  audio_2 = source.tracks(s2).audio,
  video = video,
  metadata = metadata,
  track_marks = track_marks
})

# New ways of dropping tracks:
s = single("...")

# tracks below is all of s' tracks without track marks:
let {track_marks=_, ...tracks} = source.tracks(s)
source(tracks)
```

### Encoding

For now, only the `%ffmpeg` encoder has support for multitrack. See https://github.com/savonet/liquidsoap/pull/2739 for details

### What is a track?

From the example above, you can see that a track is, in fact, just a reference to a source and a given media field, `audio`, `video`, etc. In other words tracks are just _sources restricted to a single media field_. This means that tracks are also subject to the same restrictions as source, in particular w.r.t. clocks

### Impact on the API

A lot of the API can now be pushed down to the track level. In particular, all audio/video operations should be defined on tracks and lifted to generic sources for initial simplicity and backward compatibility.

The full API lifting will be addressed in a follow-up PR but here are some examples included in this PR:
```
# Rewrite metadata on the fly using a function.
# @category Source / Track processing
# @argsof track.metadata.map
def source.metadata.map(%argsof(track.metadata.map), f, (s:source)) =
  tracks = source.tracks(s)
  source(tracks.{metadata = track.metadata.map(%argsof(track.metadata.map), f, tracks.metadata)})
end

# Pan a stereo sound.
# @category Source / Audio processing
# @argsof audio.stereo.pan
def stereo.pan(~id=null(), %argsof(audio.stereo.pan), (s:source)) =
   tracks = source.tracks(s)
   source(id=id, tracks.{audio=audio.stereo.pan(%argsof(audio.stereo.pan), tracks.audio)})
end
```

The convention is to keep top-level namespace for source and to add track-specific operators inside the `track` module, e.g. `track.metadata.map` etc.

Likewise, the API for ffmpeg inline encoders and decoders will be moved to a track-specific API. This will be addressed in a follow-up PR as well.

Finally, ffmpeg filters are now operating directly on tracks. This is not backward compatible but makes much more sense and should be fairly easy to adapt to.

And, of course, the old `drop_*` and `mux_*` operators are now deprecated!

### Technical notes and TODO

TODOs:
- FFmpeg inline encoders and decoders
- Finish converting the rest of the API
- Write doc and examples, cleanup categories

Technical notes:
* This PR harnesses the `eval_check` introduced by @smimram and removes the ad-hoc operator typecheck. This is a much better and regular solution. It wasn't working because the check was not done early enough.
* Implementing the  new muxer's `get_frame` was the real trick. Once again, we are in dire need to revisit our streaming model, which is planed to be done for the next major cycle, along with a switch to OCaml 5!


toots:
Another fun one:
```
def flanger_highpass(s) =
  def mkfilter(graph) =
    s = ffmpeg.filter.audio.input(graph, s)
    s = ffmpeg.filter.flanger(graph, s, delay=10.)
    s = ffmpeg.filter.highpass(graph, s, frequency=4000.)
    ffmpeg.filter.audio.output(graph, s)
  end

  ffmpeg.filter.create(mkfilter)
end

movie = single("/tmp/orig.mkv")
tracks = movie.tracks()

# Add a second track with the movie's audio filtered
# We currently cannot re-use the same source due to clock
# restrictions that should be revisited shortly.
movie = single("/tmp/orig.mkv")
audio_2 = flanger_highpass(movie.tracks().audio)

s = source(id="muxed", tracks.{audio_2=audio_2})

enc = %ffmpeg(
  format="mkv",
  %audio.raw(codec="aac"),
  %audio_2.raw(codec="aac"),
  %video.copy
)

output.file(fallible=true, enc, "/tmp/bla.mkv", s)
```

PR #2765, opened 2022-11-30
Fix some never stuff.

toots:
```
# Make sure that a function that takes a record of the
# type { foo?: int} can take a record of the type: {foo?: never}
def f(x) =
   (x?.foo ?? 1) + 2
end
x = {foo = 123}
let {foo = _, ...y} = x
test.equals(f(y), 3)
```


toots:
I'm merging this, pretty sure it's fine. Let me know if you have any objection @smimram

Issue #2773, opened 2022-12-04
Bug with encoder type inference


toots:
```
def g() =
  s = single("foo")

  s = s

  enc = %ffmpeg(
    format="mkv",
    %audio(codec="aac"),
    %video(codec="aac")
  )

  output.file(fallible=true, enc, "/tmp/bla.mkv", s)

  (s, enc)
end
```
Gives:
```
g :
  () ->
  source('a) * format(audio=pcm(stereo), video=canvas) = <fun>
```

While:
```
def g() =
  s = single("foo")

  #s = s

  enc = %ffmpeg(
    format="mkv",
    %audio(codec="aac"),
    %video(codec="aac")
  )

  output.file(fallible=true, enc, "/tmp/bla.mkv", s)

  (s, enc)
end
```
Gives, as expected:
```
g :
  () ->
  source(audio=pcm(stereo), video=canvas) * format(audio=pcm(stereo), video=canvas) = <fun>
```

PR #2782, opened 2022-12-08
Generalize pattern types at the same level as current code

toots:
Fixes: #2773


smimram (approved):
Seems fine to me, thanks for spotting this!!

PR #2784, opened 2022-12-09
Make sure base type on a sup is always a fresh variable.

toots:
I have seen cases where the `sup` of a two types still contains an original universal variable that then clashes during type unification between the two types and their sup.

This is the corresponding test:
```ocaml
  (* 'a *)
  let a_base = Lang.univ_t () in

  (* 'a.{gni:int} *)
  let a = Type.meth "gni" ([], Lang.int_t) a_base in

  (* 'a.{gni:int, foo?:int} *)
  let b = Type.meth ~optional:false "gni" ([], Lang.int_t) a_base in
  let b = Type.meth ~optional:true "foo" ([], Lang.int_t) b in

  (* sup 'a.{gni:int} 'a.{gni:int, foo?:int} *)
  let s = Typing.sup ~pos:a.Type.pos a b in
  Typing.(a <: s);
  Typing.(b <: s)
```

This PR simply replaces the root type of a `sup` with a fresh universal variable when it is already a universal variable. Since sups are supposed to be unified with their two lower parts afterward, this should be safe to do.

### DIFF
--- a/src/lang/typing.ml
+++ b/src/lang/typing.ml
@@ -199,7 +199,14 @@ exception Incompatible
     function so that it should always be followed by a subtyping. *)
 let rec sup ~pos a b =
   (* Printf.printf "  sup: %s \\/ %s\n%!" (Type.to_string a) (Type.to_string b); *)
-  let sup = sup ~pos in
+  let sup a b =
+    let s = sup ~pos a b in
+    match (demeth s).descr with
+      | Var _ ->
+          let v = Type.var () in
+          remeth s v
+      | _ -> s
+  in
   let mk descr = Type.make ?pos descr in
   let scheme_sup t t' =
     match (t, t') with ([], t), ([], t') -> ([], sup t t') | _ -> t'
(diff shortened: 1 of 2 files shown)


smimram:
To be honnest I am not sure this is the right way to fix the issue here. If I understand correctly, the problem comes from the fact that when you are creating the type `b` you are adding a discardable `foo` field to `'a` and thus we should add the constraint that `'a` cannot contain a field `foo`. Am I correct? If this is the case, may be this is time to introduce forbidden field constraints to universal variables...

toots:
> To be honnest I am not sure this is the right way to fix the issue here. If I understand correctly, the problem comes from the fact that when you are creating the type `b` you are adding a discardable `foo` field to `'a` and thus we should add the constraint that `'a` cannot contain a field `foo`. Am I correct? If this is the case, may be this is time to introduce forbidden field constraints to universal variables...

I'm not sure where it comes from either. The situation emerged from some debugging I did during the `tracks` development. Let me see if it still happens on the final branch.

toots:
Yeah not happening anymore. Thanks for looking into it tho!

PR #2789, opened 2022-12-15
Add support for canvas video output.

smimram:
Use [canvas](https://ocamlpro.github.io/ocaml-canvas/) to show videos.


toots:
I've updated this to compile with the latest `main`. The `macos` API requires the main window to be modified in the main thread, which seems pretty impractical given our setup.

PR #2792, opened 2022-12-17
Add multiline comments.

toots:
This is a proposal for multiline comments. Multiline comments are started with:
```
#<
```

and ended with:

```
>#
```

This mimicks the `/*` and `*/` for C/javascript but allows some nice additional ASCII art

Syntax is intended to allow several nice combinations such as:
```
#<------#
This is
a multiline comment
#------>#

#<---------------#
# Yet another
# form of multiline comment
#---------------->#

   #<
multiline comment can start and end
in the middle of a line
   >#
   
      #<
    #<------
        Also nested comments are cool
    ---->#
   >#

#<-----#
Anything following them
is executed
#---->#x = 1

#<-----#y = x + 2.
Anything after the mark
is comment
#---->#
```

### DIFF
--- a/doc/content/language.md
+++ b/doc/content/language.md
@@ -1343,6 +1343,7 @@ Simple multiline comments:
 ```
 
 Multiline comments can be nested:
+
 ```liquidsoap
 #<
 This is a top-level comment
@@ -1351,11 +1352,12 @@ This is a top-level comment
 
   #<
     This is a nested code comment
-  ># 
+  >#
 >#
 ```
 
 Fancy looking multiline comment
+
 ```liquidsoap
 #<------- BEGIN CODE COMMENT ----#
 Comments can also look like this
--- a/doc/content/language.md
+++ b/doc/content/language.md
@@ -1325,3 +1325,33 @@ output.icecast(%mp3, host="localhost", port=8000,
 ```
 
 so that passwords are not shown in the main script.
+
+### Code comments
+
+Comments can be added to your code in two ways:
+
+_Multi-line comments_ are comments that can span multiple lines. They are delimitated
+by the sequence of characters `#<` at the beginning and `>#` at the end. Anything
+in between those two sequences is considered code comment. Nested code comments
+are also accepted:
+
+```liquidsoap
+#<------- BEGIN CODE COMMENT ----#
+    This is a comment
+
+# This is also a comment
+
+  #< This is a nested code comment >#
+
+#--------- END CODE COMMENT ----->#
+```
+
+_Single-line comments_ are comments that are limited to the current line. Such comments
+are started with the character `#` without a following `<`. Anything after the initial
+`#` character and until the end of the line is considered code comment:
+
+```liquidsoap
+def f(x) = # This is a single line comment.
+  123
+end
+```
--- a/doc/content/language.md
+++ b/doc/content/language.md
@@ -1332,17 +1332,33 @@ Comments can be added to your code in two ways:
 
 _Multi-line comments_ are comments that can span multiple lines. They are delimitated
 by the sequence of characters `#<` at the beginning and `>#` at the end. Anything
-in between those two sequences is considered code comment. Nested code comments
-are also accepted:
+in between those two sequences is considered code comment.
+
+Here are some examples:
+
+Simple multiline comments:
 
 ```liquidsoap
+#< This is a comment >#
+```
+
+Multiline comments can be nested:
+```liquidsoap
+#<
+This is a top-level comment
+
+  # This is also a comment
+
+  #<
+    This is a nested code comment
+  ># 
+>#
+```
+
+Fancy looking multiline comment
+```liquidsoap
 #<------- BEGIN CODE COMMENT ----#
-    This is a comment
-
-# This is also a comment
-
-  #< This is a nested code comment >#
-
+Comments can also look like this
 #--------- END CODE COMMENT ----->#
 ```
 
--- a/tests/language/comments.liq
+++ b/tests/language/comments.liq
@@ -1,33 +1,44 @@
-#------#
+#<------#
 This is
 a multiline comment
-#------#
+with > some # of
+these characters in it.
+#------>#
 
-#---------------#
+#<---------------
 # Yet another
 # form of multiline comment
-#----------------#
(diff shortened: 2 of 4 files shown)


toots:
I changed my mind 😅 See new proposal above!

smimram:
I am not 100% convinced (but close):
1. Why use yet another character (`-`) for comments? We could use something like `####*` (at least three `#`) as marker for multiline comments.
2. There is no distinction between starting and ending markers, which means that we cannot easily comment a region which contains as multiline comment. We should have a way to do this.

toots:
Coming back to this, nested comments do not work in C: 

```c
int x = 123;

/*
this is a comment

  /*
    This is a nested comment
  */

comment continued
*/
```

Not in javascript:

```javascript
const x = 123;

/*
this is a comment

  /*
    This is a nested comment
  */

comment continued
*/
```

But work in OCaml:

```ocaml
let x = 123

(*
this is a comment

  (*
    This is a nested comment
  *)

comment continued
*)
```

So that's not universal but nice for sure!

I changed the proposal, what you think of it now?

toots:
> I am not 100% convinced (but close):
> 
> 1. Why use yet another character (`-`) for comments? We could use something like `####*` (at least three `#`) as marker for multiline comments.
> 2. There is no distinction between starting and ending markers, which means that we cannot easily comment a region which contains as multiline comment. We should have a way to do this.

This has been partially addressed!

vitoyucepi:
This form is used in tests seemingly everywhere instead of `#<`.
```ruby
#<------#
#------>#
```

Ruby highlighting.
```ruby
#<------#
This is
a multiline comment
#------>#
```

Multiline comments for file header.
```ruby
#<
Liquidsoap, a programmable audio stream generator.
Copyright 2003-2023 Savonet team
>#
```

toots:
> This form is used in tests seemingly everywhere instead of `#<`.
> 
> ```ruby
> #<------#
> #------>#
> ```
> 
> Ruby highlighting.
> 
> ```ruby
> #<------#
> This is
> a multiline comment
> #------>#
> ```
> 
> Multiline comments for file header.
> 
> ```ruby
> #<
> Liquidsoap, a programmable audio stream generator.
> Copyright 2003-2023 Savonet team
> >#
> ```

Yeah, I would like the format to be somewhat loose so you can decorate it as you see fit but, the mnimal markers are indeed `#<` and `>#`.

Do you like the proposal overall?

vitoyucepi:
`#<` may be misinterpreted by a code reader as a single-line comment because of leading `#`.

I would like to see additional tests with a comment inside a string and a regexp.
Maybe this PR should add a section about the comments to the documentation.

toots:
Okay, doc added. I think I'll merge this soon.

doc/content/language.md
```diff
@@ -1325,3 +1325,33 @@ output.icecast(%mp3, host="localhost", port=8000,
...
+### Code comments
+
+Comments can be added to your code in two ways:
+
+_Multi-line comments_ are comments that can span multiple lines. They are delimitated
+by the sequence of characters `#<` at the beginning and `>#` at the end. Anything
+in between those two sequences is considered code comment. Nested code comments
+are also accepted:
+
+```liquidsoap
```
vitoyucepi, line 1338: Maybe the simple syntax for multiline comments should be shown first?
````
Simple multiline comments
```liquidsoap
#<
  This is a comment
>#
```
Multiline comments could be nested
```liquidsoap
#<
  This is a top-level comment
  # This is also a comment
  #<
    This is a nested code comment
  ># 
>#
```
Fancy looking multiline comment
```liquidsoap
#<------- BEGIN CODE COMMENT ----#
Comments can also look like this
#--------- END CODE COMMENT ----->#
```
````

File: doc/content/language.md [hunk 2]
```diff
@@ -1325,3 +1325,33 @@ output.icecast(%mp3, host="localhost", port=8000,
...
+#<------- BEGIN CODE COMMENT ----#
+    This is a comment
+
+# This is also a comment
+
+  #< This is a nested code comment >#
+
+#--------- END CODE COMMENT ----->#
+```
```
vitoyucepi, line 1347: Maybe the simple syntax for multiline comments should be shown first?
````
Simple multiline comments
```liquidsoap
#<
  This is a comment
>#
```
Multiline comments could be nested
```liquidsoap
#<
  This is a top-level comment
  # This is also a comment
  #<
    This is a nested code comment
  ># 
>#
```
Fancy looking multiline comment
```liquidsoap
#<------- BEGIN CODE COMMENT ----#
Comments can also look like this
#--------- END CODE COMMENT ----->#
```
````

File: doc/content/language.md [hunk 3]
```diff
@@ -1325,3 +1325,33 @@ output.icecast(%mp3, host="localhost", port=8000,
...
+
+_Single-line comments_ are comments that are limited to the current line. Such comments
+are started with the character `#` without a following `<`. Anything after the initial
+`#` character and until the end of the line is considered code comment:
```
vitoyucepi, line 1351: Single-line comments after multiline comments?

tests/language/comments.liq
```diff
@@ -0,0 +1,70 @@
...
+"
+ignore(s)
+
+r = r/foo
+#<-----
+this is a comment inside a regexp,
+it is not a code comment!
+#---->
+bar/
+ignore(r)
```
vitoyucepi, line 52: `#---->`
This will trigger not terminated comment error?


I'd like to see these tests also
- `"a#b"` should be literally `a#b`;
- `r/a#b/i` should create a regex `r/a#b/i`;
- `"a#<b>#c"` should be literally `a#<b>#c`;
- `r/a#<b>#c/i` should create a regex `r/a#<b>#c/i`.

doc/content/language.md
toots: It seems more logical this way. How would you describe the single line comment syntax without mentioning multi-line comments?

vitoyucepi: Previously in multi-line comments.
> `# This is also a comment`

vitoyucepi:
And the highlighting is broken.

toots:
> And the highlighting is broken. 

That was to be expected unfortunately. Let's see what we can do..

PR #2803, opened 2022-12-23
Add typeof syntactic sugar

toots:
This PR adds the following syntax:
```ruby
(<expr1>:typeof <expr2>)
```
It is is a syntactic sugar for:
```ruby
if true then <expr1> else <expr2> end
```

TODO:
- [ ] Add doc about typecast


toots:
Closing this. I think that this would be a nice feature but it blurs the line between type-checking and evaluation and I don't think that we are ready for it just yet.

PR #2805, opened 2022-12-25
Improve `if .. then .. else .. end` type error

toots:
This PR improves on `if .. then .. else .. end` typechecking errors by typechecking directly the returned value of each branch instead of the function `() -> <value>`.


smimram (approved):
This is mostly fine, even though this might indicate that we'd rather have conditional branching as a basic construction of the language instead of a library-defined one + typing hacks?

With your solution, we might have problem if someone re-defines ìf`, although I don't expect that this is going to be the case...

toots:
> This is mostly fine, even though this might indicate that we'd rather have conditional branching as a basic construction of the language instead of a library-defined one + typing hacks?
> 
> With your solution, we might have problem if someone re-defines ìf`, although I don't expect that this is going to be the case...

You cannot re-define if I believe because it is a keyword.

PR #2806, opened 2022-12-27
Tighten up source subtyping

toots:
### Summary

We want to support the following subtyping:
```
source(audio=pcm(stereo)) <: source(?audio=pcm(stereo))
source() <: source(?audio=pcm(stereo))
```

### Background

In the current type implementation, when computing types such as the type for `l` in:
```
l = [123.{foo = 123, bla = "aabb"}, null(123).{ bla = "blo", gni = 123.}]
```
Ideally, the type for `l` should be: `[int?.{bla:int}]`

In the type implementation, this mechanism kicks in because the type for `l` is considered as `Covariant`, i.e. the type for `l` could be any type `t` such as for all elements in the list `el`, `typeof el <: t`.

Internally, this is denoted by a variance annotation in type links, i.e., the type for `l` (noted `l.t`) is:  `l.t = Link (Covariant, <currently computed type>)` so, when typechecking `l.t` against each elements in the list, we know that we can re-adjust it by changing this link to point to a greater type (meaning if old type is `t`, new type is `t'` such as: `t <: t'`). 

For instance, `l.t` starts as `'a` then, when typechecking against `123.{foo = 123, bla = "aabb"}`, it becomes `int.{foo: int, bla: string}` then, when typechecking against `null(123).{ bla = "blo", gni = 123.}`, it becomes `int?.{bla: string}` etc.

However, for constructed types such as `source`, `format` (and `ref`), we do not want this mechanism to kick in. For instance, in an encoded output of type:
```
output.encoded: (format('a), source('a)) -> ...
```
If we pass the following:
```
output.encoded(%vorbis(stereo), mean(blank()))
```
We want to make sure that this results in a typing error.

Internally, `%vorbis(stereo)` is of type: `format({audio = pcm(stereo})` and `mean(blank())` of type: `source('a.{audio: pcm(mono)})`

If the `sup` mechanism described above was to kick in, we would have no problem saying that the `sup` of `{audio = pcm(stereo}` and `'a.{audio: pcm(mono)}` is `{}` (aka `unit`) and, so, the function would happily accept: `format({})` and `source({})`.

For this reason, constructed types such as these are associated with an `Invariant` type annotation, i.e.:
```
Const { constructor = "source"; params = [(`Invariant, t)] }
```

### Changes

When type checking `Invariant` constructed type, the current implementation does:
```
t <: t'
t' <: t
```
Which doesn't work in the initial example above.

Morally, however, what we really want here is to prevent the type `'a` in `source('a)` from being adjusted via the `sup` mechanism as described above.

Thus this PR relaxes the implementation of `Invariant` constructed types as follows:
* When checking `source('a) <: source('b)`, only `'a <: 'b` is performed
* All type bindings in `'b` prior and after typechecking are changed to `Invariant` bindings

This ensures that the `sup` mechanism never kicks in.

### Notes

This PR also contains a fair bit of cleanup of code in `core` that wasn't actually used. It also removes the notion of `Contravariant` constructed type and type links. We currently do not handle contravariant anywhere and so it is confusing to have.


src/core/lang.ml
```diff
@@ -13,11 +13,10 @@ let add_protocol ~syntax ~doc ~static name resolver =
 let frame_t base_type fields = Frame_type.make base_type fields
 let internal_t () = Frame_type.internal ()
 
-(** Type of audio formats that can encode frame of a given kind. *)
-let format_t k =
+let format_t t =
   Type.make
     (Type.Constr
-       { Type.constructor = "format"; Type.params = [(`Covariant, k)] })
+       { Type.constructor = "format"; Type.params = [(`Invariant, t)] })
```
smimram, line 19:
```suggestion
       (* The type has to be invariant because we don't want the sup mechanism to be used here, see #2806. *)
       { Type.constructor = "format"; Type.params = [(`Invariant, t)] })
```

smimram (approved):
Looks fine to me!

Issue #2819, opened 2023-01-03
output.shoutcast(icy_id) is ignored always connects to sid#1


TheRealMattLear:
**Describe the bug**
I'm seeing in 2.1.2 that `output.shoutcast(icy_id=x)` does not appear to be applying the `icy_id` correctly. It just appears to connect to SID "1" no matter what. Rolling back to 2.0.6 seems to solve the problem. I've literally just checked out and compiled 2.1.3 and seeing the same problem

**To Reproduce**
```
settings.init.allow_root.set(true);
settings.log.stdout.set(true);
settings.log.level.set(4);

s = single("/root/test.mp3")

output.shoutcast(
  %mp3(samplerate=44100,bitrate=128,id3v2=true,stereo=true),
  icy_id = 2,  host = "127.0.0.1",  port = 8028,  password = "ZM4NbmBDDvX6",
  encoding = "UTF-8",  name = "test",   url = "www.mywebsite.com", genre = "hip hop",
  public = true, icy_metadata = "true",
  s
);

```

Running liquidsoap shows on the last line ```Connecting mount sid#1 for @127.0.0.1```

```
root@bullseye:~# liquidsoap ./shoutcast.liq 
2023/01/03 02:00:26 >>> LOG START
2023/01/03 02:00:25 [main:3] Liquidsoap 2.1.3
2023/01/03 02:00:25 [main:3] Using: bytes=[distributed with OCaml 4.02 or above] pcre=7.5.0 sedlex=3.0 menhirLib=20220210 curl=0.9.2 uri=4.2.0 dtools=0.4.4 duppy=0.9.2 cry=0.6.7 mm=0.8.1 ogg=0.7.3 ogg.decoder=0.7.3 vorbis=0.8.0 vorbis.decoder=0.8.0 mad=0.5.2 f
lac=0.3.1 flac.ogg=0.3.1 flac.decoder=0.3.1 dynlink=[distributed with Ocaml] lame=0.3.6 fdkaac=0.3.2 theora=0.4.0 theora.decoder=0.4.0 ffmpeg=1.1.6 taglib=0.3.10 camomile=1.0.2 faad=0.5.1
2023/01/03 02:00:25 [clock:3] Using builtin (low-precision) implementation for latency control
2023/01/03 02:00:26 [frame:4] frame.audio.channels set to: 2
2023/01/03 02:00:26 [frame:4] frame.video.default set to: false
[...]

PR #2823, opened 2023-01-03
Detect functions defining the same label multiple times.

smimram:
This avoids situations such as #2819.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -21,6 +21,7 @@ New:
 - Added syntactic sugar for record spread: `let {foo, gni, ..y} = x`
   and `y = { foo = 123, gni = "aabb", ...x}` (#2737)
 - Added `file.{copy, move}` (#2771)
+- Detect functions defining multiple arguments with the same label (#2823).
 
 Changed:
 
--- a/src/lang/runtime.ml
+++ b/src/lang/runtime.ml
@@ -125,6 +125,11 @@ let throw ?(formatter = Format.std_formatter) lexbuf =
         (if lbl = "" then "unlabeled argument"
         else Format.sprintf "argument labeled %S" lbl);
       raise Error
+  | Term.Duplicate_label (pos, lbl) ->
+      error_header ~formatter 6 pos;
+      Format.fprintf formatter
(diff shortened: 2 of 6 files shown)


toots (approved):
Looks awesome thanks for doing this!

vitoyucepi:
Hi @toots,
There are also function annotations
https://github.com/savonet/liquidsoap/blob/583135b54486c88f557aa94cabfeb0f86a9d3301/libs/shoutcast.liq#L6
https://github.com/savonet/liquidsoap/blob/f6ac06a5597898544bd44faf470f7344ebce9ab3/src/libs/shoutcast.liq#L6

How is website documentation generated?
https://www.liquidsoap.info/doc-dev/reference.html#output.shoutcast
https://www.liquidsoap.info/doc-2.1.3/reference.html#output.shoutcast

toots:
> Hi @toots, There are also function annotations
> 
> https://github.com/savonet/liquidsoap/blob/583135b54486c88f557aa94cabfeb0f86a9d3301/libs/shoutcast.liq#L6
> 
> 
> https://github.com/savonet/liquidsoap/blob/f6ac06a5597898544bd44faf470f7344ebce9ab3/src/libs/shoutcast.liq#L6
> 
> How is website documentation generated? https://www.liquidsoap.info/doc-dev/reference.html#output.shoutcast https://www.liquidsoap.info/doc-2.1.3/reference.html#output.shoutcast

Thanks for pointing out, this is fixed now.

The doc is generated during the update webasite workflow that runs on each commit on the `main` branch: https://github.com/savonet/liquidsoap/actions/workflows/update-website.yml

The doc for released version is static and checked out in the https://github.com/savonet/liquidsoap-full repository (that will need to be broken up one day!).

vitoyucepi:
> The doc is generated during the update webasite workflow that runs on each commit on the main branch: https://github.com/savonet/liquidsoap/actions/workflows/update-website.yml

Oops, I'm sorry. 
In my question, I want to point out that annotations could be inconsistent with the function definition.
BTW the documentation for 2.1.3 should be left as is.

PR #2879, opened 2023-02-05
Support concurrent computation of sources.

toots:
This PR enables a first support for concurrent encoding when compiled with OCaml 5. It is using `domainsLib` for parallelization and `Mutex` to prevent concurrent computations of sources `#get`, which is not thread-safe.

It seems that most of the lock-free logic in `domainsLib` and its ecosystem is based on active wait and short-term computation units. In a case like ours, it would seem that lifting `ocaml-duppy` to support domains and spread its queue threads to various domains would be a better fit to benefit from its passive, select-base waiting/asynchronous nature.


smimram (approved):
This is too cool! :)

src/core/source.ml
```diff
@@ -667,57 +667,58 @@ class virtual operator ?(name = "src") sources =
...
-            self#log#severe
-              "#get_frame returned a buffer without enough breaks!";
-            assert false))
-      else (
-        let memo = self#memo in
-        try Frame.get_chunk buf memo
-        with Frame.No_chunk ->
-          if not self#is_ready then silent_end_track ()
+    method get =
+      self#mutexify (fun buf ->
```
toots, line 671: Mutexify should be optional.

smimram:
First general comment, we are still missing
- changelog entry
- many **tests**!!!

doc/content/clocks.md
```diff
@@ -310,3 +312,11 @@ end
 one()
 one()
 ```
+
+### Case 2: post OCaml 5
```
smimram, line 316: I'd rather have this as *case 1*, since I expect that our users will quickly switch to OCaml 5.

smimram (commented):
Looks great, thanks!

We're still missing changelog / tests.

doc/content/clocks.md
```diff
@@ -310,3 +312,11 @@ end
...
+
+If you compile `liquidsoap` with `domainslib`, concurrent encoding should be enabled by default!
```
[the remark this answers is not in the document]
toots: `domainslib` is not available pre-ocaml 5 I feel like this is simpler to just say you need `domainslib`..

src/core/multicore.domainslib.ml
```diff
@@ -0,0 +1,64 @@
...
+
+  You should have received a copy of the GNU General Public License
+  along with this program; if not, write to the Free Software
+  Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301  USA
+
+ *****************************************************************************)
+
+let name = "domainsLib"
+
+let conf_multicore =
```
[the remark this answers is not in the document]
toots: I think that this is the point. If your script fails with this setting you're running it with the wrong version of the interpreter.

doc/content/clocks.md
toots: Are they? I wasn't planning on shipping binary builds using OCaml 5 just yet frankly.

toots:
I'm afraid this is gonna need some more involved work. First thing that will need to be cleaned is the use of lazy references, which causes all sort of race conditions with multicore ocaml. Closing for now but keeping the branch around.

toots:
I'm afraid this is gonna need some more involved work. First thing that will need to be cleaned is the use of lazy references, which causes all sort of race conditions with multicore ocaml. Closing for now but keeping the branch around.

PR #2881, opened 2023-02-05
Implement references as objects.

smimram:
Implement references as objects: `ref('a)` is now `(() -> 'a).{set : ('a) -> unit}`. For instance,
```
x = ref(5)
x.set(x() + 2)
x := x() + 2
```
(the `:=` syntax is still supported for now)

### DIFF
--- a/src/core/lang_source.ml
+++ b/src/core/lang_source.ml
@@ -343,8 +343,6 @@ let check_content v t =
             check_value v t
         (* The type says that we should drop the method. *)
         | Value.Meth (_, _, v), _ -> check_value v t
-        | Ref r, Type.Constr { Type.constructor = "ref"; params = [(_, t)] } ->
-            check_value (Atomic.get r) t
         (* We don't check functions, assuming anything creating a source is a
            FFI registered via add_operator so the check will happen there. *)
         | Fun _, _ | FFI _, _ -> ()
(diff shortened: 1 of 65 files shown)


smimram:
I find the syntax proposed above quite heavy actually. Another option would be to implement `ref('a)` as
```
(() -> 'a).{set : ('a) -> unit
```
For instance,
```
x = ref(5)
x.set(x() + 2)
```
This would have in particular the advantage that any reference is directly a valid getter!

One of the aims was to remove the `!` notation on the long term. I think that this would be 

What do you think?

toots:
I think that:

```
(() -> 'a).{set : ('a) -> unit
```
should be the way to go, this is how the settings are already implemented!

How about type safety? In OCaml we have this:
```ocaml
# let f =
    let x = ref [] in
    fun v -> x := v::!x;;
val f : '_a -> unit = <fun>
```
Where the type `'_a` can. only be set once and for all.

smimram:
The _value restriction_ seems to work here (= `'_a` in OCaml): you cand test it with `test/failures/ref.liq`.

src/core/lang_source.ml
```diff
@@ -542,6 +542,7 @@ let iter_sources ?on_reference ~static_analysis_failed f v =
             List.iter (function _, _, Some v -> iter_value v | _ -> ()) proto
         | FFI (proto, _) ->
             List.iter (function _, _, Some v -> iter_value v | _ -> ()) proto
```
toots, line 544: I think we should simply call `on_imprecise` when reaching a `FFI`. We can't check the function's body in this case so this makes sense and will suit our case here.

toots (approved):
Looking great just one remark.

PR #2910, opened 2023-02-18
Fix copy-pasted functions description in pulseaudio_io.ml

vitoyucepi:
The description for some functions in pulseaudio was copy-pasted from portaudio module.


smimram:
Thanks!

Issue #2809, opened 2022-12-28
http.post.file use different variable for timeout and different defaults from the others http.post params


cdgraff:
Inside `http.post` and `http.post.stream`  variable is called `timeout_ms` and has 10000 as the default

`timeout_ms (of type int?, which defaults to 10000): Timeout for network operations in milliseconds.`

Inside`http.post.file` variable is called `timeout` and is NULL as default

`timeout (of type int?, which defaults to null): Timeout in ms.`

Will be good normalize to easy and clean writing

thanks!

toots:
Hi,

That's a great suggestion! Have you ever tried to contribute to the code? This would be an easy one to pick!

PR #2920, opened 2023-02-24
Turn all timeout into float values in seconds

toots:
This is a bit of a breaking change but I think that it would prove valuable: require that all timeout settings and parameters be labelled `timeout` (or `read_timeout` etc) and hold a floating point value in seconds.

Fixes: #2809


smimram:
NB : it won't be a breaking change if you merge #2887 :)

toots:
Well the change from milliseconds to seconds is gonna make things break still..

smimram (approved):

Issue #2906, opened 2023-02-17
long time for start LiquidSoap - 20 second 


HDSVR:
Hello im install LiquidSoap 2.1.3 from OCAML, hardware nano pi neo 512 (Armbian arm, allwinner h3 cortex a7) 
Create simple script play one folder to Alsa sound and start script .  long time for start playing 
2023/02/17 09:37:01 [clock:3] Using builtin (low-precision) implementation for late                                                                                                                                                          ncy control 21 second load this module .....   it/s very loong ((  How fix this time

LiquidSoap 1.4.3 load my simple script playing 2 second 

LOGG

2023/02/17 09:37:01 [main:3] Liquidsoap 2.1.3
init(0):init-start                         @start-triggers
init(0):init-start                         @stop-triggers
2023/02/17 09:37:01 [main:3] Using: bytes=[distributed with OCaml 4.02 or above] pc                                                                                                                                                          re=7.5.0 sedlex=3.0 menhirLib=20220210 curl=0.9.2 uri=4.2.0 dtools=0.4.4 duppy=0.9.                                                                                                                                                          2 cry=0.6.7 mm=0.8.1 mad=0.5.2 dynlink=[distributed with Ocaml] lame=0.3.6 fdkaac=0                                                                                                                                                          .3.3 ffmpeg=1.1.7 alsa=0.3.0 taglib=0.3.10 camomile=1.0.2
init(0):init-start                         @stop
2023/02/17 09:37:01 [clock:3] Using builtin (low-precision) implementation for late                                                                                                                                                          ncy control
init(0):init-start                         @return
2023/02/17 09:37:22 [frame:3] Using 44100Hz audio, 25Hz video, 44100Hz main.
init(0):init-start                         @stop-depends
2023/02/17 09:37:22 [frame:3] Video frame size set to: 1280x720
init(0):init-start                         @start-atom

Many Sanks

toots:
Hi @HDSVR and thanks for reporting. We are aware of the issue. We have been shifting a lot of our internal operators to the standard script library and also expanded it a lot. This is giving the software a lot of flexibility but also means that the runtime cost of typechecking all this script code has increased.

We are planning to tackle this with first a cached of the resolved typechecks and then as time allows an optimization of the abstract syntactic tree so that only the code you effectively use in your script is loaded and executed.

We will keep you posted of our progress and close this issue when the launch time has been reduced back to an acceptable one.

[...]

PR #2951, opened 2023-03-09
Convert term methods to actual map.

toots:
We have previously discarded some memory and startup time optimization to get `2.2.x` out the door but, with recent reports about [start time](https://github.com/savonet/liquidsoap/issues/2906) and [memory usage](https://github.com/savonet/liquidsoap/issues/2906) I feel like the need to address that before a release is rising.

I suggest that we aim toward a dead code elimination solution. If your script is just doing `output.dummy(blank())`, we should:
* Parse all the standard library
* Eliminate all that's not used (pretty much everything in this example)
* Typecheck the remaining code and run it.

This should solve a lot of our issues regarding startup time and memory/CPU utilisation while avoiding more involved optimizations such as caching parsing and typechecking (which could still be done later).

Dead code elimination should be optinal and disabled in the `./liquidsoap` script (for quick local testing) as well as in all the API/settings/.. export functions (duh!).

This PR is the first step in this direction, making term methods actual mappings attached to terms so that can be later properly filtered during dead code elimination. Same should follow for values later.


smimram:
Thanks for this!

Quick preliminary comments:

1. This is clearly a better way to implement methods. I had thought of this but never did it because it's much easier to forget about some methods during evaluation. If we go this way, we should make as many assertions as possible in order make sure that we don't silently drop methods, etc.
2. The same should be done for types (and right now the typing phase is not neglectible): have a `methods` field in types with all the types of the methods as a map.
3. We should make an effort to **quantify improvements**: could you please provide tests so that we can observe effects both on execution speed and memory?

toots:
> Thanks for this!
> 
> Quick preliminary comments:
> 
> 1. This is clearly a better way to implement methods. I had thought of this but never did it because it's much easier to forget about some methods during evaluation. If we go this way, we should make as many assertions as possible in order make sure that we don't silently drop methods, etc.

Methods are immutable maps so I don't think that this should be an issue.

> 2. The same should be done for types (and right now the typing phase is not neglectible): have a `methods` field in types with all the types of the methods as a map.

I don't think so, types needs to keep the Russian doll structure from type masking, for instance:
```
x = ...

y = x.{foo = "bla"}

y.gni = 123
z = y.foo.split(",")
x.foo + 2
```
We do need to retain the underlying `'a` type for `x` in `y` with all its (possibly) masked methods).

> 3. We should make an effort to **quantify improvements**: could you please provide tests so that we can observe effects both on execution speed and memory?

Definitely, will do next. I'm not expecting a clear improvement here. I'm replacing a nested representation with a immutable map. This is mostly intended to simplify future handling of dead code elimination and etc. but keeping track of these incremental improvements will definitely help.

toots:
Some early performance reports. We get a consistent ~35% decrease in computation time over large records, which was expected since lookup is now prolly close to `O(1)`.

General evaluation performance is a mix bag but that was expected. First, because of the addition of systematic methods operations on eval, we were hitting some stack overflow. Thus, the branch is reworking some list method to avoid that by bringing a limited set of functions back to the core and re-implementing others using error raising, which might not be the most efficient but will avoid ugly stack overflow.

Overall, the evaluation slowdown on erathosteme seems to be in the ~8% ballpark, which isn't too bad.

Next, I'll do some memory measurements, which should also be interesting and, I expect, in favor of the new implementation.

`main`, `4.14.1`, no `flambda`:
```
Computing record with  2 fields took 0.363829851151 seconds.
Computing record with  5 fields took 0.726735115051 seconds.
Computing record with 10 fields took 1.2417280674 seconds.
Computing record with 20 fields took 2.4899828434 seconds.

Computing sum of fields (big record) took 0.460048913956 seconds.

Computing Erathostenes (imperative) took 3.874751091 seconds.
Computing Erathostenes (recursive) took 4.1829829216 seconds.
```

`record-map`, `4.14.1`, no `flambda`:
```
Computing record with  2 fields took 0.400293111801 seconds.
Computing record with  5 fields took 0.792161941528 seconds.
Computing record with 10 fields took 1.39449000359 seconds.
Computing record with 20 fields took 2.65891098976 seconds.

Computing sum of fields (big record) took 0.293876171112 seconds.

Computing Erathostenes (imperative) took 4.12653493881 seconds.
Computing Erathostenes (recursive) took 4.19970202446 seconds.
```

`main`, `4.14.0`,`flambda`:
```
Computing record with  2 fields took 0.31799697876 seconds.
Computing record with  5 fields took 0.670955896378 seconds.
Computing record with 10 fields took 1.2678539753 seconds.
Computing record with 20 fields took 2.47235417366 seconds.

Computing sum of fields (big record) took 0.437700986862 seconds.

Computing Erathostenes (imperative) took 3.67962694168 seconds.
Computing Erathostenes (recursive) took 3.71345496178 seconds.
```

`record-map`, `4.14.0`,`flambda`:
```
Computing record with  2 fields took 0.392857789993 seconds.
Computing record with  5 fields took 0.776586055756 seconds.
Computing record with 10 fields took 1.37337303162 seconds.
Computing record with 20 fields took 2.65152192116 seconds.

Computing sum of fields (big record) took 0.292984008789 seconds.

Computing Erathostenes (imperative) took 3.97981405258 seconds.
Computing Erathostenes (recursive) took 4.02685117722 seconds.
```

toots:
Ok memory consumption report:
`main`:
```
Big record memory consumption: 20.59 MB
Big record term-only memory consumption: 4.08 MB
```

`record-map`:
```
Big record memory consumption: 21.04 MB
Big record term-only memory consumption: 16.38 kB
```

As expected. Also: yeah we do need to optimize type checking.. woof..

toots:
Memory usage before loading the whole standard library:

`main`: `46.15 MB`
`record-map`: `46.35 MB`

Memory usage after loading the whole standard library:

`main`: `188.37 MB`
`record-map`: `187.22 MB`

Not a really significant change here but, again, we should gain a _lot_ when we implement dead code elimination..

toots:
After chatting with @smimram I believe that this PR should be merged and we should actually move all critical `list` functions back to the core.

PR #2953, opened 2023-03-14
Add support for the proprietary stereotool operator.

toots:
This PR adds support for the proprietary `stereotool` processing shared library as disclosed in https://github.com/savonet/liquidsoap/discussions/1782#discussioncomment-4939967

The operator provides a low level `track.audio.stereotool` track operator that processes the audio data and a higher level `stereotool` operator that processes the audio of a source with a `audio` track.

~~The `stereotool` operator also takes care of delaying metadata and track marks to match the delay introduced by the audio processing~~. _Nevermind_  This part of the PR was too involved and was removed. Track marks are still too tricky to allow this. Typical delay I'm seeing is 90ms so that's not too bad either for now. Will reconsider once we're moved to immutable content.

Other tracks are dropped as they would be getting out of sync after applying the processing. It is still possible to add them back but this should be reserved for advanced users that can also potentially use a `ffmpeg` filter to delay the video accordingly and etc.

Doc:

`track.audio.stereotool`:
```Process the given audio track with StereoTool.

Type: (?id : string?, library_file : string, ?license_key : string?,
 ?presets : string?, ?load_type : string, pcm('a)) -> pcm('a)

Category: Track / Audio processing

Arguments:

 * id : string? (default: null)
     Force the value of the track ID.

 * library_file : string
     Path to the shared library file.

 * license_key : string? (default: null)


 * presets : string? (default: null)
     Path to a preset file to load when initializing the operator.

 * load_type : string (default: "totalinit")
     Load type for preset. One of: "totalinit", "all_settings", "audiofm", "audio",
     "processing", "repair", "repair_no_pnr" or "sublevel_pnr".

 * (unlabeled) : pcm('a)


Methods:

 * api_version : () -> int
     API version.

 * latency : () -> float
     Get the operator's latency.

 * software_version : () -> int
     Software version.

 * unlincensed_used_features : () -> string?
     Check if the license is valid for the current settings.

 * valid_license : () -> bool
     Check if the license is valid for the current settings.
```

`stereotool`:
```
Process an audio source using stereotool

Type: (?id : string?, library_file : string, ?license_key : string?,
 ?presets : string?, ?load_type : string, source(audio=pcm('a), 'b)) ->
source(audio=pcm('a))

Category: Source / Audio processing

Arguments:

 * id : string? (default: "stereotool")
     Force the value of the track ID.

 * library_file : string
     Path to the shared library file.

 * license_key : string? (default: null)


 * presets : string? (default: null)
     Path to a preset file to load when initializing the operator.

 * load_type : string (default: "totalinit")
     Load type for preset. One of: "totalinit", "all_settings", "audiofm", "audio",
     "processing", "repair", "repair_no_pnr" or "sublevel_pnr".

 * (unlabeled) : source(audio=pcm('a), 'b)


Methods:
... same as track.audio.stereotool + regular source methods.
```

### DIFF
--- a/dune-project
+++ b/dune-project
@@ -37,6 +37,7 @@
     bjack
     camlimages
     camomile
+    ctypes-foreign
     dssi
     faad
     fdkaac
--- a/liquidsoap.opam
+++ b/liquidsoap.opam
@@ -40,6 +40,7 @@ depopts: [
   "bjack"
   "camlimages"
   "camomile"
+  "ctypes-foreign"
   "dssi"
   "faad"
   "fdkaac"
--- a/src/config/stereotool_option.disabled.ml
+++ b/src/config/stereotool_option.disabled.ml
@@ -1,0 +1,2 @@
+let detected = "no (requires ctypes-foreign)"
+let enabled = false
--- a/src/config/stereotool_option.enabled.ml
+++ b/src/config/stereotool_option.enabled.ml
@@ -1,0 +1,1 @@
+noop.enabled.ml
--- a/src/core/dune
+++ b/src/core/dune
@@ -659,6 +659,14 @@
  (modules ssl_base builtins_http_ssl))
 
 (library
+ (name liquidsoap_stereotool)
+ (libraries stereotool liquidsoap_core)
+ (library_flags -linkall)
+ (wrapped false)
+ (optional)
+ (modules stereotool_op))
+
+(library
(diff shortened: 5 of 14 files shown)


smimram:
Very cool, thanks!

moleculezz:
I'm on x86_64. What files from the Stereo Tool generic plugin do I have to copy to the system?

PR #2973, opened 2023-03-23
Enhance `thread.when`, add CPU usage.

toots:
PR from random work enhancing and testing `thread.when` and adding CPU usage metrics.


smimram (commented):
Looks pretty much like `thread.when`...

toots:
It is a duplicate thanks for catching that. Will report doc there.

PR #3032, opened 2023-04-26
Ffmpeg filter cleanup

toots:


toots:
This is not perfect but a step in the right direction. The next phase I think is to understand how to properly shutdown filters. Liquidsoap assumes a source can shutdown and restart but a filter only gets one final EOF frame..

PR #3036, opened 2023-04-29
Add log colors!

toots:
This should help readability:

Screenshot 2023-04-29 at 10 35 25 AM

Questions:
* Colors for debug?
* Colors for timestamp?
* Colors for source id?

I don't want to turn the logs into a Christmas tree but I feel like these could be useful as well.


toots:
Full on Christmas tree would look like this:

Screenshot 2023-04-29 at 10 54 39 AM

toots:
More with warning example:

Screenshot 2023-04-29 at 11 03 27 AM

smimram:
Oh yes this is much more readable (in particular, it's very nice to be able to see directly important messages...)

PR #3049, opened 2023-05-03
Use synchronized lazy values.

smimram:
This allows being compatible with ocaml >= 5.


toots (approved):
This is great thanks!

Couple of remarks:
* Why not use a private type with lazy as underlying implementation? This would prevent any bogus `Lazy.force`
* Since `Synclazy` is not really needed for the language per se, you could consider extracting it to its own library the way `stereotool` was extracted. This way, we can swap out the implementation later.

smimram:
I have added an `mli` file (this is better than `private` I think). I cannot put syncLazy as a separate module because some public libraries (e.g. `liquidsoap_lang`) depend on it so that it would have to be made a public library too...

toots:
> I have added an `mli` file (this is better than `private` I think). 

Yeah that's what I meant!

> I cannot put syncLazy as a separate module because some public libraries (e.g. `liquidsoap_lang`) depend on it so that it would have to be made a public library too...

Confirmed. I'm surprised it worked with `stereotool`.

Issue #3010, opened 2023-04-17
Impossible to get metadata from FLAC stream


jduffas:
**Describe the bug**
Impossible to get metadata from FLAC source with Liquidsoap 2.x

**To Reproduce**

I use B.U.T.T. to stream a FLAC stream, and get it with input.harbor or input.http on icecast2 server.
I got no metadata, 

my code :
`source_Flac = input.http("http://localhost:8000/source_TEST")

source = mksafe(source_Flac)

output.icecast(%fdkaac(aot="mpeg4_aac_lc", bitrate=128), host = "localhost", port = 8000, password = "xxx", mount = "singsingaac128TEST", name="Radio Sing Sing aac 128 Kb/s", description="Un nouveau souffle pour vos oreilles", url="http://sing-sing-bis.org", genre="eclectic", source)`

source can also be 
`source_Flac = input.harbor("Source_TEST",port=8080,password="xxx")`

no matter what the output format is (I tried aac, mp3, etc.)

**Expected behavior**

**Version details**

- Ubuntu Focal
- Version 2.1.4 or even Dev version of liquidsoap
- icecast2 server version 1.4.4 or 2.4.99.3 (2.5 beta 3)

In the contrary, if I use Liquidsoap version 1.4.3 I got metadata with no problem.

**Install method**
I installed the Ubuntu : liquidsoap_2.1.4-ubuntu-focal-1_amd64.deb
idem with the 2.2.0+git@https://github.com/savonet/liquidsoap/commit/2a3bf62e17face878615366423bab450d9cc017d version.
the icecast is the 2.5 Beta 3 (I got the same result with 2.4.4 version)

**Common issues**
Only FLAC source with 2.x version doesn't get metadata.

toots:
Patch submitted to ffmpeg: http://ffmpeg.org/pipermail/ffmpeg-devel/2023-April/308780.html

Issue #2917, opened 2023-02-22
Metadata missing with ogg in icecast stream


hydrosIII:
I am running liquidsoap to catch a stream from the internet and send it to and fm transmitter. This stream is in ogg. Recently we changed the stream to mp3 format and we started to see metadata in the logs. This metadata were not visible before. But makes the logs semantic, and maybe is possible to pass it to the transmitter. So we want to know if is possible to read the metadata in the ogg format. When we pass this stream through liquidsoap to an icecast stream in mp3 and send it over to a harbor in another liquidsoap the metadata starts appearing again, so the medata is not lost but is not correctly read. 

Liquidsoap version Liquidsoap 2.1.3, installed trough package.
Debian bullseye
kernel 5.10.0-21-rt-amd64
I can confirm that this happens in previous versions of liquidsoap. 

Example of the log in ogg:

````
feb 17 19:00:40 violetaradio liquidsoap[139840]: [vorbis @ 0x7f5484005600] Ignoring comment header
feb 17 19:05:43 violetaradio liquidsoap[139840]: [vorbis @ 0x7f5484005600] Ignoring comment header
feb 17 19:05:44 violetaradio liquidsoap[139840]: [vorbis @ 0x7f5484005600] Ignoring comment header
feb 17 19:05:44 violetaradio liquidsoap[139840]: [vorbis @ 0x7f5484005600] Ignoring comment header
feb 17 19:05:44 violetaradio liquidsoap[139840]: [vorbis @ 0x7f5484005600] Ignoring comment header
feb 17 19:09:17 violetaradio liquidsoap[139840]: [vorbis @ 0x7f5484005600] Ignoring comment header
feb 17 19:11:47 violetaradio liquidsoap[139840]: [vorbis @ 0x7f5484005600] Ignoring comment header
 ````

Example of the log in mp3:
````
feb 16 13:11:22 violetaradio liquidsoap[102189]: 2023/02/16 13:11:22 [buffer.adaptative_producer_0:3] Buffer emptied, start buffering...
feb 16 13:11:22 violetaradio liquidsoap[102189]: 2023/02/16 13:11:22 [mksafe:3] Switch to safe_blank with transition.
feb 16 13:11:25 violetaradio liquidsoap[102189]: [https @ 0x7f3b240035c0] Metadata update for StreamTitle: Identivarias 3 Dos directora>
[...]

PR #3062, opened 2023-05-06
Make ogg muxer compliant with the ffmpeg demuxer (and the spec).

### LINKED ISSUE #2965: Icecast ogg/flac output keeps disconnecting when using SRT input [2.2.0 nightly]
**Describe the bug**
I'm testing out replacing a Harbor input with a SRT input. Everything seems to work well, except that one of my three Icecast outputs disconnects when killing the SRT input and falling back to an audio file. There are three outputs to Icecast: `output.icecast(%ogg(%flac)`, `output.icecast(%mp3(bitrate=192,samplerate=48000)` and `output.icecast(%fdkaac(channels=2, samplerate=48000, bitrate=96, afterburner=true, aot='mpeg4_aac_lc', transmux='adts', sbr_mode=true)`. Only the ogg/flac output disconnects when switching from the SRT input to a static file.

In the logs there is an entry `Error while sending data: could not write data to host: Unix.Unix_error(Unix.EPIPE, "write", "")!`

This happens when transitioning from the SRT source to the fallback source. Not when transitioning from the fallback source to the SRT source.

**To Reproduce**
The script that I run is [found here.](https://github.com/oszuidwest/liquidsoap-ubuntu/blob/srt/radio.liq)

**Expected behavior**
All streams to Icecast to stay connected.

**Version details**

- OS: Ubuntu 22.04
- Version a7eea64_2.2.0

**Install method**
Nightly `.deb`  package

```
root@liq220:/var/log/liquidsoap# tail -f radio.log 
2023/03/19 15:23:29 [ZW_FM:3] Connection setup was successful.
2023/03/19 15:25:44 [decoder:2] Decoding "/var/audio/fallback.ogg" ended: Ffmpeg_decoder.End_of_file.
2023/03/19 15:25:45 [single_0:3] Prepared "/var/audio/fallback.ogg" (RID 0).
2023/03/19 15:28:28 [decoder:2] Decoding "/var/audio/fallback.ogg" ended: Ffmpeg_decoder.End_of_file.
2023/03/19 15:28:28 [single_0:3] Prepared "/var/audio/fallback.ogg" (RID 0).
2023/03/19 15:31:05 [clock.input.srt:3] Delegating synchronisation to active sources
2023/03/19 15:31:05 [mksafe:3] Switch to studio with transition.
2023/03/19 15:31:05 [replay_metadata_16:3] Content type is {audio=pcm(stereo)}.
2023/03/19 15:31:06 [switch_1:3] Switch to blank.strip_0 with transition.
2023/03/19 15:31:06 [replay_metadata_17:3] Content type is {audio=pcm(stereo)}.
2023/03/19 15:31:36 [stderr:3] Error while executing OCaml read callback: Srt.Error(443488021, "Connection was broken")
2023/03/19 15:31:36 [studio:2] Feeding failed: Avutil.Error(Generic error in an external library)
2023/03/19 15:31:36 [mksafe:3] Switch to safe_blank with forgetful transition.
2023/03/19 15:31:36 [clock.input.srt:3] Delegating synchronisation to CPU clock
2023/03/19 15:31:49 [mksafe:3] Switch to studio with transition.
2023/03/19 15:31:49 [replay_metadata_18:3] Content type is {audio=pcm(stereo)}.
2023/03/19 15:31:49 [clock.input.srt:3] Delegating synchronisation to active sources
2023/03/19 15:31:49 [ZW_FM:2] Error while sending data: could not write data to host: Unix.Unix_error(Unix.EPIPE, "write", "")!
2023/03/19 15:31:49 [ZW_FM:3] Closing connection...
2023/03/19 15:31:49 [ZW_FM:3] Will try to reconnect in 3.00 seconds.
2023/03/19 15:31:53 [ZW_FM:3] Connecting mount /ZWFM.stl for source@localhost...
2023/03/19 15:31:53 [ZW_FM:3] Connection setup was successful.
```

The fallback file is `https://upload.wikimedia.org/wikipedia/commons/6/66/Aaron_Dunn_-_Sonata_No_1_-_Movement_2.ogg`

toots:
There's been a couple of issues reported lately pointing to some problems with chained ogg bitstreams, most notably https://github.com/savonet/liquidsoap/issues/2965, https://github.com/savonet/liquidsoap/issues/3010, https://github.com/savonet/liquidsoap/issues/2917 so I started investigating it.

The issues that we are tracking are:
* The validity of chained ogg bitstreams (that is bitstreams with more than one track)

Are the stream we implement valid? Where is the spec? What are the other implementations saying?

* Support for metadata in chained ogg bitstreams

Can other demuxers read metadata we produce on chained bitstreams? Can we read metadata on chained bitstreams?

This has raised some interesting issues that should be relevant for the icecast streaming community at large, now that liquidsdoap is used widely in this field.

### ⚠️ Invalid chained ogg bitstream ⚠️

Back in the days when we wrote our ogg muxer, and we're talking about 2009, some 14 years ago (see: https://github.com/savonet/liquidsoap/commit/cc5a44da019e135bae1b719879dd5e97bc06ef7b), there wasn't a lot of clarity in how to synchronously end a logical ogg stream. That is, take an encoder at any point and ask it to end the stream and flush all remaining pages in it.

`libvorbis` provided an API for it but this was the only library being explicitly coupled with `libogg`. For all other libraries, the encoder might not have any data to flush when ending it and `libogg` required at least some data to produce the last page.

Thus, we went with the trick of submitting an empty ogg packet with the `e_o_s` flag set on it. Nothing in the spec prohibited it and it seems reasonable to expect that a decoder would simply consider an empty packet as the end of stream as is standard in unix.

Fast forward 14 years, two things happened:
* The FFmpeg implementation for ogg demuxing considers an empty packet as invalid data.
* The opus spec started using empty packets for packet loss control

This means that the 2016 rfc for ogg/opus encapsulation now prohibits empty packets: https://datatracker.ietf.org/doc/html/rfc7845#section-6. A better knowledge of opus codec from the 2012 codec spec could have pointed to this: https://datatracker.ietf.org/doc/html/rfc6716#section-3.4. This is the same time we did our own implementation of the ogg encapsulation (see: https://github.com/savonet/ocaml-opus/commit/c0cd32b3bb91a2f5638285eef3f149cf8b472f76)

This means that our original assumption is no longer valid and that we are producing invalid ogg bitstreams at least from the perspective of FFmpeg (for all non-vorbis codecs) and all fully compliant opus decoders.

To fix that, a semi hacked `Ogg.Stream.terminate` was implemented in https://github.com/savonet/ocaml-ogg/commit/78edbb2dcc38eebda0c566e5f206d65b81a38912 to allow the production of a final, empty EOS page when needed, something that, this time, is actually explicitly allowed by the spec: https://datatracker.ietf.org/doc/html/rfc3533#section-4
> Eos pages may be 'nil' pages, that is, pages
   containing no content but simply a page header with position
   information and the eos flag set in the page header.

This PR adds support for this new function to all ogg codecs that previously relied on pushing an empty eos packet. This applies to speex, theora and opus.

For flac, this PR and https://github.com/savonet/ocaml-flac/commit/10aecfc2908a99e9a90d3c498521a555035bfbd1 switch to `libflac` native ogg encapsulation implementation, which also makes us compliant with their spec and future-proof. This should fix #2965

### Chained metadata

The other side of this problem is that, for too long, we were testing ogg streams in a closed off environment, that is, decoding streams produced by our own encoder. But now that `input.http` is using ffmpeg internally, it became clear that something was off with ogg chained bitstream metadata.

After fixing our internal bitstream, FFmpeg started being able to parse all metadata in ogg/opus streams, albeit with a memory leak as the library forgets to clear previous metadata when adding new, which results in the new metadata being appended. This PR implements a workaround for that and a patch will be submitted to FFmpeg soon.

However, ogg/flac streams are still missing chained stream metadata. This is due to a different ways that FFmpeg parses chained ogg/flac bitstream. A patch has already been submitted and will hopefully be applied at some point.

### DIFF
--- a/.github/opam/liquidsoap-windows.opam
+++ b/.github/opam/liquidsoap-windows.opam
@@ -106,7 +106,7 @@ conflicts: [
   "magic-windows" {< "0.6"}
   "mem_usage-windows" {< "0.0.3"}
   "curl-windows" {< "0.9.2"}
-  "ogg-windows" {< "0.7.0"}
+  "ogg-windows" {< "0.7.4"}
   "opus-windows" {< "0.2.0"}
   "portaudio-windows" {< "0.2.0"}
   "pulseaudio-windows" {< "0.1.4"}
--- a/dune-project
+++ b/dune-project
@@ -103,7 +103,7 @@
     (mad (< 0.5.0))
     (magic (< 0.6))
     (mem_usage (< 0.0.3))
-    (ogg (< 0.7.0))
+    (ogg (< 0.7.4))
     (opus (< 0.2.0))
     (portaudio (< 0.2.0))
     (posix-time2 (< 2.0.2))
--- a/src/core/dune
+++ b/src/core/dune
@@ -415,6 +415,9 @@
  (library_flags -linkall)
  (wrapped false)
  (optional)
+ (foreign_stubs
+  (language c)
+  (names ogg_flac_encoder_stubs))
  (modules liq_flac_ogg_decoder ogg_flac_encoder ogg_flac_duration))
 
 (library
--- a/src/core/io/ffmpeg_io.ml
+++ b/src/core/io/ffmpeg_io.ml
@@ -113,6 +113,13 @@ class input ?(name = "input.ffmpeg") ~autostart ~self_sync ~poll_delay ~debug
           Ffmpeg_decoder.mk_decoder ~streams ~target_position:(ref None) input
         in
         let buffer = Decoder.mk_buffer ~ctype:self#content_type self#buffer in
+        (* FFmpeg has memory leaks with chained ogg stream so we manually
+           reset the metadata after fetching it. *)
+        let get_metadata stream =
+          let m = Av.get_metadata stream in
+          Av.set_metadata stream [];
+          m
+        in
         let get_metadata () =
           normalize_metadata
             (Ffmpeg_decoder.Streams.fold
@@ -120,10 +127,10 @@ class input ?(name = "input.ffmpeg") ~autostart ~self_sync ~poll_delay ~debug
                  m
                  @
                  match stream with
-                   | `Audio_frame (stream, _) -> Av.get_metadata stream
-                   | `Audio_packet (stream, _) -> Av.get_metadata stream
-                   | `Video_frame (stream, _) -> Av.get_metadata stream
-                   | `Video_packet (stream, _) -> Av.get_metadata stream)
+                   | `Audio_frame (stream, _) -> get_metadata stream
+                   | `Audio_packet (stream, _) -> get_metadata stream
+                   | `Video_frame (stream, _) -> get_metadata stream
+                   | `Video_packet (stream, _) -> get_metadata stream)
                streams
                (Av.get_input_metadata input))
         in
--- a/src/core/ogg_formats/ogg_flac_encoder.ml
+++ b/src/core/ogg_formats/ogg_flac_encoder.ml
@@ -20,6 +20,9 @@
 
  *****************************************************************************)
 
+external set_stream_eos : Ogg.Stream.stream -> unit
+  = "liq_ocaml_ogg_stream_set_eos"
+
 let create_encoder ~flac ~comments () =
   let samplerate = Lazy.force flac.Flac_format.samplerate in
   let p =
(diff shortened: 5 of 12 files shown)


gabsoftware:
Aw bummer, the CI failed for that one because of some apt update fun, so it's not in the 2.2x rolling release yet.

toots:
Should be good now!

gabsoftware:
Yay, thanks! Will test ASAP! 👍

Issue #3067, opened 2023-05-09
[Harbor] input.harbor.ssl don't accept connections after some time


codeurimpulsif:
**Describe the bug**

I use `input.harbor.ssl` to input external stream but after some time (I don't know exactly how many time, few days for sure) I can't connect anymore.

Note: This issue seem similar to #1403 where I leave a comment.

**To Reproduce**

The script:
```
#!/usr/bin/liquidsoap

# Harbor settings
settings.harbor.bind_addrs.set(["#{IPV4_BIND}", "#{IPV6_BIND}"])
settings.harbor.ssl.certificate.set("#{TLS_CERTIFICATE_PATH}/fullchain.pem")
settings.harbor.ssl.private_key.set("#{TLS_CERTIFICATE_PATH}/privkey.pem")

# TLS Live (using Harbor)
live = input.harbor.ssl(id="live",
  buffer=2.,
  max=10.,
  port=8005,
  icy=true,
  icy_metadata_charset = "UTF-8",
  metadata_charset = "UTF-8",
  auth=auth,
  "live"
)

output.dummy(mksafe(live))
```

A netstat (`netstat -altupn | grep <harbor-port> | grep "CLOSE_WAIT"`) or ss (`ss -tap state CLOSE-WAIT | grep "liquidsoap"`) command show me there is a lot of open connections (250 exactly) to the harbor port in `CLOSE_WAIT` state.

I can reproduce the issue by sending lot of connections (using `nc -z <harbor-ip> <harbor-port>`, curl don't seem to reproduce), after some time I can't connect anymore on harbor port and open connections in CLOSE_WAIT state are still here, never flushed.

Liquidsoap logs shows `Failed to accept new client: SSL accept() error: error:00000000:lib(0):func(0):reason(0)` each time I open a connection (and of course add a connection in `CLOSE_WAIT` state to the previous netstat or ss list).

Then at some point (I can't say when exactly) a normal streaming source (I use Butt) can't connect anymore.

Then it reach the limit with log message `Failed to accept new client: Too many open files in accept()`, but only when I try to reproduce with this test. In normal conditions it take days to append and there is no `Too many open files in accept()` log messages, only the `SSL accept() error` ones.

**Expected behavior**

- Connection to the harbor port need to still be possible without restarting Liquidsoap

**Version details**

- OS: Debian 11 (Raspberry Pi 4B, aarch64)
- Version: `2.1.4-debian-bullseye-1`

**Install method**

[...]

PR #3071, opened 2023-05-11
Use a single context for all incoming SSL connections from a given transport. 

toots:
This might a fd leak from `libssl` but this should mitigate most issues.

Fixes: #3067


toots:
This PR has been incorporated in https://github.com/savonet/liquidsoap/pull/3074

PR #3115, opened 2023-05-31
Mockup for a media library.

smimram:
We are able to do something like
```
m = medialib(["~/Music/misc"])
output(playlist.list(m.find(artist_contains="Sting")))
```

Before I complete it further, any comments on useablity / design?

We should be able to store the db persistently (using json) and update only new files...

### DIFF
--- a/src/libs/medialib.liq
+++ b/src/libs/medialib.liq
@@ -1,0 +1,28 @@
+def medialib(directories=[])
+  db = ref([])
+
+  def scan()
+    for d = list.iterator(directories) do
+      for f = list.iterator(file.ls(absolute=true, recursive=true, d)) do
+        log.debug("adding file: #{f}")
+        m = file.metadata(f)
+        m = ("filename",f) :: m
+        db := (f,m) :: db()
+      end
+    end
+  end
+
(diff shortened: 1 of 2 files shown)


toots:
I think that it's a great idea. What do you think of making this a separate `liquidsoap-medialibrary` package? Ideally, this would become a module once we have it and, perhaps, integrate with `opam` to have proper liquidsoap packages!

toots (commented):
I think that this an awesome idea and a tool that will be very useful to our users.

I mostly made some generic comments. I really think that this should be on its own independent opam package. Couple of reasons I can think is:
* Visibility: the tool existence would be more obvious to the user
* Maintainability: once people start using it, more likely than not there will be requests for improvements, bug fixes etc. Having it separate would allow for point releases without having to wait for a main release. We also try to stick to stable bugfixes on minor liquidsoap relesse. A separate package would allow to decouple development effort on this module from the main release cycles.
* Load: we already know the standard library is too big. In fact, I think we should consider having a `liquidsoap-libs-extra` package for the next release.

If the scope is too limited to a separate package, we could bundle it up with other tools and make it a `liquidsoap-libs-mediatools` package.

PR #3117, opened 2023-06-01
Fix `request.dynamic` queueing logic

### LINKED ISSUE #3026: request.dynamic retry_delay value not used
**Describe the bug**
The value I set is not taken into account. This is an issue in both 2.1.4 and 2.2.x

**To Reproduce**
```

def get_request() =
  ...
end

music = request.dynamic(
  id="music", 
  prefetch=1,
  retry_delay=5.,
  get_request
)
```

If liquidsoap throws a runtime error in get_request, ie uncaught json parsing, then get_request is called many times very quickly instead of once every 5 second per the set retry_delay

**Expected behavior**
request.dynamic should respect the retry_delay value set in the script

**Version details**

- Liquidsoap 2.2.0+git@2a3bf62e1
- Docker version 20.10.22, build 3a2c30b
- Image: savonet/liquidsoap:rolling-release-v2.2.x

**Install method**
Docker

**Common issues**
Did not find a related issue

toots:
This should fix #3026

### DIFF
--- a/src/core/outputs/pipe_output.ml
+++ b/src/core/outputs/pipe_output.ml
@@ -546,8 +546,7 @@ class external_output p =
       Unix.open_process_out process
 
     method close_chan chan =
-      try ignore (Unix.close_process_out chan)
-      with Sys_error msg when msg = "Broken pipe" -> ()
+      try ignore (Unix.close_process_out chan) with Sys_error _ -> ()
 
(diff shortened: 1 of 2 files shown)


toots:
I'm merging this. There's a change in behavior in that the requests won't start being queued until the source is being actively animated through `is_ready` but that seems reasonable.

PR #3120, opened 2023-06-02
POC: import/export.

toots:
This is a first stab at a proper module import/export.

Syntax:

Export `"module.liq"`
```liquidsoap
let export foo = "123"
def export bla() =
  print("bla")
end
```

Import:
```liquidsoap
let import { foo, bla } = "module.liq"
```

The implementation works by replacing `"module.liq"` in the import statement with the type-checked term:

```liquidsoap
let _export_ = ()

let _export_.foo = "123"

def _export_.bla() =
  print("bla")
end

_export_
```

Based on this, it should be possible to cache the result of the evaluation of `"module.liq"` to speed up future executions.

### DIFF
--- a/src/lang/dune
+++ b/src/lang/dune
@@ -104,6 +104,7 @@
   extralib
   ground_type
   hooks
+  import
   json_base
   json_lexer
   json_parser
--- a/src/lang/import.ml
+++ b/src/lang/import.ml
@@ -1,0 +1,63 @@
+(*****************************************************************************
+
+  Liquidsoap, a programmable audio stream generator.
+  Copyright 2003-2023 Savonet team
+
+  This program is free software; you can redistribute it and/or modify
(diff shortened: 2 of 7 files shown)


toots:
> A general remark: adding `export` everywhere is very cumbersome. I'd rather have everything exported by default and have a `private` keyword for non-exported values.

I'll think about it but there are a couple of reasons I could see against it:

* This means the export stuff would be everywhere in all file parsed, not just those intended to be a module
* The pattern of explicit export is more common that implicit export. It is also easier to proof read if there's every an issue with an export
* We accept variable override by default but I think (not implemented) that we should actually refuse multiple export with the same name as it is hard to proof read/debug.
* Eventually, it'd be nice to be able to do some dead code elimination. Promoting an explicit pattern would make this more efficient as well.

Issue #3153, opened 2023-06-16
ID3v2 metadata in HLS streams


toots:
### Discussed in https://github.com/savonet/liquidsoap/discussions/3152

<div type='discussions-op-text'>

<sup>Originally posted by **fkane** June 16, 2023</sup>
Hi, I'm trying to integrate my station's HLS stream into Sonos devices. In order to display metadata such as artist, title, and cover art, they require it to be embedded in ID3v2 frames, and have specific requirements on how to embed URL's to the album art.

Is there any existing mechanism for embedding ID3v2 metadata into HLS streams, and any existing hooks I can use to implement the WXXX tag for a URL to the album art in the way they want?

Their requirements are [here](https://developer.sonos.com/build/content-service-add-features/supported-audio-formats/), and pasted below for convenience:

For HLS content, use [ID3v2 frames](http://id3.org/id3v2.3.0#Declared_ID3v2_frames) for metadata, as defined in the [ID3 tag version 2.3.0 documentation](http://id3.org/id3v2.3.0). Sonos players read the following ID3v2 tags:

COMM—comments
PRIV—private frame
TALB—album/move/show
TCMP—iTunes compilation
TCOM—composer
TCON—content type
TIT2—title/song name/content
TPE1—lead performer, soloist
TPE2—band/orchestra
TPOS—part of a set
TRCK—track number
WXXX—user defined url link

Use the WXXX tag if you want your stream to have artwork that changes, like an Internet radio stream that changes artwork for each song. We use this tag to specify the size of the artwork and the URL. To do this, you must separate these strings with a null character, such as \0. The first string should start with “artworkURL_” followed by the width of the image in pixels and “x”. For example, this string will be set to “artworkURL_640x” for an image that is 640 pixels wide. The second string should be the full album art URI (example: http://server/path/art.jpg”). Together, the strings would look like this: “artworkURL_640x\0http://example.com/artwork/1341414.jpg”.</div>

Moonbase59:
I haven’t yet checked what’s possible with HLS, just keep in mind there are people using other formats than MP3 (AAC, OGG, FLAC, …) that use different tagging schemes.

Also, the ID2v2 `WXXX` tags do have some drawbacks, for instance the URL needs to be encoded in ISO-8859-1, and (at least in ID3v2.4) there can be a multitude of `WXXX` tags for different purposes. The above is a _very Sonos-specific_ solution.

[...]

PR #3154, opened 2023-06-17
Add id3v2 export to HLS output.

toots:
This PR implements support for ID3 metadata for HLS output.

From the doc:

## Metadata

HLS outputs supports metadata in two ways:

- Through a `timed_id3` metadata logical stream with the `mpegts` format.
- Through regular ID3 frames, as requested by the [HLS specifications](https://datatracker.ietf.org/doc/html/rfc8216#section-3.4) for `adts`, `mp3`, `ac3` and `eac3` formats.
- There is currently no support for in-stream metadata for the `mp4` format.

Metadata parameters are passed through the record methods of the streams' encoders. Here's an example

```liquidsoap
output.file.hls(
  "/path/to/directory",
  [
   ("aac",
      %ffmpeg(format="adts", %audio(codec="aac")).{
        id3_version = 3
       }),
   ("ts-with-meta",
      %ffmpeg(format="mpegts", %audio(codec="aac")).{
        id3_version = 4
     }),
   ("ts",
      %ffmpeg(format="mpegts", %audio(codec="aac")).{
        id3 = false
      }),
   ("mp3",
      %ffmpeg(format="mp3", %audio(codec="libmp3lame")).{
        replay_id3 = false
      })
  ],
  source
)
```

Parameters are:

- `id3`: Set to `false` to deactivate metadata on the streams. Defaults to `true`.
- `id3_version`: Set the `id3v2` version used to export metadata
- `replay_id3`: By default, the latest metadata is inserted at the beginning of each segment to make sure new listeners always get the latest metadata. Set to `false` to disable it.

Metadata for these formats are activated by default. If you are experiencing any issues with them, you can disable them by setting `id3` to `false`.

Fixes: #3153

### DIFF
--- a/doc/content/hls_output.md
+++ b/doc/content/hls_output.md
@@ -69,6 +69,48 @@ Liquidsoap also provides `output.harbor.hls` which allows to serve HLS streams d
 liquidsoap. Their options should be the same as `output.file.hls`, except for harbor-specifc options `port` and `path`. It is
 not recommended for listener-facing setup but can be useful to sync up with a caching system such as cloudfront.
 
+## Metadata
+
+HLS outputs supports metadata in two ways:
+
+- Through a `timed_id3` metadata logical stream with `mpegts` and `mp4` formats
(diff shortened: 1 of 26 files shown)


Russsgithub:
This is great. Thanks

PR #3162, opened 2023-06-20
Add alpha parameter.

### LINKED ISSUE #3160: LS 2.2.0+git@dcf405ff4: Alpha=00 ignored in `video.add_rectangle(color=0xAARRGGBB, …)`
**Describe the bug**
A small oversight probably: Alpha values of 0x00 in `video.add_rectangle(color=0xAARRGGBB, …` are ignored and produce a solid color instead.

**To Reproduce**
Minimal script:
```
#
# Alpha 00 ignored in video.add_rectangle()
#

v = video.testsrc(width=1280, height=720)

# this should not display (fully transparent AARRGGBB)
# but shows as a red rectangle (transparency ignored if =00)
v = video.add_rectangle(color=0x00eb0000, height=30, width=320, x=0, y=532, v)

# this is a bad workaround (use alpha =01 instead)
v = video.add_rectangle(color=0x01eb0000, height=30, width=320, x=0, y=562, v)

output.sdl(v)
```

Current workaround: Use alpha=0x01 instead, which is _almost_ fully transparent but normally not distinguishable.

**Expected behavior**
Alpha=0x00 work and "display" a fully transparent color. This is sometimes needed to hide something that is actually there, like my "Live: " indicator, which shall only show when a live DJ is connected (I use this with _AzuraCast_):

video m2t - VLC media player_010

**Version details**

- OS: Linux Mint 21.1, Ubuntu 22.04 LTS Docker
- LS Version: 2.2.0+git@dcf405ff4

**Install method**
`.deb` from release-assets.

smimram:
Fixes #3160.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -115,7 +115,7 @@ Changed:
 - Reworked remote request file extension resolution (#2947)
 - REMOVED `osx-secure-transport`. Doubt it was ever used, API deprecated
   upstream (#3067)
-- Renamed `rectangle` to `add_rectangle`.
+- Renamed `rectangle` to `add_rectangle`, and similarly for `line`.
 
 Fixed:
 
--- a/src/core/operators/video_effects.ml
+++ b/src/core/operators/video_effects.ml
@@ -467,7 +467,7 @@ let _ =
 
 let _ =
   let return_t = return_t () in
-  Lang.add_operator ~base:Modules.video "line"
+  Lang.add_operator ~base:Modules.video "add_line"
     ([
        ( "",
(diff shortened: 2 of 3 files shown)


toots (approved):
Looks great thanks! Is that a new operator? Do we need a breaking change warning in the changelog? Or a backward compatible params parsing?

Thks!

smimram:
The support for alpha channel was completely broken, I "fixed" this a few days ago in #3144. I don't think that many people have had time to use that, so I don't think that backward compat is really an issue here.

PR #3165, opened 2023-06-22
Only raise a warning when using an invalid color.

smimram:
Refs #3157.


src/core/operators/video_effects.ml
```diff
@@ -23,6 +23,19 @@
 open Mm
 open Source
 
+let log = Log.make ["video"]
+
+let rgb_of_int c =
+  let c =
+    if c < 0 || c > 0xffffff then (
+      log#important "not a valid color: 0x%x" c;
```
toots, line 31:
```suggestion
      log#important "color 0x%x is greater than maximum assignable value 0xffffff" c;
```

toots (approved):
Looking good!

PR #3175, opened 2023-06-28
Test with OCaml 5.

smimram:


toots (approved):
Thanks! opam CI checks are disabled until we release all pending dependencies.

PR #3204, opened 2023-07-11
Improve `output.file` doc.

smimram:
And improve logic in case of double-reopen with metadata.

It was not clear to me that the following should work, but it should:
```
output.file(%mp3, {time.string("archive/%Y-%m-%d/%H-%M-%S.mp3")}, radio,
            should_reopen=fun(~metadata, ~error) -> if null.defined(metadata) or null.defined(error) then null() else 3600. end)
```


toots (approved):
Thanks!

PR #3208, opened 2023-07-12
Add back `mux_*` and `drop_*` functions.

smimram:
The new mechanism for manipulating tracks is nice, but it requires users to understand in depth the way records works. I suggest that we keep the muxin and dropping functions. In passing I have renamed them to `source.mux.*` and `source.drop.*`.


toots (approved):
Thanks.

PR #3210, opened 2023-07-12
Add support for never type.

### LINKED ISSUE #3203: Sources without some fields
The syntax
```
s = (single("test.png"):source(audio=none))
```
for specifying that there is no audio is not working anymore. How do we do this now?

smimram:
Source types are now extensible by default, i.e. we specify an open record rather than a closed one. I think that this is reasonable because most of the time you want to ensure that a track has a specific type (e.g. audio is in `pcm` format) rather than want to specify all the track types (most of the time we want to have a metadata track but don't want to have to write this down).

In order to have the following example pass
```
s = (sine():source(video=none))
output(s)
```
I needed to slightly modify the subtyping algorithm since `sine()` has no field `video` whereas we want it to be a subtype of a record with a `never` field.

Fixes #3203.

### DIFF
--- a/src/lang/parser.mly
+++ b/src/lang/parser.mly
@@ -164,7 +164,7 @@ open Parser_helper
 %type <Parser_helper.ty_content_arg> ty_content_arg
 %type <Parser_helper.ty_content_args> ty_content_args
 %type <Type.t> ty_source
-%type <(string * Parser_helper.ty_content) list> ty_source_params
+%type <bool * (string * Parser_helper.ty_content) list> ty_source_params
 %type <Type.t list> ty_tuple
 %type <Term.pattern> var_pattern
 %type <Parser_helper.varlist> varlist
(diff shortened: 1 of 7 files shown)


toots (changes requested):
I don't think that source typing should be extensible by default otherwise we have no way of writing closed types. Also I believe it is more intuitive to be closed by default. We should add this syntax: `source(audio=none, ...)` which shouldn't be too hard to do.

smimram:
Ok the syntax for extensible sources is now implemented! :)

src/lang/parser.mly
```diff
@@ -324,12 +324,13 @@ meth_ty:
...
-  | VARLPAR RPAR                  { mk_source_ty ~pos:$loc $1 [] }
-  | VARLPAR ty_source_params RPAR { mk_source_ty ~pos:$loc $1 $2 }
+  | VARLPAR RPAR                  { mk_source_ty ~pos:$loc ~extensible:false $1 [] }
+  | VARLPAR ty_source_params RPAR { mk_source_ty ~pos:$loc ~extensible:(fst $2) $1 (snd $2) }
 
 ty_source_params:
-  | VAR GETS ty_content { [$1,$3] }
-  | VAR GETS ty_content COMMA ty_source_params { ($1,$3)::$5 }
+  | VAR GETS ty_content { false, [$1,$3] }
+  | DOTDOTDOT { true, [] }
```
toots, line 332: You want to allow:
```liquidsoap
source(audio=pcm, ..., video=canvas)
```
This seems odd? I was more thinking:
```ocaml
ty_source:
  | VARLPAR RPAR                  { mk_source_ty ~pos:$loc ~extensible:false $1 [] }
  | VARLPAR DOTDOTDOT RPAR        { mk_source_ty ~pos:$loc ~extensible:true $1 [] } 
  | VARLPAR ty_source_params RPAR { mk_source_ty ~pos:$loc ~extensible:false $1 $2 }
  | VARLPAR ty_source_params COMMA DOTDOTDOT RPAR { mk_source_ty ~pos:$loc ~extensible:true $1 $2 }
```

tests/language/typing.liq
```diff
@@ -188,6 +188,12 @@ def f() =
   ([({}:{foo?:int}), (), {foo = 123}]:[{foo?:int}])
   ([(), {foo = 123}]:[{foo?:int}])
 
+  # This one is more subtle than it seems because we have no field video on the
+  # left and a video:never field on the right when testing for the subtyping,
+  # see #3210.
+  s = (sine():source(video=none))
+  s = (sine():source(video=none,...))
```
toots, line 195: I think we want the same thing :-)

toots (commented):
One more suggestion!

src/lang/parser.mly
smimram: This is actually what I first wrote but you get a parser conflict if you write it this way...

And once you go to my solution, there is no real reason of for forbidding `...` in the middle (if you do you get more parser cases...)

toots: Cool that works for me then.

PR #3213, opened 2023-07-12
Use TLS1.3.

smimram:
This avoid the following (justified) warning:
```
Alert deprecated: SSLv23
SSL 2.0 was deprecated in 2011 by RFC 6176.
```

### DIFF
--- a/src/core/builtins/builtins_ssl.ml
+++ b/src/core/builtins/builtins_ssl.ml
@@ -51,7 +51,7 @@ let ssl_socket transport ssl =
   end
 
 let server ~read_timeout ~write_timeout ~password ~certificate ~key transport =
-  let context = Ssl.create_context Ssl.SSLv23 Ssl.Server_context in
+  let context = Ssl.create_context Ssl.TLSv1_3 Ssl.Server_context in
   let () =
     ignore
       (Option.map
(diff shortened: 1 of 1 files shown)


toots:
I've been reluctant to do that unless we have a user setting to allow more permissive protocols. Otherwise, we are restricting the use of the software in legacy situations.

smimram:
2011 seems to be real legacy...

toots:
With the new transport abstraction it shouldn't be too hard to add an option to `transport.http.ssl`

toots (approved):
Added parameter.

src/core/builtins/builtins_ssl.ml
```diff
@@ -159,6 +170,29 @@ let _ =
...
       in
+      let protocol_val = List.assoc "protocol" p in
+      let protocol =
+        match Lang.to_string protocol_val with
+          | "ssl.2.3" ->
+              log#severe "ssl.2.3 protocol was deprecated in 2011 by RFC 6176.";
+              Ssl.SSLv23 [@alert "-deprecated"]
+          | "ssl.3" ->
+              log#severe
+                "ssl.3 protocol was deprecated in June 2015 by RFC 7568";
```
toots, line 181:
```suggestion
                "ssl.3 protocol was deprecated in June 2015 by RFC 7568.";
```

vitoyucepi:
It is not clear to me how to specify protocols.
1. What is `ssl.2.3`?
   I know from the source that it will enable both *SSLv2* and *SSLv3*.
   Without proper documentation, it will look like the previously unknown mysterious protocol *SSLv2.3*.
1. What will happen if I set `tls.1`?
   Is it the minimum or the only supported protocol version?

I know that *TLSv1.3* is a modern protocol that doesn't require cipher suites. Older protocols usually require cipher suite tweaking to ensure reasonable level of security.
I use [Mozilla SSL configuration generator](https://ssl-config.mozilla.org/). It provides sane configurations for some of the most popular servers.

toots:
@vitoyucepi I switch to min/max protocol and added more documentation, this should make it more clear.

PR #3223, opened 2023-07-13
Use min/max protocol in SSL

toots:


src/core/builtins/builtins_ssl.ml
```diff
@@ -139,12 +152,27 @@ let _ =
...
+        Some (Lang.string "ssl.3"),
+        Some
+          "Minimal accepted SSL protocol. One of, from least recent to most \
+           recent: `\"ssl.3\"`, `\"tls.1\"`, `\"tls.1.1\"`, `\"tls.1.2\"` or \
+           `\"tls.1.3\"`. The most recent available protocol between client \
+           and server is negotiated when initiating communication between \
+           minimal and maximal protocol version. All protocols up to \
+           `\"tls.1.2\"` and above are now deprecated so you might want to set \
+           this value to one of those two." );
+      ( "max_protocol",
```
vitoyucepi, line 166: What's the point of having maximum protocol?

src/core/builtins/builtins_ssl.ml
```diff
@@ -139,12 +152,27 @@ let _ =
...
         Lang.string_t,
         Some (Lang.string "tls.1.3"),
         Some
-          {|Accepted SSL protocol. One of: `"ssl.2.3"`, `"ssl.3"`, `"tls.1"`, `"tls.1.1"`, `"tls.1.2"` or `"tls.1.3"`|}
-      );
+          "Maximal accepted SSL protocol. One of, from least recent to most \
+           recent: `\"ssl.2.3\"`, `\"ssl.3\"`, `\"tls.1\"`, `\"tls.1.1\"`, \
```
vitoyucepi, line 171: `ssl.2.3` is missing everywhere.

toots: It's not accepted by these functions. I've removed it from the doc.

src/core/builtins/builtins_ssl.ml
vitoyucepi: I think function should have `min_protocol` only
`http.transport.ssl(min_protocol="TLSv1.2"])`
or accept a list of protocols
`http.transport.ssl(protocols=["TLSv1", "TLSv1.1", "TLSv1.2", "TLSv1.3"])`

toots: I made everything `nullable` so by default it's the highest. I don't think the API allows to select only a subset of protocols.

PR #3225, opened 2023-07-13
Add univ_descr to constraints to describe them when they're the only one on their type.

toots:
It's only used for pcm track type but could definitely be useful in other cases.

Screenshot 2023-07-13 at 8 28 54 AM

### DIFF
--- a/src/core/types/format_type.ml
+++ b/src/core/types/format_type.ml
@@ -197,13 +197,14 @@ let is_format f m =
   let module Content = (val m : Content) in
   Content.is_format f
 
-let check_track ~t modules =
+let check_track ?univ_descr ~t modules =
   {
     Type.t;
     constr_descr =
       Printf.sprintf "a track of type: %s"
         (Utils.concat_with_last ~last:"or" ", "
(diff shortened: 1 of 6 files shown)


src/core/types/format_type.ml
```diff
@@ -253,7 +255,8 @@ let internal_tracks =
 let track =
   {
     Type.t = Track;
-    constr_descr = "a source track";
+    constr_descr = "a track";
```
smimram, line 258: "track" is used for two things in liquidsoap:
- a track = a song
- a track = a media contents (audio / video / midi / metadata / etc.)
I'd suggest calling _channel_ the second one to make the distinction clear. What do you think?

```suggestion
    constr_descr = "a channel";
```

toots: Yes but channel is also for audio. How about `a media track`?

toots, line 258:
```suggestion
    constr_descr = "a media track";
```

smimram (approved):

PR #3228, opened 2023-07-13
Fix `http.transport.ssl` function documentation

vitoyucepi:
Remove maximum protocol version `ssl.2.3` from the `http.transport.ssl` function documentation.
Fixup for #3223.


toots:
Thanks!

Issue #3202, opened 2023-07-10
Make image in video.add_image a getter — takes wrong image.


Moonbase59:
**Describe the bug**
See also: #3178 

https://github.com/savonet/liquidsoap/commit/f4e378b5625fc4ac2f8ad67233e190921bc98923 doesn’t correctly work yet—it takes wrong images, I suspect the [new_file/current_file logic](https://github.com/savonet/liquidsoap/commit/f4e378b5625fc4ac2f8ad67233e190921bc98923) being the cause

**To Reproduce**
Minimal script:
```
#
# test4.liq -- test changing video.add_image
#

logo = ref("niteradio-cover.png")

v = video.testsrc(width=1280, height=720)
v = video.add_image(x=50, y=50, width=100, height=100, file=logo, v)
v = video.add_text(size=20, x=50, y=670, color=0x000000, logo, v)

thread.run(every=5., {logo.set(list.pick(["niteradio-cover.png","moonbase.jpg"]))})

output.audio_video(mksafe(v))
```
(Just use two other images instead of `niteradio-cover.png` and `moonbase.jpg`.)

You’ll see it starts off correctly and then develops a "shift by one", displaying the wrong image (the correct one is shown as text in the lower left of the video).

**Expected behavior**
Use the correct image.

**Version details**

- OS: Linux Mint 21.1, Ubuntu 22.04.2 LTS Server
- LS Version: 2.2.0+git@0a90fcf99

**Install method**
`.deb` from release-assets.

toots:
@Moonbase59 any chance you could try the builds from https://github.com/savonet/liquidsoap/actions/runs/5580827296 ?

Issue #3200, opened 2023-07-10
Improve `image`.


smimram:
A comment on https://github.com/savonet/liquidsoap/commit/f4e378b5625fc4ac2f8ad67233e190921bc98923 before I forget.

It would be nice to add support for getters in `image` too and use `image` in order to implement `add_image`.

Moonbase59:
Totally agree.

PR #3237, opened 2023-07-17
Improve `image` and `video.add_image`.

toots:
* Make `image` accept getter
* Unify `image` and `video.add_image`

Fixes: #3202, #3200


smimram (approved):
This was clearly the thing to do, thanks!

PR #3270, opened 2023-07-28
Export metrics during tests.

smimram:
The idea is to have one file in a simple yaml format exporting metrics from tests.


tests/test.liq
```diff
@@ -79,6 +79,12 @@ def test.raise(fn) =
   try ignore(fn()) catch _ do () end
 end
 
+# Add a metric
+def test.metric(~category, ~name, ~value, ~unit)
+  data = "- name: \"#{(name:string)}\"\n  category: \"#{(category:string)}\"\n  value: #{(value:float)}\n  unit: \"#{(unit:string)}\"\n  time: #{(time():float)}\n\n"
+  file.write(append=true, data=data, "/tmp/metrics.yaml")
```
toots, line 85: I would suggest to use `./metrics.yaml` or something like that so that the file is located inside `_build/default/tests`

smimram: I used to do that, but the tests can be run from various places so that an absolute path is much easier to handle...

.github/workflows/ci.yml
```diff
@@ -212,6 +212,30 @@ jobs:
         run: |
           cd /tmp/liquidsoap-full/liquidsoap
           sudo -u opam -E ./.github/scripts/test-posix.sh ${{ matrix.target }}
+      - name: Finalize metrics
+        run: |
+          sudo -u opam -E git config --global --add safe.directory '*'
+          cd /tmp/liquidsoap-full/liquidsoap
```
toots, line 218:
```suggestion
          cd /tmp/liquidsoap-full/liquidsoap
          mkdir /tmp/${{ github.run_number }}/metrics
```

toots (approved):
Thanks for your patience with this!

Issue #3265, opened 2023-07-26
[v2.2.0] output.harbor OGG container broken


DigiBC:
**Describe the bug**

As already confirmed in #3255 there's another issue with <code>output.harbor</code> in version 2.2.0:
Using the OGG container results in an empty or garbled stream, even though the log doesn't indicate any problems after successful connection with a streaming client.
Despite connection, none of the tested streaming clients could render the stream.
(By the way, other streams like MP3 work flawlessly.)

**To Reproduce**

I've used the following simple code examples:
```
output.harbor(%vorbis, sine(), mount="/stream")
```
```
output.harbor(%ogg(%flac), sine(), mount="/stream")
```
Streaming clients like VLC report errors: 
```
ogg debug: File length is 0 bytes
ogg warning: couldn't find any ogg logical stream
```
CURL shows the following:
```
~$ curl -v --http0.9 http://localhost:8000/stream
*   Trying 127.0.0.1:8000...
* Connected to localhost (127.0.0.1) port 8000 (#0)
> GET /stream HTTP/1.1
> Host: localhost:8000
> User-Agent: curl/7.81.0
> Accept: */*
> 
* Mark bundle as not supporting multiuse
< HTTP/1.1 200 OK
< Content-type: application/ogg
* no chunk, no close, no size. Assume close to signal end
< 
```

**Expected behavior**

It worked flawlessly in version 2.1.4 and earlier.

**Version details**

- OS: Linux Mint 21.2 (Ubuntu 22.04 'Jammy') amd64
- Version: Liquidsoap Rolling Release 2.2.1 (e3b25cd)

**Install method**

Install of intermediate package '[liquidsoap-e3b25cd_2.2.1-ubuntu-jammy-1_amd64](https://github.com/savonet/liquidsoap/suites/14550322259/artifacts/824023742)'

PR #3276, opened 2023-07-28
Make encoder header more robust by using a callback.

toots:
Fixes: #3265

### DIFF
--- a/src/core/encoder/encoder.ml
+++ b/src/core/encoder/encoder.ml
@@ -314,10 +314,7 @@ let dummy_hls encode =
 
 type encoder = {
   insert_metadata : Export_metadata.metadata -> unit;
-  (* Encoder are all called from the main
-     thread so there's no need to protect this
-     value with a mutex so far.. *)
-  mutable header : Strings.t;
+  header : unit -> Strings.t;
   hls : hls;
   encode : Frame.t -> int -> int -> Strings.t;
   stop : unit -> Strings.t;
@@ -348,6 +345,7 @@ let get_factory fmt =
       (* Protect all functions with a mutex. *)
       let m = Mutex.create () in
       let insert_metadata = Tutils.mutexify m insert_metadata in
+      let header = Tutils.mutexify m header in
       let {
         init;
         init_encode;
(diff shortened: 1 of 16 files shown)


DigiBC:
@toots

Maybe I'm too late because you're working on a fix already, but here's an observation of the original issue that I'd like to share with you:
Most streaming clients fail with an error while VLC keeps the HTTP connection and continues "listening". As soon as there's a track change in Liquidsoap which results in a new logical OGG stream, VLC suddenly starts playback. (VLC playback isn't seamless on track changes, but that's a flaw of VLC.)

toots:
@DigiBC yeah that tracks. We had a mechanism to keep the last ogg header on hand to send to the client on initial connection and it got broken somehow. This PR fixes it.

Couple of things worth noting:
* FFmpeg is bad with chained ogg streams so `input.http` might not be super well behaved with live ogg streams
* The header support is only for the various internal `%ogg` encoder (including `%vorbis` and etc). It won't work with `%ffmpeg` ogg support. It should work with `%wav` tho, which is nice but anecdotical.

Overall, and quite regretfully, looking at back over time really shows that the ogg spec for chaining tracks is pretty bad. It was never well understood or implemented correctly and we might have had it wrong for some formats for years too (see [this explanation](https://www.liquidsoap.info/blog/2023-05-17-ogg-compliance/)).

At this point, I would recommend _avoiding_ the ogg container unless you are willing to strip all track marks and metadata. The number of decoder/clients who get it wrong is just too big.

DigiBC:
@toots

Many thanks for the explanation!
I agree that OGG FLAC has quite some flaws, but so far I've had only good experience with <code>output.harbor</code> feeding a digital radio audio encoder with built-in GStreamer and ICY metadata support on the same machine. The stream works perfectly and without any glitches for months...
I'd like to continue to do that.   :relaxed:

toots:
Ha yes, if you have control over the listener or for a limited exposure it's a different story of course.

DigiBC:
After testing the artefact _liquidsoap-fix-3265_2.3.0-ubuntu-jammy-1_amd64.deb_ I can confirm that it works again!
Thank you!
:thumbsup:

toots:
Great! Just pushed the patch to the next rolling release!

PR #3298, opened 2023-08-06
feat: Enhance data value set in http.response.json

### LINKED ISSUE #3296: `response.json` should produce compact JSON
I'd like to see `response.json` return compact by default.
https://github.com/savonet/liquidsoap/blob/220ff717e805f0de2b3a8ddacc89ec8bf8500321/src/libs/http.liq#L197-L203

**Describe the solution you'd like**
```ruby
def json(~comact=true, v) = 
  if headers_sent() then 
    error.raise(error.invalid, "HTTP response has already been sent for this value!") 
  end 
  content_type := "application/json; charset=utf-8" 
  data := json.stringify(v, compact=compact) 
end 
```

**Describe alternatives you've considered**
N/A

**Additional context**
By default, `JSON.stringify` from js produces minified json.

vitoyucepi:
## Summary
- Add the `compact` named argument, set to `true` by default, into `http.response.json`.
- Forward the `compact` named argument from `http.response.json` into `json.stringify` to produce minified JSON by default.
- Add a newline symbol to the end of the JSON string sent by `http.response.json`.

Closes #3296.

### DIFF
--- a/src/libs/http.liq
+++ b/src/libs/http.liq
@@ -194,12 +194,12 @@ def http.response(~http_version="1.1",
     header("Location", location)
   end
 
-  def json(v) =
+  def json(~compact=true, v) =
     if headers_sent() then
       error.raise(error.invalid, "HTTP response has already been sent for this value!")
     end
     content_type := "application/json; charset=utf-8"
-    data := json.stringify(v)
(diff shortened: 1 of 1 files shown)


toots (approved):
Looks great, thanks! Can you add an entry in `CHANGES.md`?

PR #3299, opened 2023-08-06
[2.2] feat: Enhance data value set in http.response.json

vitoyucepi:
## Summary
- Add the `compact` named argument, set to `true` by default, into `http.response.json`.
- Forward the `compact` named argument from `http.response.json` into `json.stringify` to produce minified JSON by default.
- Add a newline symbol to the end of the JSON string sent by `http.response.json`.

Backport of #3298.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,6 +5,8 @@ Changed:
 - BREAKING: on HLS outputs, `on_file_change` events are
   now `"created"`, `"updated"` and `"deleted"`, to better
   reflect the new atomic file operations (#3284)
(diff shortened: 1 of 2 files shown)


toots (approved):
Great thanks.

I didn't realize you intended to backport it. 

Usually my changelog policy with commits going in both `main` and rolling release is to add the changelog entry in the rolling release and make it silent in `main` with the assumption that the next major release comes after the pending rolling release, where the changes is mentioned first.

What do you think?

vitoyucepi:
I'll move changelog entry from #3298 into this PR.

PR #3355, opened 2023-08-27
Allow a float getter for the offset of `on_offset`.

smimram:
This addresses #3345.


toots (approved):
Looks good thank you!

PR #3366, opened 2023-08-29
Count number of protocols.

smimram:


toots (commented):
Could we implement this using a scripted operator that counts the number of keys in a record?

smimram:
I don't think that the protocols are currently exported on the Liquidsoap side unfortunately.

PR #3376, opened 2023-08-31
fix: Deprecated settings format

vitoyucepi:
## Summary
- Replace old style `set()` function with new assignment `:=` style for settings and recommendations.

## Additional
I don't know how to trigger this error
https://github.com/savonet/liquidsoap/blob/cc8f858da0f29f126e620b9af78a9edaf5f3772e/src/runtime/main.ml#L556-L561

### DIFF
--- a/src/libs/server.liq
+++ b/src/libs/server.liq
@@ -2,6 +2,6 @@
 # @category Interaction
 # @param ~port Port on which we should listen.
 def server.telnet(~port=1234) =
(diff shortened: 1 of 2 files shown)


toots:
The code you pointed at seems to be used for `settings.init.daemon.pidfile.path` and `settings.log.file.path`

toots (approved):
Thanks!

PR #3377, opened 2023-08-31
[2.2] fix: Deprecated settings format

vitoyucepi:
## Summary
- Replace old style `set()` function with new assignment `:=` style for settings and recommendations.

Backport of #3376.


toots:
Thanks!

Issue #3418, opened 2023-09-21
interactive.harbor handles does not handle [&<>"'] characters well


ghostnumber7:
**Describe the bug**
When using `interactive.harbor()` there are a few issues with the `/interactive` form:
* If a field description contains a `<` character it breaks the interfase
* If an `interactive.string()` value contains some combinations of `'<>` if breaks the interfase
* If an `interactive.string()` value contains some combinations of `+&=` the value is not saved correctly

**To Reproduce**
```
interactive.harbor()
metadataTransformRegex = interactive.string(
  description = "Metadata transform regex (Example: StreamTitle='(?<artist>[^~]*?)\\s*~(?<title>[^~]*?)\\s*~ )",
  "metadataTransformRegex", "'>oops"
)
```

Go to `localhost:8000/interactive` and see it broken

**Expected behavior**
HTML entities should be escaped and form should be url encoded when submitting

**Version details**

- Ubuntu
- 2.2.0

**Install method**
opam

toots:
Fixed with 81f4bcae61368d72f306489800d7bbc6565934ba and 113ead700

PR #3419, opened 2023-09-21
fixes interactive.harbor issues. ref #3418

ghostnumber7:
fiexes #3418

### DIFF
--- a/src/libs/string.liq
+++ b/src/libs/string.liq
@@ -381,3 +381,21 @@ let string.char.ascii.number = list.init(10, fun (c) -> c + 48)
 def string.char.ascii.random(range=[...string.char.ascii]) =
   string.char(list.nth(range, random.int(min=0, max=list.length(range) - 1)))
 end
+
+# Escape HTML entities.
+# @category String
+def string.escape.html(s) =
+  string.replace(
+    pattern = "[&<>\"']",
+    fun (c) -> begin
+      if c == "&" then "&amp;"
+      elsif c == "<" then "&lt;"
+      elsif c == ">" then "&gt;"
+      elsif c == "\"" then "&quot;"
+      elsif c == "'" then "&apos;"
+      else c
+      end
+    end,
+    s
+  )
+end
--- a/src/libs/extra/interactive.liq
+++ b/src/libs/extra/interactive.liq
@@ -524,29 +524,13 @@ def interactive.harbor(
     add("</head><body>")
     add("<h1>#{title}</h1>")
 
-    def escape_html(s) =
-      string.replace(
-        pattern = "[&<>\"']",
-        fun (c) -> begin
-          if c == "&" then "&amp;"
-          elsif c == "<" then "&lt;"
-          elsif c == ">" then "&gt;"
-          elsif c == "\"" then "&quot;"
-          elsif c == "'" then "&apos;"
-          else c
-          end
-        end,
-        s
-      )
-    end
-
     def add_var(nv) =
       let (name, v) = nv
       description = interactive.description(name)
       description =
         if description == "" then name else "#{description} (#{name})" end
 
-      add("<label for=#{name}>#{escape_html(description)}</label>")
+      add("<label for=#{name}>#{string.escape.html(description)}</label>")
       common =
         "id='#{name}' name='#{name}' class='interactive' onchange=\"send()\""
 
@@ -587,7 +571,7 @@ def interactive.harbor(
         v.type == "string"
       then
         v = list.assoc(name, variables_string())
-        add("<input type='text' #{common} value='#{escape_html(v.ref())}'>")
+        add("<input type='text' #{common} value='#{string.escape.html(v.ref())}'>")
       elsif
         v.type == "unit"
       then
--- a/src/libs/extra/interactive.liq
+++ b/src/libs/extra/interactive.liq
@@ -495,7 +495,7 @@ def interactive.harbor(
           }
         }
         if (interactive[i].type != 'button') {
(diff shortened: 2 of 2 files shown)


smimram:
Thanks for this PR! It would be nice if `escape_html` could be moved to `string.liq` and renamed to `string.escape.html` since it could be useful outside of this context.

ghostnumber7:
> Thanks for this PR! It would be nice if `escape_html` could be moved to `string.liq` and renamed to `string.escape.html` since it could be useful outside of this context.

Moved. I'm not sure about the "docs" entry tho.

src/libs/string.liq
```diff
@@ -381,3 +381,21 @@ let string.char.ascii.number = list.init(10, fun (c) -> c + 48)
...
+# @category String
+def string.escape.html(s) =
+  string.replace(
+    pattern = "[&<>\"']",
+    fun (c) -> begin
+      if c == "&" then "&amp;"
+      elsif c == "<" then "&lt;"
+      elsif c == ">" then "&gt;"
+      elsif c == "\"" then "&quot;"
+      elsif c == "'" then "&apos;"
```
vitoyucepi, line 395: Is it safe to use `&apos;` instead of `&#39;`?

src/libs/string.liq
```diff
@@ -381,3 +381,21 @@ let string.char.ascii.number = list.init(10, fun (c) -> c + 48)
...
+# Escape HTML entities.
+# @category String
+def string.escape.html(s) =
+  string.replace(
+    pattern = "[&<>\"']",
+    fun (c) -> begin
+      if c == "&" then "&amp;"
+      elsif c == "<" then "&lt;"
+      elsif c == ">" then "&gt;"
+      elsif c == "\"" then "&quot;"
```
vitoyucepi, line 394:
```suggestion
      elsif c == '"' then "&quot;"
```
I think it's OK to use single quotes to prevent escaping, but the linter may not be happy.

src/libs/string.liq
```diff
@@ -381,3 +381,21 @@ let string.char.ascii.number = list.init(10, fun (c) -> c + 48)
...
+      elsif c == "'" then "&apos;"
+      else c
+      end
+    end,
+    s
+  )
```
vitoyucepi, line 400:
```suggestion
  r/[&<>"']/g.replace(
    fun (c) -> begin
      if c == "&" then "&amp;"
      elsif c == "<" then "&lt;"
      elsif c == ">" then "&gt;"
      elsif c == '"' then "&quot;"
      elsif c == "'" then "&#39;"
      else c
      end
    end,
    s
  )
```
I'd prefer to use [`r/_/g.replace(fn, str)`](https://www.liquidsoap.info/doc-dev/language.html#regular-expressions) regular expression syntax.

ghostnumber7: the `&#` aproach is actually shorter to implement if I can get the char code in any way (and would use it for all of them in that case), but could not find a way to do that. Is there any?

vitoyucepi:
Could you add some [tests](https://github.com/savonet/liquidsoap/blob/main/tests/language/string.liq#L115)?

```ruby
def test_escape_html() =
  test.equals(string.escape.html("&"), "&amp;")
  test.equals(string.escape.html("<"), "&lt;")
  test.equals(string.escape.html(">"), "&gt;")
  test.equals(string.escape.html('"'), "&quot;")
  test.equals(string.escape.html("'"), "&#39;")
  test.equals(string.escape.html("&<>\"'"), "&amp;&lt;&gt;&quot;&#39;")
  test.equals(string.escape.html("not escaped"), "not escaped")
  test.equals(string.escape.html("&quot;double escape&quot;"), "&amp;quot;double escape&amp;quot;")
  test.equals(string.escape.html("\\"), "\\")
  test.equals(string.escape.html("/"), "/")
  test.equals(string.escape.html("`"), "`")
end

test.check(test_escape_html)
```

src/libs/string.liq
ghostnumber7: It should be the same

ghostnumber7:
> Could you add some [tests](https://github.com/savonet/liquidsoap/blob/main/tests/language/string.liq#L115)?
> 
> ```ruby
> def test_escape_html() =
>   test.equals(string.escape.html("&"), "&amp;")
>   test.equals(string.escape.html("<"), "&lt;")
>   test.equals(string.escape.html(">"), "&gt;")
>   test.equals(string.escape.html('"'), "&quot;")
>   test.equals(string.escape.html("'"), "&#39;")
>   test.equals(string.escape.html("&<>\"'"), "&amp;&lt;&gt;&quot;&#39;")
>   test.equals(string.escape.html("not escaped"), "not escaped")
>   test.equals(string.escape.html("&quot;double escape&quot;"), "&amp;quot;double escape&amp;quot;")
>   test.equals(string.escape.html("\\"), "\\")
>   test.equals(string.escape.html("/"), "/")
>   test.equals(string.escape.html("`"), "`")
> end
> 
> test.check(test_escape_html)
> ```

might be a dumb question, but how do I run tests locally to see they are working? :)

ghostnumber7:
@vitoyucepi how about doing this instead of the current approach? It should work just the same 

```
# Escape HTML entities.
# @category String
def string.escape.html(s) =
   r/[&<>"']/g.replace(
    fun (c) -> "&##{string.binary.to_int(c)};",
    s
  )
end
```

vitoyucepi:
> might be a dumb question, but how do I run tests locally to see they are working? :)

Sorry, I haven't yet figured out how to build liquidsoap from the local source using dune.
I pretend everything works fine and wait for the CI results.

PR #3436, opened 2023-10-02
test: Update deprecated reference notation

vitoyucepi:
## Summary

Update references in tests to remove warnings about deprecated functions. Use `r()` instead of the deprecated `!r` to get the value of ref type variables.


toots:
Thanks!

PR #3437, opened 2023-10-02
test: Add forgotten `test.pass()` call to the string tests.

vitoyucepi:
## Summary

The function `test_escape_html()` has no `test.pass()` call at the end.


toots:
Thanks!

PR #3452, opened 2023-10-06
Use getter arguments in `blank.detect`.

smimram:
Fixes #3449.


toots:
Thanks! I think I'll backport this one to `rolling-release-v2.2.x`

toots (approved):

PR #3478, opened 2023-10-18
feat: Add new assertion functions

vitoyucepi:
## Summary
This PR adds the following new assertion functions to the test module:

- `test.not_equals(first, second)` - Asserts two values are not equal.
- `test.almost_equals(~places=7, first, second)` - Asserts two values are almost equal.
- `test.not_almost_equals(~places=7, first, second)` - Asserts two values are not almost equal.

Additionally, tests in `language/math.liq` and `streams/replaygain.liq` has been rewritten using these new assertions.

## Details

The new assertion functions are necessary to provide additional validation capabilities beyond just `test.equals`:

- `test.not_equals(first, second)`
   This allows asserting two values are not equal, which is not possible with just `test.equals`.
- `test.almost_equals(~places=7, first, second)`
  `test.equals` only does an exact match, but `almost_equals` allows approximately comparing floats within a precision.
- `test.not_almost_equals(~places=7, first, second)`
  Similar to `almost_equals`, this provides a way to validate two floats are not approximately equal within a precision tolerance.

### DIFF
--- a/tests/test.liq
+++ b/tests/test.liq
@@ -111,6 +111,71 @@ def test.equals(v, v') =
   end
 end
 
+def test.not_equals(first, second) =
+  if
+    first == second
+  then
+    error.raise(
+      error.failure,
+      "expected\r\n
+      #{string.quote(string(first))} == #{string.quote(string(second))}\r\n
+      to be false.\r\n"
+    )
+
+    test.fail()
+  end
+end
+
+# Compares two float values for approximate equality.
+# Values are considered almost equal if their difference
+# is within 10^-digits.
+# @param ~digits The number of decimal digits to compare. Default is 7.
+# @flag hidden
+def test._almost_equals(~digits=7, first, second) =
+  diff = abs(float(first) - float(second))
+  if
+    diff * float(pow(10, digits)) < 0.5
+  then
+    true.{diff=diff}
+  else
+    false.{diff=diff}
+  end
+end
+
+# Compares two float values for approximate equality.
+# Values are considered almost equal if their difference
+# is within 10^-digits.
+# @param ~digits The number of decimal digits to compare. Default is 7.
+def test.almost_equals(~digits=7, first, second) =
+  is_almost_equals = test._almost_equals(digits=digits, first, second)
+  if not is_almost_equals then
+    error.raise(
+      error.failure,
+      "#{string.quote(string(first))} != #{string.quote(string(second))}
+       up to #{digits} digits, diff = #{is_almost_equals.diff}."
+    )
+
+    test.fail()
+  end
+end
+
+# Compares two float values for approximate inequality.
+# Values are considered not almost equal if their difference
+# is greater than 10^-digits.
+# @param ~digits The number of decimal digits to compare. Default is 7.
+def test.not_almost_equals(~digits=7, first, second) =
+  is_almost_equals = test._almost_equals(digits=digits, first, second)
+  if is_almost_equals then
+    error.raise(
+      error.failure,
+      "#{string.quote(string(first))} == #{string.quote(string(second))}
+       up to #{digits} digits, diff = #{is_almost_equals.diff}."
+    )
+
+    test.fail()
+  end
+end
+
 def test.raises(fn, e=null()) =
   try
     ignore(fn())
--- a/tests/language/math.liq
+++ b/tests/language/math.liq
@@ -1,23 +1,12 @@
-success = ref(true)
+#!../../liquidsoap ../test.liq
 
-def t(x, y) =
-  if
-    x != y
-  then
-    print(
-      "Failure: got #{x} instead of #{y}"
-    )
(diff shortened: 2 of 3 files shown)


tests/test.liq
```diff
@@ -111,6 +111,51 @@ def test.equals(v, v') =
...
+      "expected\r\n
+      #{string.quote(string(first))} == #{string.quote(string(second))}\r\n
+      to be not equal.\r\n"
+    )
+
+    test.fail()
+  end
+end
+
+def test.almost_equals(~places=7, first, second) =
```
toots, line 129: I think `digits` would be more clear as it specifically names what is being checked. Some comment too:
```suggestion
# Test float equality
# @param ~digits Test equality up-to the given number of decimal digits
def test.almost_equals(~digits=7, first, second) =
```

tests/test.liq
```diff
@@ -111,6 +111,51 @@ def test.equals(v, v') =
...
+      error.failure,
+      "#{string.quote(string(first))} != #{string.quote(string(second))}
+       up to #{places} digits, diff = #{diff}."
+    )
+
+    test.fail()
+  end
+end
+
+def test.not_almost_equals(~places=7, first, second) =
```
toots, line 144: Let's define this using the previous one so as to not duplicate code logic and potential bugs.

```suggestion
def test.not_almost_equals(~digits=7, first, second) =
  not test.almost_equals(digits=digits, first, second)
```

toots (commented):
Looks great thanks for doing this! Some comments on naming. I'm gonna fix the metrics upload on citest.

tests/test.liq
vitoyucepi: How about
```ruby
# Compares two float values for approximate equality.
# Values are considered almost equal if their difference
# is within 10^-digits.
# @param ~digits The number of decimal digits to compare. Default is 7.
def test.almost_equals(~digits=7, first, second) =
```

toots: Works for me!

tests/test.liq
toots: Ha word. Yeah that works. Unfortunately, `%argsof` cannot be used with functions from the same script file currently..

tests/test.liq
```diff
@@ -111,6 +111,51 @@ def test.equals(v, v') =
...
 
+def test.not_equals(first, second) =
+  if
+    first == second
+  then
+    error.raise(
+      error.failure,
+      "expected\r\n
+      #{string.quote(string(first))} == #{string.quote(string(second))}\r\n
+      to be not equal.\r\n"
```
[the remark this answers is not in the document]
vitoyucepi: Done

tests/test.liq
vitoyucepi: Done

tests/test.liq
vitoyucepi: I'm not fully understand this suggestion.

I changed
```ruby
def test.not_almost_equals(~places=7, first, second) =
  diff = abs(float(first) - float(second))
  if
    diff * float(pow(10, places)) <= 0.5
  then
    error.raise(
      error.failure,
      "#{string.quote(string(first))} == #{string.quote(string(second))}
       up to #{places} digits."
    )

    test.fail()
  end
end
```
to
```ruby
def test.not_almost_equals(~digits=7, first, second) =
  diff = abs(float(first) - float(second))
  if
    not test.almost_equals(digits=digits, first, second)
  then
    error.raise(
      error.failure,
      "#{string.quote(string(first))} == #{string.quote(string(second))}
       up to #{digits} digits, diff = #{diff}."
    )

    test.fail()
  end
end
```

vitoyucepi: @toots, maybe if I introduce the hidden function that does the math

```ruby
# @flag hidden
def test._almost_equals(~digits=7, first, second) =
  diff = abs(float(first) - float(second))
  if
    diff * float(pow(10, places)) < 0.5
  then
    true.{diff=diff}
  else
    false.{diff=diff}
  end
end
```
then I could reuse it in my code
```ruby
def test.almost_equals(~digits=7, first, second)
  is_almost_equals = test._almost_equals(digits=digits, first, second)
  if not is_almost_equals then
    error.raise(
      error.failure,
      "#{string.quote(string(first))} == #{string.quote(string(second))}
       up to #{digits} digits, diff = #{is_almost_equals.diff}."
    )
    test.fail()
  end
end

def test.not_almost_equals(~digits=7, first, second)
  is_almost_equals = test._almost_equals(digits=digits, first, second)
  if is_almost_equals then
    error.raise(
      error.failure,
      "#{string.quote(string(first))} == #{string.quote(string(second))}
       up to #{digits} digits, diff = #{is_almost_equals.diff}."
    )
    test.fail()
  end
end
```

toots: Yes definitely. Also since the test library is not used to generate the doc you don't have to worry to much about hiding it.

PR #3522, opened 2023-11-12
Add an option to disable fallibility check.

smimram:
This is sometimes useful, especially when checking scripts in examples where we don't have the files...


toots (approved):
Looks great thanks!

Issue #3303, opened 2023-08-07
Different http handlers may cause typing errors


vitoyucepi:
**Describe the bug**
If I use different handlers with different routes they may trigger typing exception.

<details>
<summary>Exception</summary>

```
At 1.liq, line 9, char 0-0:

Error 9: Failure: invalid sup: {a : 'A?} !< {a? : 'A?, b? : 'B?} (Typing error: _.{a : 'A} vs _.{a? : 'A?})
Raised at Stdlib.failwith in file "stdlib.ml", line 29, characters 17-33
Called from Liquidsoap_lang__Typing.(<:) in file "src/lang/typing.ml", line 427, characters 13-164
Called from Liquidsoap_lang__Typing.(<:).(fun) in file "src/lang/typing.ml", line 522, characters 20-27
Called from Stdlib__List.fold_left in file "list.ml", line 121, characters 24-34
Called from Liquidsoap_lang__Typing.(<:) in file "src/lang/typing.ml", line 497, characters 12-1023
Called from Liquidsoap_lang__Typing.unify_meth in file "src/lang/typing.ml", line 360, characters 7-63
Called from Liquidsoap_lang__Typing.(<:).(fun) in file "src/lang/typing.ml", line 522, characters 20-27
Called from Stdlib__List.fold_left in file "list.ml", line 121, characters 24-34
Called from Liquidsoap_lang__Typing.(<:) in file "src/lang/typing.ml", line 497, characters 12-1023
Called from Liquidsoap_lang__Typing.(<:) in file "src/lang/typing.ml", line 639, characters 6-12
Called from Liquidsoap_lang__Typechecking.check.(fun) in file "src/lang/typechecking.ml", line 315, characters 37-45
Called from Stdlib__List.fold_left in file "list.ml", line 121, characters 24-34
Called from Liquidsoap_lang__Typechecking.check.(fun) in file "src/lang/typechecking.ml", line 299, characters 18-990
Called from Liquidsoap_lang__Typechecking.check.(fun).check in file "src/lang/typechecking.ml", line 141, characters 4-39
Called from Liquidsoap_lang__Typechecking.check.(fun) in file "src/lang/typechecking.ml", line 273, characters 10-45
Called from Liquidsoap_lang__Typechecking.check.(fun).check in file "src/lang/typechecking.ml", line 141, characters 4-39
Called from Liquidsoap_lang__Typechecking.check.(fun) in file "src/lang/typechecking.ml", line 389, characters 10-48
[...]

PR #3524, opened 2023-11-12
Add univ type annotation

toots:
I was looking at https://github.com/savonet/liquidsoap/issues/3303 and started wondering about type annotation for universal variables.

This PR changes the following thing:
* Switch type annotation for universal variables from `'a` a-la OCaml to `univ(a)`. I think using `univ` is much more user-friendly.
* Add support for universal variables type annotation. Once we switch to `univ`, it becomes fully parsable! I'm not sure if this has a huge impact but it's nice to have.
* Cleanup some old code.


smimram:
It looks nice, but I am afraid about readability: can you show examples of types functions with many variables (e.g. `ffmpeg.filter.audio_video.output`), is it still readable?...

toots:
Looks pretty good to me:

Screenshot 2023-11-15 at 1 12 38 PM

toots:
We can revisit later when/if we want to add a more generic type declaration layer.

PR #3589, opened 2023-12-13
Allow bigarray for string value, use it to store large metadata.

toots:


toots:
Turned out the memleak was completely unrelated. Phfew! Glad we had cold feet with this one..

PR #3590, opened 2023-12-13
Add support for bigstring.

toots:
This PR adds support for bigstring to the language. Bigstrings are represented by bigarrays and can also potentially be `mmap`'d data coming directly from a file.

Its expected use is to handle large strings to be written and read from different sources, in particular:
* Socket
* Files
* Metadata (cover art)

### DIFF
--- a/.github/opam/liquidsoap-core-windows.opam
+++ b/.github/opam/liquidsoap-core-windows.opam
@@ -28,6 +28,8 @@ depends: [
   "sedlex-windows" {>= "3.2"}
   "menhir"
   "menhirLib-windows"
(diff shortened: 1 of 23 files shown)


toots:
The original issue this PR was designed to address (supposed memory accumulation due to cover files being large strings) was actually unrelated.

It would be great to have a notion of big string to be able to e.g. do `file.contents("large.file")` and have this not live in memory but this has some serious potential issues. For instance, any operation that needs the string in memory would result in a memory explosion without having clear visibility about it.

Thus, closing for now. Still something to think of but, perhaps. with a `bigstring` type separate from the main` string` type.

PR #3597, opened 2023-12-15
Add support for chatGPT

smimram:


src/libs/extra/openai.liq
```diff
@@ -0,0 +1,73 @@
...
+      }
+    ) = ans
+    ans
+  catch _ : [error.json] do
+    let json.parse (e :
+      {error: {message: string, type: string, code: string}}
+    ) = ans
+    e = e.error
+    error.raise(
+      error.openai,
```
toots, line 42: We need to figure out the pattern for error namespaces. I think they shouldn't be too specific like this?

src/libs/extra/openai.liq
```diff
@@ -0,0 +1,73 @@
...
+          }
+        ],
+        created: int,
+        model: string,
+        object: string,
+        usage: {completion_tokens: int, prompt_tokens: int, total_tokens: int}
+      }
+    ) = ans
+    ans
+  catch _ : [error.json] do
```
toots, line 36: This is not just gonna be `error.json`, curl can also return its own errors.

src/libs/extra/openai.liq
```diff
@@ -0,0 +1,73 @@
...
+  ~model="gpt-3.5-turbo",
+  ~system="You are a helpful assistant.",
+  prompt
+) =
+  messages = [{role="system", content=system}, {role="user", content=prompt}]
+  ans = chatgpt(key=key, model=model, messages)
+  ans
+end
+
+openai = ()
```
toots, line 60: Why is there both a `openai` and `chatgpt` namepsace? Shouldn't `chatgpt` be under `openai`?

toots (commented):
Looking good! Some initial questions..

toots:
Merging this. Thanks for the contribution!

PR #3601, opened 2023-12-20
Add parents option to `file.mkdir`.

### LINKED ISSUE #3600: file.mkdir to support "recursive" dir creation / parent dir creation
`file.mkdir` doesn't support "recursive"/parent dir creation similar to the linux command "mkdir -p xxx". It would be useful for it to support this functionallity.

```
newfolder = "/srv/hlschunks/#{datetime.year}/#{datetime.month}/#{datetime.day}/#{datetime.hour}/#{datetime.min}"

//Current Method
ignore(process.run("mkdir -p #{newfolder}"))

// Ideal Method
file.mkdir(newfolder)
```

smimram:
Fixes #3600.

### DIFF
--- a/src/core/builtins/builtins_files.ml
+++ b/src/core/builtins/builtins_files.ml
@@ -129,8 +129,9 @@ let _ =
       let dir = List.assoc "" p |> Lang.to_string in
       try
         let rec recmkdir dir =
-          if not (Sys.file_exists dir) then recmkdir (Filename.dirname dir);
-          Unix.mkdir dir perms
+          if not (Sys.file_exists dir) then (
+            recmkdir (Filename.dirname dir);
+            Unix.mkdir dir perms)
         in
         let mkdir dir =
           if parents then recmkdir dir else Unix.mkdir dir perms
--- a/src/core/builtins/builtins_files.ml
+++ b/src/core/builtins/builtins_files.ml
@@ -133,10 +133,7 @@ let _ =
             recmkdir (Filename.dirname dir);
             Unix.mkdir dir perms)
         in
-        let mkdir dir =
-          if parents then recmkdir dir else Unix.mkdir dir perms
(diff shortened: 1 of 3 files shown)


src/core/builtins/builtins_files.ml
```diff
@@ -120,10 +124,16 @@ let _ =
...
       try
-        Unix.mkdir dir perms;
+        let rec recmkdir dir =
+          if not (Sys.file_exists dir) then (
+            recmkdir (Filename.dirname dir);
+            Unix.mkdir dir perms)
+        in
+        if parents then recmkdir dir else Unix.mkdir dir perms;
         Lang.unit
       with _ -> Lang.unit)
```
toots, line 138: Should we raise a runtime error instead?

src/core/builtins/builtins_files.ml
```diff
@@ -112,6 +112,10 @@ let _ =
   Lang.add_builtin ~base:file "mkdir" ~category:`File
     ~descr:"Create a directory."
     [
+      ( "parents",
```
toots, line 115: I think `recursive` is a more common term, that's also the name of the command line option.

toots (approved):
Looking good! Two comments.

src/core/builtins/builtins_files.ml
toots: Hmm no my bad I got confused with `rm`.

src/core/builtins/builtins_files.ml
smimram: Let's keep the simple behavior for now.

scottgrobinson:
Hey @toots @smimram - Sorry for the tag on this. Any chance we can look at merging this into 2.2.5? It missed 2.2.4 (I think?)
Thank you!

PR #3603, opened 2023-12-21
Add hex and oct integers

toots:
This PR adds octal and hexadecimal ground terms so that they can be represented properly in the doc!

Screenshot 2023-12-21 at 10 41 46 AM


smimram (approved):
Nice tiny PR, go for it!

PR #3620, opened 2024-01-09
Add operation to generate CUE files.

### LINKED ISSUE #3608: Possible to record live stream including a `.cue` file?
For recording DJ live streams, it would be advantageous being able to not only _record_ the live stream, but also get a `.cue` file together with the recording. **Would this be possible?**

I think LS would be the logical place to generate a `.cue` file, since it knows date, time, and all metadata while playing out.

Reasons:
- Easier reporting what was broadcast, and when.
- Seeing (and jumping to) the titles in later playback of the recording.
- Auto-generating a "show" textual representation for station forums, etc. (What was played at which date/time.)

When using [IDJC](https://idjc.sourceforge.io/) as streaming software, we can already have that (generated by IDJC), and it would be great if something like this would be possible in Liquidsoap, too.

IDJC outputs a recording of what was broadcast with a configurable name, and includes the metadata (and start of broadcast) in the corresponding cue file.

**File example:**

```
aufzeichnung.2019-06-24.160520-r01.cue
aufzeichnung.2019-06-24.160520-r01.mp3
```

**`.cue` file content (abbreviated):**

```cue
TITLE "[2019-06-24][16:05:20]"
PERFORMER "Recorded with IDJC"
FILE "aufzeichnung.2019-06-24.160520-r01.mp3" MP3
  TRACK 01 AUDIO
    TITLE "A.C.D.C. (2006)"
    PERFORMER "Joan Jett and the Blackhearts"
    REM ALBUM "Sinner"
    INDEX 01 00:00:00
  TRACK 02 AUDIO
    TITLE "No Metal, all Rock: 13 Jahre - 13 Titel"
    PERFORMER "Moonbase"
    INDEX 01 00:01:07
  TRACK 03 AUDIO
    TITLE "A.C.D.C. (2006)"
    PERFORMER "Joan Jett and the Blackhearts"
    REM ALBUM "Sinner"
    INDEX 01 01:27:58
  TRACK 04 AUDIO
    TITLE "Runnin' Wild (2007)"
    PERFORMER "Airbourne"
    REM ALBUM "Runnin' Wild"
    INDEX 01 04:48:50
```

smimram:
Fixes #3608.


src/libs/list.liq
```diff
@@ -198,6 +198,28 @@ def list.assoc(~default=null(), key, l) =
...
+  end
+end
+
+# Associate a value to a key in an association list. This functions is similar
+# to `list.assoc` excepting that it returns `null` if no value exists for the
+# key.
+# @category List
+def list.assoc.nullable(key, l) =
+  if list.assoc.mem(key, l) then list.assoc(key, l) else null() end
+end
```
toots, line 221: nit: you could use a `try ... catch` to avoid enumerating the list twice..

src/libs/source.liq
```diff
@@ -255,3 +255,48 @@ def prepend(~id=null("prepend"), ~merge=false, s, f) =
   d = source.dynamic(track_sensitive=true, next)
   fallback(id=id, track_sensitive=true, [d, s])
 end
+
+stdlib_file = file
+
+# Generate a CUE file for the source.
+# @category Source / Track processing
+# @param filename Path where the CUE file should be written.
+def source.cue(s, filename, ~title=null(), ~performer=null(), ~file=null())
```
toots, line 264: Could we move this to `src/libs/extra`?

toots (commented):
I've been waiting for this! Thanks!

Couple of remarks:
* I think this should go on the extra lib
* Are you interested in extracting the time split logic into its own function (provided we don't have one already!)
* Can we add test to see what the file looks like?
* Can we add parameters to make it work with infinite sources? Like keeping the last `n` entries? Could be disabled by default.
* Can we make file change atomic? Otherwise you will run into situations where a partial file might be read. This happened with HLS sources. You could factor out this code from `output.hls`:
```ocaml
          Fun.protect
            ~finally:(fun () -> try Sys.remove tmp_file with _ -> ())
            (fun () ->
              (try Unix.rename tmp_file filename
               with Unix.Unix_error (Unix.EXDEV, _, _) ->
                 self#log#important
                   "Rename failed! Directory for temporary files appears to be \
                    on a different file system. Please set it to the same one \
                    using `temp_dir` argument to guarantee atomic file \
                    operations!";
                 Utils.copy ~mode ~perms tmp_file filename;
                 Sys.remove tmp_file);
```

smimram:
I think I have taken most of your suggestions in account. I propose that we merge as is and improve if usecases do show up...

src/libs/source.liq
Moonbase59, line 1: @smimram: Looks like you assume hours:minutes:seconds for the INDEX times, but they are actually `MM:SS:FF` (minutes, seconds, frames). Hydrogenaudio tells us there are 75 frames to a second: https://wiki.hydrogenaud.io/index.php?title=Cue_sheet#Most_often_used

Moonbase59:
@toots @smimram Please check https://github.com/savonet/liquidsoap/pull/3620/commits/50da0a300477652ed429d9c456b304889f7bee63#r1448803161 (`MM:SS:FF` INDEX times).

src/libs/source.liq
Moonbase59: Frames are numbered 00…74.

PR #3633, opened 2024-01-13
More improvements on `source.cue`.

smimram:


Moonbase59:
Thanks for following up on this! As soon as we have a new `.deb` (Ubuntu Jammy), I’ll do more testing.

- Using a `map_metadata` seems a good idea. So we can modify metadata before writing the `.cue` sheet any way needed.

- DATE still seems an int? We should be able to support things like `YYYY`, `YYYY-MM` and `YYYY-MM-DD`.

- I found out that `max_length` leaves only the last "n" tracks in the CUE sheet, numbered track 01 .. n, but with correct INDEX 01 offsets. What’s the use case for this?

toots:
You beat me to it but also, this needs more tests! Gonna send a follow-up!

PR #3634, opened 2024-01-13
Cue more!

toots:
More stuff for `cue` after @Moonbase59 remarks
* Rename `max_length` to `max_tracks`. Even with the documentation, the parameter name is confusing.
* Switch `data` to a un-escaped string to allow more flexibility w.r.t. the spec.
* Switch back to metadata for detecting new track marks. In most cases this will do the same but I think that this is closest to what people will expect in most situations.
* Add a `deduplicate_using` to deduplicate metadata based on title/artist/album.
* Add tests!

### DIFF
--- a/src/libs/extra/source.liq
+++ b/src/libs/extra/source.liq
@@ -351,7 +351,7 @@ def source.cue(
         null.defined(date)
       then
         w(
-          "DATE #{(null.get(date) : int)}"
+          "DATE #{(null.get(date) : string)}"
         )
       end
       if
--- a/src/libs/extra/source.liq
+++ b/src/libs/extra/source.liq
@@ -262,17 +262,19 @@ end
 stdlib_file = file
 
 # Generate a CUE file for the source. This function will generate a new track in
-# the file for each metadata of the source.
+# the file for each metadata of the source. This function tries to map metadata to
+# the appropriate CUE file standard values. You can use the `map_metadata` argument
+# to add your own pre-processing. The following metadata are recognized on tracks:
+# `"title"`, `"artist"`, `"album"`, `"isrc"`, and `"cue_year"`.
 # @category Source / Track processing
 # @flag extra
 # @param filename Path where the CUE file should be written.
-# @param max_tracks Set the maximum of track info retained by the function. \
-#                   Infinite when `null`.
+# @param ~last_tracks Only report the number of last tracks.
 # @param ~title Title of the stream.
 # @param ~file File where the stream is stored.
 # @param ~file_type Format in which the stream is stored.
 # @param ~comment Comment about the stream.
-# @param ~date Year for the stream.
+# @param ~year Year for the stream.
 # @param ~map_metadata Function to apply to metadata before writing the CUE file (useful for pre-processing metadata).
 # @param ~temp_dir Temporary directory for atomic write.
 # @param ~deduplicate_using To avoid duplicate entries, duplicate metadata are \
@@ -283,13 +285,13 @@ def source.cue(
   ~title=null(),
   ~performer=null(),
   ~file=null(),
-  ~file_type="mp3",
+  ~file_type=null(),
   ~comment=null(),
-  ~date=null(),
+  ~year=null(),
   ~map_metadata=fun (m) -> (m : [(string*string)]),
-  ~max_tracks=null(),
+  ~last_tracks=null(),
   ~temp_dir=null(),
-  ~deduplicate_using=["title", "artist", "album"],
+  ~deduplicate_using=["title", "artist", "album", "isrc", "cue_year"],
   ~delete=true,
   filename,
   s
@@ -299,6 +301,8 @@ def source.cue(
   then
     stdlib_file.remove(filename)
   end
+
+  file_type = file_type ?? stdlib_file.extension(leading_dot=false, file ?? "")
 
   is_first = ref(true)
 
@@ -330,12 +334,14 @@ def source.cue(
     end
 
     # Write a tag.
-    def tag(~indent=0, name, (value:string?)) =
+    def tag(~indent=0, ~quote=true, name, (value:string?)) =
+      quote = if quote then fun (v) -> string.quote(v) else fun (v) -> v end
+
       if
         null.defined(value)
       then
         s =
-          "#{string.spaces(indent)}#{name} #{string.quote(null.get(value))}"
+          "#{string.spaces(indent)}#{name} #{quote(null.get(value))}"
         w(s)
       end
     end
@@ -351,13 +357,11 @@ def source.cue(
         "REM COMMENT",
         comment
       )
-      if
-        null.defined(date)
-      then
-        w(
-          "DATE #{(null.get(date) : string)}"
-        )
-      end
+      tag(
+        quote=false,
+        "REM DATE",
+        null.map(string.of_int, year)
+      )
       if
         null.defined(file)
       then
@@ -372,8 +376,11 @@ def source.cue(
         begin
           let {position = p, time = t, metadata = m} = entry
 
-          w(
-            "  TRACK #{string.of_int(digits=2, p)} AUDIO"
+          tag(
+            indent=2,
+            quote=false,
+            "TRACK",
+            "#{string.of_int(digits=2, p)} AUDIO"
           )
           tag(indent=4, "TITLE", list.assoc.nullable("title", m))
           tag(indent=4, "PERFORMER", list.assoc.nullable("artist", m))
@@ -382,6 +389,13 @@ def source.cue(
             "REM ALBUM",
             list.assoc.nullable("album", m)
           )
+          tag(
+            indent=4,
+            quote=false,
+            "REM DATE",
+            list.assoc.nullable("cue_year", m)
+          )
+          tag(indent=4, quote=false, "ISRC", list.assoc.nullable("isrc", m))
 
           frames = int_of_float((t - floor(t)) * 75.)
           t = int_of_float(t)
@@ -390,8 +404,11 @@ def source.cue(
           m = string.of_int(digits=2, minutes)
           s = string.of_int(digits=2, seconds)
           f = string.of_int(digits=2, frames)
-          w(
-            "    INDEX 01 #{m}:#{s}:#{f}"
+          tag(
+            indent=4,
+            quote=false,
+            "INDEX 01",
+            "#{m}:#{s}:#{f}"
           )
         end,
       entries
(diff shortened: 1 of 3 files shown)


Moonbase59:
Phew, this goes fast—everyone working instead of clubbing on a Saturday night!

On LS 2.3.0+git@4a2d37e37, when using a stream as input, the TITLEs were missing:

```
TITLE "[2024-01-13][23:32:00]"
PERFORMER "Recorded with Liquidsoap"
REM COMMENT "http://server14613.streamplus.de:35194/;.mp3"
FILE "aufzeichnung.2024-01-13.233200.mp3" MP3
  TRACK 01 AUDIO
    INDEX 01 00:00:00
  TRACK 02 AUDIO
    INDEX 01 00:04:24
  TRACK 03 AUDIO
    INDEX 01 00:04:27
  TRACK 04 AUDIO
    INDEX 01 03:40:60
  TRACK 05 AUDIO
    INDEX 01 03:55:09
  TRACK 06 AUDIO
    INDEX 01 03:55:11
```

toots:
> Phew, this goes fast—everyone working instead of clubbing on a Saturday night!

It's still early here!

> On LS 2.3.0+git@[4a2d37e](https://github.com/savonet/liquidsoap/commit/4a2d37e37b3ab3a19fb5396e28d59862f1560eff), when using a stream as input, the TITLEs were missing:
> 
> ```
> TITLE "[2024-01-13][23:32:00]"
> PERFORMER "Recorded with Liquidsoap"
> REM COMMENT "http://server14613.streamplus.de:35194/;.mp3"
> FILE "aufzeichnung.2024-01-13.233200.mp3" MP3
>   TRACK 01 AUDIO
>     INDEX 01 00:00:00
>   TRACK 02 AUDIO
>     INDEX 01 00:04:24
>   TRACK 03 AUDIO
>     INDEX 01 00:04:27
>   TRACK 04 AUDIO
>     INDEX 01 03:40:60
>   TRACK 05 AUDIO
>     INDEX 01 03:55:09
>   TRACK 06 AUDIO
>     INDEX 01 03:55:11
> ```
> 
> Stream definitely has metadata, example:
> 
> ```
> 2024/01/13 23:36:01 [nowplaying:4] META: [("icy-br", "128"), ("icy-genre", "Rock, Hardrock, Metal"), ("icy-name", "Radio Paranoid - Skyrider - Nightflight 666"), ("icy-notice1", "<BR>This stream requires <a href=\"http://www.winamp.com/\">Winamp</a><BR>"), ("icy-notice2", "SHOUTcast Distributed Network Audio Server/Linux v1.9.8<BR>"), ("icy-pub", "1"), ("icy-url", "http://www.radio-paranoid.net/"), ("title", "Dio - Caught in the Middle"), ("url", "http://www.audiorealm.com")]
> ```

Yeah. Track and metadata has been rewritten in `main` so you should expect this to need a little more time to adjust. I'd suggest testing my branch where I've switched back to using metadata and added a metadata de-duplicator.

Moonbase59:
The "map_metadata" function also receives no metadata (just [] when I log m).

Your branch: That’d be `cue-more`?
Does that generate a `.deb` so I can test?

Moonbase59:
Using your branch did it, thanks for the hint.

Now we’re talking:

```
TITLE "[2024-01-14][00:50:14]"
PERFORMER "Recorded with Liquidsoap"
REM COMMENT "https://radio.niteradio.net/listen/niteradio/radio.mp3"
DATE 2024-01-14
FILE "aufzeichnung.2024-01-14.005014.mp3" MP3
  TRACK 01 AUDIO
    TITLE "All Nights"
    PERFORMER "Asuntar"
    INDEX 01 00:09:15
```

DATE working (have to look up if it should be `DATE` or `REM DATE` on the album level).
Metadata mapping working (splitting the TITLE to TITLE and PERFORMER).

Could this super-hacky brute-force metadata mapping be written more elegantly?

```
def mapmeta(m)
  artist = ref(m["artist"])
  title = ref(m["title"])
  #log(level=4, label="cue", "META: #{metadata.cover.remove(m)}")
  log(level=4, label="cue", "META: #{m}")
  if string.length(m["artist"]) <= 0 then
    if string.contains(substring=" - ", m["title"]) then
      let (a, t) = string.split.first(separator=" - ", m["title"])
      artist := a
      title := t
      log(level=4, "a=#{a}, t=#{t}")
    end
  end
  [("title", "#{title()}"), ("artist", "#{artist()}")]
end
```

Moonbase59:
@toots Did some more testing re DATE on CD and TRACK level. It seems most cue sheet parsers recognise `REM DATE YYYY`—on both CD and TRACK level, but only a very few accept `DATE …`. Also only a very few understand the `YYYY-MM` and `YYYY-MM-DD` formats (probably due to ID3v2 initially having only YEAR before we got the more detailed date formats). Tested with audio & video players, cue sheet cutters (Flacon) and some other (tools, media centers like LMS, etc.). _Not_ tested with `foobar2000`, a popular Windows application, since I have no Windows machine.

So I propose to use `REM DATE` instead of `DATE`.

We don’t get all possible metadata with all inputs, but probably should think of feasible data to identify tracks with _some_ formats.

- Shoutcast/Icecast streams usually have only one "Title" field, most ofen separated as "Artist - Title".
- Exceptions are OGG streams, as far as I know, these can contain tag data.
- HLS can contain tag data like Artist, Title, Album, Date, ISRC, as far as I know.
- Recording from (internal) sources that were generated from annotated or tagged files/playlists could in theory have lots of tags.

The main goal of using a cue sheet together with a recording MP3/FLAC/WAV will normally be
- legal reasons
- air checks
- keeping a show for later retransmission
- record and list what a DJ transmitted, for licensing reasons
- auto-generating DJ set lists—after the fact—to be published, for instance in a forum.

To _identify_ a track, it might be nice to include these metadata _if_ available:
- "CD" basis (file, album)
  - TITLE
  - PERFORMER
  - REM DATE (YYYY, YYYY-MM, or YYYY-MM-DD)
  - REM COMMENT
- on a TRACK basis (shows can have a multitude of albums, unlike the original CD)
  - TITLE
  - PERFORMER
  - REM ALBUM
  - REM DATE (formats as above)
  - ISRC (this becoming ever more important, and often tagged)

A cue sheet _can_ start with a track number higher than 1 (originally for multi-disc sets), and must be seuqntially counting up, but I guess we can ignore that for our purposes (except maybe the mysterious "max_tracks" setting).

Guess we can also ignore most other options like UPC/EAN etc., since we’re not really talking about CD/DVD/BluRay media here. It’s mainly to create usable recordings and being able to see the titles broadcast, and jump around in a recording.

Moonbase59:
Did several 2-hour testing blocks over night, and the code (cue-more branch) works robust, except for my additional notes above of course:

- Shoutcast 1.9.8 MP3 stream, fed by sc_trans
- Icecast MP3 stream, fed by AzuraCast*
- HLS stream, fed by AzuraCast (4s segments, overlap 2)*
- Local playlist

* = On these, the cue points are often a few seconds off. I _think_ this might be an AzuraCast issue, since I see metadata changing a few seconds off in their web player, too. Maybe it’s just due to the crossfading…

I will now do a 2-hour test from a local playlist, to verify this. Using a playlist and no crossfade, the cue points should be exact.

Moonbase59:
Local playlist to MP3+CUE roughly okay, but some non-exact cues. This seems to depend heavily on the player, for instance Audacious and VLC gave very different results for the same cue point.

I _believe_ the CUE file to be correct, but it seems exact seeking in a MP3 file is not easy for every MP3 audio player…

Retrying the playlist 2-hour test using FLAC as target now.

Btw, `file_type` isn’t taken from the file extension in this version; it’s always `MP3`, even when outputting to a `.flac` file. Fortunately, we can set it separately and thus make parsers happy that expect, for instance `WAVE` with FLAC files.

src/libs/extra/source.liq
```diff
@@ -426,5 +430,6 @@ def source.cue(
     end
   end
 
-  source.on_track(s, handle_metadata)
```
smimram, line 429: Not sure to understand here: why don't we use tracks and not metadata?

smimram (commented):
It would be nice to use the file extension to get the default file type as suggested.

src/libs/extra/source.liq
toots: Metadata can happen without a track mark. In most situations, e.g. when working with files, track marks and metadata happen at the same time but it's not guaranteed.

smimram: I agree with that but I thought it would be more logical to generate cue entries for tracks rather than arbitrary metadata

src/libs/extra/source.liq
```diff
@@ -284,8 +287,9 @@ def source.cue(
   ~comment=null(),
```
smimram, line 287: It would be nice to have `file_type` default to file extension when this is defined instead of hardcoding mp3

toots: Done

src/libs/extra/source.liq
toots: I'm trying to capture the most general use-case. I think metadata+deduplicate gives us the largest use-case cross-section..

toots:
I've pushed the following:
* Renamed `max_tracks` to `last_tracks` to clarify
* Changed `date` to `year`, made it an `int` again. Since you reported that `YYYY` is the most common format I think we should capture that and be explicit about it. Advanced user can re-use our code typically.
* Added `"isrc"` and `"cue_year"` to map to `ISRC` and `REM DATE` per-track. Since date metadata formats is most likely a mess, I thought an explicit `"cue_year"` that you have to specify manually would be the most appropriate.
* Changed `DATE` for `REM DATE`
* Added doc about recognized meta in the operator's description.

Seems to me that this is plenty for now!

toots:
Sorry I forgot:
* Pick `file type` from file extension when it is not provided.

smimram (approved):

Moonbase59:
Yep, plenty. Unfortunately, some things now impossible (long date).

Since no one will have `cue_year` in their file’s tags, I’d need to create that from the many other possible metadata in a file.

**But the `metadata_map` function seems _not_ to receive a file’s complete metadata:**

If I do a

```
def mapmeta(m)
  log(level=4, label="cue", "META: #{m}")
end

...

radio =
  source.cue(
  ...
  map_metadata = mapmeta,
  ...
  )
```

I only see:

```
2024/01/15 21:08:25 [cue:4] META: [("album", "10 000 Hz Legend"), ("artist", "Air"), ("title", "Caramel Prisoner")]
```
but the file has _much_ more metadata, as can be seen when using the same logging in a `radio.on_metadata(…)` display.

toots:
@Moonbase59 Make sure to disable `taglib`. The set of metadata it can read is limited. I'm about to deprecate it I believe.

Moonbase59:
@toots Ah, ok, thanks for the pointer!

Do you by chance know if taglib still/again being developed? (I use it in loudgain, and at that time it was the best/most flexible lib to handle tags uniformly in many file formats.)

toots:
I'm about to push the deprecation. It will make sure that taglib is tried last. The settings could work too.

It looks like it's still active. It was cool when there were a gazillion of formats popping out back then but, nowadays, there's a handful of formats that capture most of people's need and we all support them in `ocaml-metadata` with many more available fields.

Moonbase59:
Yeah, it was really cool back in the days, especially when working with C… I hope they continue it, if ever I come round to updating loudgain… sigh.

Since "fields" are so different in different file types, do you internally use some mapping like MusicBrainz Picard does? So we’d easily find a tag by a "common name" in various file types like M4A, MP3, ID3v2.3, ID3v2.4, FLAC, etc.? _(Hint: If not, I strongly suggest to use [Picard’s Tag Mapping](https://picard-docs.musicbrainz.org/en/appendices/tag_mapping.html)—it has proved worthy for more than a decade.)_

Issue #3630, opened 2024-01-12
HLS Output: Fragments length does not adhere to EXT-X-TARGETDURATION


RobinvdGriend:
**Describe the bug**
I've been trying to get HLS working, but I've hit upon very poor performance: HLS players constantly stall when they try to access the stream generated by liquidsoap. I think the problem might be related to the following.

The `#EXT-X-TARGETDURATION` can be set by the `segment_duration` argument of `output.file.hls`. The problem is that the generated segments themselves are not created with the correct length. Indeed, as you can see in the example below the `#EXTINF` of each fragment is 20.00 and the generate `.ts` fragments are also 20 seconds when openend in VLC. The HLS Spec says that fragment length should never exceed the `TARGETDURATION` for this can have performance problems.

Generated .m3u8:
```
#EXTM3U
#EXT-X-TARGETDURATION:10
#EXT-X-VERSION:3
#EXT-X-MEDIA-SEQUENCE:0
#EXT-X-DISCONTINUITY-SEQUENCE:0
#EXTINF:1.400,
video_high_1.ts
#EXTINF:20.000,
video_high_2.ts
#EXTINF:20.000,
video_high_3.ts
```

**To Reproduce**
Here is the full script:
```
#!/bin/env liquidsoap
settings.init.allow_root := true

# Studio input
studio = input.rtmp("rtmp://0.0.0.0:1935/studio") 
studio = buffer(fallible=true, buffer=5., studio)

# Fallback
pause_animation = single("/assets/pause_animation.mp4")
pause_sound = single("/assets/pause_sound.wav")
pause = source(source.tracks(pause_animation).{audio=source.tracks(pause_sound).audio})

s = fallback(track_sensitive=false,[studio, pause])

# Output formats
video_high = %ffmpeg(format="mpegts",
  %audio.(codec="aac", b="128k"),
  %video.(codec="libx264", b="2000k", preset="faster")
)

video_low = %ffmpeg(format="mpegts",
  %audio.(codec="aac", b="128k"),
  %video(codec="libx264", b="400k", preset="faster")
)

video_formats = [
  ("video_high", video_high),
  ("video_low", video_low),
]

# Video output
output.file.hls(playlist="video.m3u8",
  temp_dir="/output/tmp",
  "/output/hls",
  video_formats,
  s
)
```
The fallback is generated by muxing an animation loop with a sound loop. The main input is gotten from RTMP.

**Expected behavior**
[...]

PR #3636, opened 2024-01-14
Split HLS segment on the latest keyframe when available.

toots:
This PR changes the HLS segment splitting logic to be compliant with the HLS spec[^1]. When reaching the value specified for `EXT-X-TARGETDURATION`, the segment will be forced to be split.

In this case, the split will occur at the latest keyframe in the current segment. If no such keyframe is available, the split occurs right away and a warning is emitted.

It is, thus, recommended to fix the keyframe frequency to a value that will ensure at least one keyframe per segment.

Metadata and new tag splits do not force a segment split. It is expected that setting a correct keyframe frequency w.r.t. segment length will also produce an appropriate keyframe interval for those splits.

Fixes: #3630

[^1]: https://datatracker.ietf.org/doc/html/rfc8216#section-4.3.3.1

### DIFF
--- a/doc/content/hls_output.md
+++ b/doc/content/hls_output.md
@@ -43,14 +43,13 @@ To make sure that all these requirements operate correctly, you should make sure
 keyframe per segment.
 
 For instance, if your segments are at most `2s` long and encoded using `libx264`, you can use the `keyint` and `keyint-min` parameters, which are expressed
-in number of frames. If your video frame rate is `25fps` (liquidsoap's default), you should have at least one keyframe every `12` frames, which is `25/2` rounded
-to the lowest integer:
+in number of frames. If your video frame rate is `25fps` (liquidsoap's default), you should have at least one keyframe every `2 * 25 = 50` frames.
 
 ```liquidsoap
 %ffmpeg(
   ...,
   %video(codec="libx264",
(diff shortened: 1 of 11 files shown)


doc/content/hls_output.md
```diff
@@ -28,6 +28,32 @@ Liquidsoap also provides `output.harbor.hls` which allows to serve HLS streams d
...
+you will see a warning in the logs.
+
+Segment split is forced when reaching the value specified by `EXT-X-TARGETDURATION` to follow the HLS specifications. For metadata and extra tags, segment split will
+occur at the next keyframe.
+
+To make sure that all these requirements operate correctly, you should make sure to set a keyframe frequency in your encoder's settings that generates at lease one
+keyframe per segment.
+
+For instance, if your segments are at most `2s` long and encoded using `libx264`, you can use the `keyint` and `keyint-min` parameters, which are expressed
+in number of frames. If your video frame rate is `25fps` (liquidsoap's default), you should have at least one keyframe every `12` frames, which is `25/2` rounded
```
RobinvdGriend, line 46: It could be that I'm misunderstanding, but if the segment length is 2 seconds, this would be 1 keyframe per 50 frames right?

toots: Whoops correct thanks!

PR #3660, opened 2024-01-24
Cache some video effect.

toots:


smimram:
Your improvement is cute, but I would rather do this on the Liquidsoap side: morally, we should define
```
add_rectangle(s) = add(s, video.rectangle())
```
and similarly for `add_line`...

PR #3664, opened 2024-01-25
Fix n+1 issue when checking file mime and extension.

toots:
This PR fixes a `n+1` issue when checking files for decoding: we were re-computing mime type and file extension on each call.

Besides, there seems to be some file descriptor issues with the magic library. This PR also switches to a more recent and currently maintained library. Since the library is pure OCaml with no external dependency, it also added to the set of required dependencies, making file mime type always available.


toots:
> Looks nice thanks! One small question: it seems that [ocaml-magic-mime](https://github.com/mirage/ocaml-magic-mime) only does detection of the mime based on the file extension whereas [ocaml-magic](https://github.com/Chris00/ocaml-magic) was using libmagic which can also detect file based on their content. Don't we loose something here?

From my recollection, we were very excited initiall about `libmagic` and then realized that its heuristics were pretty approximate when detecting based on content. For more of our file-based operations, we should assume that file extension is much more reliable. Even since we switched to keeping the original extension in the decoding process, our content detection has been much more robust.

I don't think it hurts to consider mime types for files but, in-fine, the file extension check will always be there and more accurate.

Moonbase59:
@toots @smimram On Linux, file extensions are rather irrelevant, they’re mostly used, but more as a friendly nod to Windows users. Please be sure to consider case-sensitive filesystems, where, for instance, `.jpg` and `.JPG` _are_ a difference. (And we also have `.jpeg` and `.JPEG` in this case.)

Just my 2¢.

toots:
> @toots @smimram On Linux, file extensions are rather irrelevant, they’re mostly used, but more as a friendly nod to Windows users. Please be sure to consider case-sensitive filesystems, where, for instance, `.jpg` and `.JPG` _are_ a difference. (And we also have `.jpeg` and `.JPEG` in this case.)
> 
> Just my 2¢.

File extensions are converted to lower-case before comparison. You can see the full list [here](https://www.liquidsoap.info/doc-dev/settings.html#file-extensions-used-for-guessing-audio-formats), feel free to let us know if you see anything missing.

I don't know where you get that file extensions are irrelevant on linux. They are still very much used in my experience. Some file formats, most importantly mp3, are _very_ hard to guess by looking at a file's content so the file extensions is the best indicator we can have for that kind of content. In general, we want to promote a certain level of predictability in the media that is used by our users.

Moonbase59:
> Some file formats, most importantly mp3, are _very_ hard to guess by looking at a file's content so the file extensions is the best indicator we can have for that kind of content. In general, we want to promote a certain level of predictability in the media that is used by our users.

Yeah, sure, most of us moved to that, for exactly reasons like that. And of course it’s easier to read in a file listing, if you’re a human. That’s why I think it won’t be much of a problem. And good to know things are lowercase-compared, thanks!

PR #3703, opened 2024-02-03
Add crossfade delay, tests and plot cababilities.

toots:
This PR adds delays to `fade.in` and `fade.out` crossfades. They work in different ways:
* For `fade.in`, delay adds extra blank data before the track starts with its fade in
* For `fade.out`, delay delays the start of the fade out within the buffered data.

A facility to plot and test fades is added. Here's an example:

```liquidsoap
a = sine(duration=20., 440.)
a =
  metadata.map(
    update=false,
    fun (_) ->
      [
        ("liq_cross_duration", "5."),
        ("liq_fade_out_type", "exp"),
        ("liq_fade_out_curve", "10"),
        ("liq_fade_out_delay", "2"),
        ("liq_fade_out", "3.")
      ],
    a
  )

b = sine(duration=20., 880.)
b =
  metadata.map(
    update=false,
    fun (_) ->
      [
        ("liq_fade_out_type", "lin"),
        ("liq_fade_in_delay", "2"),
        ("liq_fade_in", "2.")
      ],
    b
  )

s = sequence([a, b])
```
This results in a crossfade of the following shape:

plot


src/libs/extra/fades.liq
```diff
@@ -0,0 +1,77 @@
...
+    s =
+      source.on_frame(
+        s,
+        {
+          if
+            null.defined(png) and not plotted()
+          then
+            ignore(
+              process.run(
+                "gnuplot -e #{process.quote(gnuplot_cmd(null.get(png)))}"
```
smimram, line 65: I am being picky here, but this could be done fully in liq I think (see `video.plot` to plot and `video.still_frame` to save it).

src/libs/fades.liq
```diff
@@ -49,38 +49,43 @@ def mkfade(
   end
 
   shape =
-    if type == "sin" then
+    if
```
smimram, line 52: It would be nice to apply auto-indentation on main, so that we can see the reall diff here.

toots: Yeah, this is the `v2.2.x` branch, the auto-indentation wasn't applied yet. You can see a proper diff in the `main` backport: https://github.com/savonet/liquidsoap/pull/3704

src/libs/extra/fades.liq
toots: Definitely down to improve. I'm only using the generated image when updating the test data to make sure that the data is reflective of what I'm expecting the crossfade to be doing.

PR #3728, opened 2024-02-09
HOTFIX: Fix crossfade sequence when having a tail after transition.

toots:
There is a problem in the new mechanism to compute crossfade over the common buffer: this code:
```ocaml
new Sequence.sequence ~merge:true [compound; after_tail]
```
Will skip over to `after_tail` on the first track mark from the coumpound source.

It seems good to try and keep the new crossfade computation over the common buffer so, the safe solution is to add a new, opt-in behavior to `sequence` that allows it to play from the same source while it is available instead of moving to the next source on the first track mark.

This is achieved by the new `single_track` option, sets to the default, current behavior (`true`).

This will require a bump of the release assets.


vitoyucepi:
Hi @toots,
Is it necessary to merge this into rolling-release-v2.2.x?

toots:
I usually re-create the `rolling-release-v2.2.x` branch from the latest stable release once the first bugfix happen.

PR #3753, opened 2024-02-19
Add autocue

toots:
This PR adds a super neat feature implemented by @RM-FM that allows to automatically compute cue-in/cue-out points as well as crossfade parameters!

The feature relies on the FFmpeg [ebur128](https://ffmpeg.org/ffmpeg-filters.html#ebur128-1) filter. First, the file is sliced into frames of `100ms` for which LUFS data is returned. Then @RM-FM's code uses that data to detect the right parameters.

The functionality is implemented the same way that replaygain is: either via a metadata decoder, enabled for all files using `enable_autocue_metadata()` or via a `autocue:` protocol, which can be added on per-uri basis.

If a file or request already has metadata overrides, the autocue functionality will not return anything and let the overrides bubble up.


Warblefly:
This looks neat! I was hoping it could be done, but I don't understand the LS language enough — great that other friends do!

gAlleb:
Great stuff. Thank you, guys! Gonna listen )

PR #3766, opened 2024-02-23
Adapt to new channel layout.

toots:
This PR switches to the new ffmpeg channel API.

It is a breaking change from the build perspective but shouldn't break anything from the user side.

It is required because ffmpeg has picked up the dev pace, which is great, but is also now moving fast with new API. `6.x` has a backward compatible API for channels but `7.0` was just released without it and, soon, most recent platform/OS will only ship `7.0` so we have to move on with it.

On the plus side, our `ffmpeg` binding now compiles without a single warning!

On the down side, this forces the use of backports for OS/platforms that do not yet have at least ffmpeg `6.0`. This is the case for debian `bookworm` which will now build and ship packages relying on [deb-multimedia.org](https://www.deb-multimedia.org/) to get a recent ffmpeg version.

Likewise, the ubuntu platforms are bumped to `noble` (latest LTS) and `oracular` (latest release).

### DIFF
--- a/.github/docker/Dockerfile.production
+++ b/.github/docker/Dockerfile.production
@@ -1,4 +1,4 @@
-FROM debian:trixie-slim
+FROM debian:bookworm-slim
 
 ARG DEB_FILE
 ARG DEB_DEBUG_FILE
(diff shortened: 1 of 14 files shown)


vitoyucepi:
Hi @toots,
I didn't see this PR in time, but I should note that officially the update from *ubuntu:22.04* to *ubuntu:24.04* will not be available until the *24.04.1* release in mid-August.
Is it possible to continue building deb packages for *jammy*, or will there be no *2.3* for it because of *ffmpeg* compatibility?

toots:
At the moment, `rolling-release-v2.3.x` is targeting `noble` and `oracular`. I think that, as a general rule, we want to support the latest stable/LTS and the latest regular release/`testing` (for debian).

toots:
But, also, I don't think that `2.3.0` will be out before at least September. Too many people are out right now to get enough pre-release testing.

PR #3767, opened 2024-02-24
Generic after_run.

toots:
In the process of rewriting the `clock` API, the most challenging concept seems to be when to unify and instantiate clocks?

Initially, sources are created with placeholders for their clock. Once the code that creates them has run, the clocks are:
1. Unified, e.g. in `output.ao(playlist("..."))`, the clock of the output is unified with the clock of the `playlist`
2. Instantiated, i.e. an actual clock is created and its main thread is started.

In order to give the user the ability to declare multiple outputs, for instance, or to manually assign clocks, we need to make sure that the clock instantiation is done after all the relevant code has been executed.

Consider for instance this code:
```liquidsoap
def create_two_output() =
  s = playlist("...")
  clock.assign_new([s])
  output.dummy(s)

  s = playlist("...")
  clock.assign_new([s])
  output.dummy(s)
end
```
This code creates two sources and output, each in their own clock.

Here's another simple example:
```
def create_two_output() =
  s = playlist("...")
  output.dummy(s)
  output.dummy(s)
end
```
Here, we want the clocks for each output to be the same. If each `output` instantiates their own clock too early, the clock unification will fail.

Currently, this is done via a `Clock.collect_after` handler which needs to wrapped around all code allocations. This is cumbersome, confusing and creates an unnecessary dependency from the `core` in the language.

This PR offers a different approach: a `on_after_eval` callback registration. The implementation works as follows:
* When code is executed, an effect handler is installed by the top-level call to collect all `on_after_eval` callbacks
* When the top-level call terminates, the collected callbacks are executed.

This creates a generic local, concurrent and lock-free mechanism to execute specific callbacks after some code has ran.

This is then used in `add_operator` and `add_track_operator` to make sure that the clock collection is ran each time any part of the script creates a clock but only after the whole execution has returned.

The implementation relies on the new effect handlers from OCaml `5.x`. This is challenging because of the following missing/in-progress features:
* `js_of_ocaml`: ~effects are supported but the build is broken with a new bug. Reported [here](https://github.com/ocsigen/js_of_ocaml/issues/1569)~. Fixed by temporarily vendoring the lock-free queue code to prevent a hard dependency on `threads`.
* `memtrace`: there is a [PR](https://github.com/ocaml/ocaml/pull/12923) pending to bring it back!
* windows: the cross-compiler used to provide windows binaries need to be updated.

This is a rather hard switch but we have been considering taking advantage of the multicore for a while. So far, the plan was to use a library with a `4.x` backward compatible API but perhaps it is time to consider a hard switch to OCaml `5.x`?


src/js/dune
```diff
@@ -1,5 +1,8 @@
 (env
  (dev
+  (js_of_ocaml
+   (flags (:standard --enable effects))
+   (build_runtime_flags (:standard --enable effects)))
```
hhugo, line 5: You should also set the flags for non `dev` profiles (e.g. release).

toots: Thanks!

toots:
PR is ready!

* `saturn` dependencies have been temporarily vendored pending a fix removing the hard dep on `threads`
* Windows image with OCaml `5.1.1` is now available!

PR #3783, opened 2024-03-01
feat: Add forced_major_collections back to the GC stat record

vitoyucepi:
## Summary
This commit adds `forced_major_collections` back to the `runtime.gc.stat()` and `runtime.gc.quick_stat()` results because the minimum required ocaml version is 4.14.
It was removed in 889cc4f8 to be compatible with ocaml 4.08.


vitoyucepi:
Maybe I should test the presence of the keys in the result of `runtime.gc.stat()` and `runtime.gc.quick_stat()`.

toots (approved):
Thanks!

Issue #3789, opened 2024-03-06
Percent encoding in URLs with http functions


vitoyucepi:
**Describe the bug**
When I try to use percent-encoded entities in the url of the `http.get()` it decodes them, thus breaking the url.

**To Reproduce**
1. Start the http server.
1. Start Wireshark.
1. `http.get("http://192.168.0.1/%2f")` -> `GET //`.
   `http.get("http://192.168.0.1/\%2f")` -> `GET /%5C%2F`.
   `http.get("http://192.168.0.1/%2e")` -> `GET /`.
1. `curl http://192.168.0.1/%2f` -> `GET /%2F`.
   `curl http://192.168.0.1/\%2f` -> `GET /\%2F`.
   `curl http://localhost:8000/%2e` -> `GET /%2e`.

**Expected behavior**
Percent-encoded entities should be left as they are.

**Version details**

- OS: *debian:12* in docker
- Version: *2.2.4-1*

**Install method**
*savonet/liquidsoap:v2.2.4*

**Common issues**
https://github.com/savonet/liquidsoap/discussions/3782#discussioncomment-8693869

toots:
The issue was introduced while trying to fix https://github.com/savonet/liquidsoap/issues/2551

PR #3793, opened 2024-03-09
Make url normalization optional, add warning when normalization changes the url.

toots:
Fixes: #3789

### DIFF
--- a/src/core/builtins/builtins_http.ml
+++ b/src/core/builtins/builtins_http.ml
@@ -22,6 +22,13 @@
 
 type request = Get | Post | Put | Head | Delete
 
+let string_of_request = function
+  | Get -> "get"
+  | Post -> "post"
+  | Put -> "put"
+  | Head -> "head"
+  | Delete -> "delete"
+
 let request_with_body = [Get; Post; Put]
 let http = Modules.http
 let http_transport = Modules.http_transport
@@ -36,6 +43,7 @@ let add_http_request ~base ~stream_body ~descr ~request name =
   let header_t = Lang.product_t Lang.string_t Lang.string_t in
   let headers_t = Lang.list_t header_t in
   let has_body = List.mem request request_with_body in
+  let log = Log.make ["http"; string_of_request request] in
   let request_return_t =
     Lang.method_t
       (if (not has_body) || stream_body then Lang.unit_t else Lang.string_t)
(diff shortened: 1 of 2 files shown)


elizaveta-markova:
I get "... or use 'normalize_url=false.'" But use it how? The only relevant operator I call is playlist, and playlist doesn't accept that argument.

toots:
Hi @elizaveta-markova ! I think that this for the upcoming `2.3.0` version only!

toots:
Ha sorry my bad it should also be in https://github.com/savonet/liquidsoap/releases/tag/rolling-release-v2.2.x

The changes are indeed only applied to the `http.get` and similar functions. Let me double check that we do need them for `playlist`.

elizaveta-markova:
Thank you, Romain. Incidentally, the messages seem to be false positives,
anyway. I don't see any difference between the two names that are said to
be different, and the files play.

BTW I did not file anything here on the other issue I mentioned on Discord.
It's not a bug, just me trying to explain what happened. But I don't think
what I was trying to do is supposed to work, anyway. A stream of a still
image followed by a video doesn't really make sense: it should just be a
short looping video followed by the main video.

Thanks for the software and the charming book. It's much more than I have
any need for, but I've had fun and learned a lot.

Liz

toots:
Thanks!

This message is confusing and also the parameter might not be available for instance when a URL is resolved through the file resolution protocol. 

I added a global setting and improved the message:

```
          "Requested url %s is different from normalized url: %s. URL are \
           normalized by default to ensure maximum compatibility with e.g. \
           URLs with spaces in them. However, this can also cause issues so we \
           recommend passing normalized URLs. Url normalization can be \
           disabled on a case-by-case basis using the  `normalize_url=false` \
           parameter or globally using the `settings.http.normalize_url` \
           setting."
```

Hope this will make more sense!

elizaveta-markova:
Setting is good. But you've got %s and %s in the message. Is that why I
can't see a difference?

toots:
Yes, sorry it is populated with the actual URLs.

elizaveta-markova:
Populated, yes, but with the same thing twice, rather than with the two putatively different names? Or can %s and %s refer to two different strings?

toots:
One will be the original stream passed to the function and the other the normalized URL. If they differ, this can create issues so the log is here to surface and explain what is going on.

PR #3865, opened 2024-04-08
Autocue - Introducing sustained ending support

RM-FM:
This is an improvement of autocue to optimize crossfade and cue out points in case of sustained endings. This new algorithm is analysing the loudness behavior after the initially calculated crossfade and cue out point. If it's still a part of the song that is likely to be of relevance, the crossfade is recalculated based on that data.

The relevance is determined on the basis of the loudness reduction immediately after the original crossfade point and a possible "plateau" (i.e. no continuous fade-out).

### DIFF
--- a/src/libs/autocue.liq
+++ b/src/libs/autocue.liq
@@ -258,11 +258,34 @@ def autocue.internal.implementation(uri) =
 
     # Extract timestamps for cue points
     # Iterate over loudness data frames and set cue points based on db thresholds
-    cue_in_found = ref(false)
     last_ts = ref(0.)
     current_ts = ref(0.)
+    cue_found = ref(false)
+    cross_start_idx = ref(0.)
+    cross_stop_idx = ref(0.)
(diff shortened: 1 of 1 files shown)


RM-FM:
@toots After some tweaking, I'm very happy with the result.

src/libs/autocue.liq
```diff
@@ -468,9 +753,6 @@ def enable_autocue_metadata() =
   settings.crossfade.assume_autocue := true
 
   def autocue_metadata(~metadata, fname) =
-    print(
-      "fname: #{fname}, metadata: #{metadata}"
-    )
```
toots, line 473: Whoops I think that was me.

RM-FM:
@toots I've completed my tests. Please feel free to test as well...
Default is `settings.autocue.internal.sustained_endings_enabled := true`.

toots:
This has been integrated with #3870 and will be merged via that PR.

PR #3876, opened 2024-04-17
Add private/swap memory and telnet command.

toots:


tests/language/mem_usage.liq
```diff
@@ -5,6 +5,24 @@ def f() =
...
+
+  test.metric(
+    category="memory",
+    name=
+      "Memory usage: process swapped memory",
+    value=float(mu.process_swapped_memory) / (1024. * 1024.),
+    unit="Mio",
+    min=0.
+  )
+
```
vitoyucepi, line 25:
```suggestion
  test.metric(
    category="memory",
    name=
      "Memory usage: process private memory",
    value=float(mu.process_private_memory) / (1024. * 1024.),
    unit="MiB",
    min=0.
  )

  test.metric(
    category="memory",
    name=
      "Memory usage: process swapped memory",
    value=float(mu.process_swapped_memory) / (1024. * 1024.),
    unit="MiB",
    min=0.
  )

```

I know what's *Mio*, but maybe *MiB* is better?

vitoyucepi (changes requested):

tests/language/mem_usage.liq
toots: Good catch, thanks!

PR #3931, opened 2024-05-21
Autocue for 2.3.x

toots:
This is a backport of the `v2.2.5` autocue library with breaking changes:
* Full remove of `smart` mode. We should prioritize `autocue`. This also makes it possible to have breaking changes
* Split crossfade duration into `start` and `end` duration. This makes the whole logic _much_ more clear.


Moonbase59:
Renaming `liq_fade_in` & `liq_fade_out` in `fades.liq` to `liq_fade_in_duration` & `liq_fade_out_duration` will break user’s pre-tagged files if they use `autocue.cue_file`.

Splitting the `liq_cross_duration` into `liq_cross_start_duration` and `liq_cross_end_duration` seems a good idea, although we would have to have different `autocue.cue_file.liq` and `cue_file` versions for 2.3.0 then.

Is there a setting I could query to see if the current autocue is compatible, like a Liquidsoap version number (should be SemVer-compatible, i.e. x.y.z)?

toots:
> Renaming `liq_fade_in` & `liq_fade_out` in `fades.liq` to `liq_fade_in_duration` & `liq_fade_out_duration` will break user’s pre-tagged files if they use `autocue.cue_file`.

I don't think that we are changing this.

> Splitting the `liq_cross_duration` into `liq_cross_start_duration` and `liq_cross_end_duration` seems a good idea, although we would have to have different `autocue.cue_file.liq` and `cue_file` versions for 2.3.0 then.

`liq_cross_duration` was determined to be an _internal_ parameter during the autocue process. As far as I can see, it is not returned by https://github.com/Moonbase59/autocue

> Is there a setting I could query to see if the current autocue is compatible, like a Liquidsoap version number (should be SemVer-compatible, i.e. x.y.z)?

Yes, you have `liquidsoap.version` and [liquidsoap.version.at_least](https://www.liquidsoap.info/doc-2.2.5/reference.html#liquidsoap.version.at_least)

toots:
After some positive early testing by @gAlleb and feedback that this'll need to be shipped in some sort of AzuraCast image to be able to be tested there, I'm merging this to move this code in front of more people.

PR #3959, opened 2024-06-10
Cache plugins

smimram:
Plugins such as LV2 take quite some time to load. We try to cache those here.


src/lang/term/term_cache.ml
```diff
@@ -80,58 +26,9 @@ let retrieve ?name ~trim parsed_term : Term.t option =
...
-        | exn ->
-            let bt = Printexc.get_backtrace () in
-            let exn = Printexc.to_string exn in
-            if Sys.getenv_opt "LIQ_DEBUG_CACHE" <> None then
-              Startup.message "Error while loading cache: %s\n%s" exn bt
-            else Startup.message "Error while loading cache: %s" exn;
-            None)
+      if Cache.enabled () then
+        Cache.retrieve ?name (cache_filename ?name ~trim parsed_term)
+      else None)
```
toots, line 31: Should move `report` inside the enabled branch so we don't report when cache is disabled.

toots (approved):
Looking good!

PR #3979, opened 2024-06-20
Propagate constants

toots:


toots:
This has not proved to same much memory consumption and the associated complexity is pretty daunting.

PR #3982, opened 2024-06-21
Update `queue` in `request.queue`

austin0209:
Whenever we use `set_queue` in `requests.queue` it updates the backing `queue` list, however when we try to get the queue using `requests.queue.queue()`, we only get the requests within `s.queue`, which is set to empty after calling `set_queue`. 

Therefore, even though the queue should have the items specified in `set_queue`, we get an empty list when calling `queue`. See the `show_queue` server function for reference.

### DIFF
--- a/src/libs/request.liq
+++ b/src/libs/request.liq
@@ -186,7 +186,7 @@ def request.queue(
     s.set_queue([])
   end
 
-  def queue() =
+  def get_queue() =
     [...s.queue(), ...(queue())]
   end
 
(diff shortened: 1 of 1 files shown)


austin0209:
For additional context, this happens when you push the requests into the queue using the `push` telnet function, then call `set_queue`. `queue()` works prior to calling `set_queue`, however once `set_queue` is called once, `queue()` is broken.

My guess is the telnet function pushes the requests into `s.queue` which is why `queue()` works in the first place.

toots:
Thanks for this, this looks great. I'm gonna push a backport to `rolling-release-v2.2.x` as well.

PR #3988, opened 2024-06-23
Convert types.

toots:


toots:
I'm gonna let this one go. I see a meager `~4%` improvement on memory footprint with an increase in typechecking time, most likely due to the data copies induced by the fact that inline record cannot be processed outside of the scope of their variant. I was a nice exploration, though.

PR #3992, opened 2024-06-26
Fix telnet functions for request.queue

austin0209:
This is a continuation of #3982 

There are issues with the telnet functions because they do not use the overridden methods of the request.queue implementation. 

For example, `flush_and_skip` does not actually clear the entire queue since it does not use the overwritten `set_queue` which clears the internal `s.queue()` and modifies the backing `queue`. Since it uses the default implementation of `set_queue` (and not the overwritten one), it does not set the backing `queue` to `[]`, so any requests that were in the backing `queue` are left over.

Fixed this by overwriting the methods beforehand, and using the result of that when specifying the telnet functions.

### DIFF
--- a/src/libs/request.liq
+++ b/src/libs/request.liq
@@ -95,6 +95,32 @@ def request.queue(
     default()
 %endif
 
+  def add(r) =
+    log.severe(
+      label=s.id(),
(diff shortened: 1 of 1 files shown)


toots:
This looks great again. I'm not sure why the telnet function was returning the full queue but not the corresponding method. This seems like an oversight.

PR #4001, opened 2024-07-01
Calculate resources consumption during CI

toots:


toots:
This was very useful data but I think it is slightly too cumbersome to keep permanently in the CI.

PR #4006, opened 2024-07-05
asan build.

toots:


toots:
This was a PR to trigger a build with ASAN. Closing now but cool to know this is possible and could be re-used.

Issue #4005, opened 2024-07-05
Cover Manager: current_cover_file could be removed


nightridefm:
https://github.com/savonet/liquidsoap/blob/33a62b4858c0d1a9b442fe7aae13f7a0b43c89d0/src/libs/extra/metadata.liq#L28-L42

If there is a name collision with the incoming `filename` vs `last_cover_file()`, the current cover file will be removed and cause source errors

vitoyucepi:
Hi @nightridefm,
Could you share the logs for this case?
As far as I remember the `filename` can be either `default` or the result `file.temp`.

nightridefm:
@vitoyucepi it can, but if for whatever reason the previous file.temp name is the same as the next filename, it will be removed
at least `last_cover != default` should become `last_cover != default and last_cover != filename`

I am actually writing my own version that uses `file.exists()` as well, just in case.

Sample of logs spamming my server, burning CPU and causing catchup errors
```
2024/07/05 04:23:52 [request:3] Nonexistent file or ill-formed URI "/tmp/cover-manager85f0fe/cover_ada73c.jpeg"!
2024/07/05 04:23:52 [request:3] Nonexistent file or ill-formed URI "/tmp/cover-manager85f0fe/cover_ada73c.jpeg"!
2024/07/05 04:23:52 [request:3] Nonexistent file or ill-formed URI "/tmp/cover-manager85f0fe/cover_ada73c.jpeg"!
2024/07/05 04:23:52 [request:3] Nonexistent file or ill-formed URI "/tmp/cover-manager85f0fe/cover_ada73c.jpeg"!
2024/07/05 04:23:52 [request:3] Nonexistent file or ill-formed URI "/tmp/cover-manager85f0fe/cover_ada73c.jpeg"!
2024/07/05 04:23:52 [request:3] Nonexistent file or ill-formed URI "/tmp/cover-manager85f0fe/cover_ada73c.jpeg"!
2024/07/05 04:23:52 [request:3] Nonexistent file or ill-formed URI "/tmp/cover-manager85f0fe/cover_ada73c.jpeg"!
2024/07/05 04:23:52 [request:3] Nonexistent file or ill-formed URI "/tmp/cover-manager85f0fe/cover_ada73c.jpeg"!
2024/07/05 04:23:52 [request:3] Nonexistent file or ill-formed URI "/tmp/cover-manager85f0fe/cover_ada73c.jpeg"!
[...]

PR #3651, opened 2024-01-22
Add cover manager

toots:
This PR adds an extra API that facilitates the handling of covers to be used in video streams.

PR #4024, opened 2024-07-13
Remove intermediary last cover.

toots:
@vitoyucepi is there any reason we're keeping the last two covers in this function?


vitoyucepi:
I don't know the original PR has only one cover.
Ref: ef4d2bac3adad72ed563b65362536f90e53bf71c, #3651.

toots:
> I don't know the original PR has only one cover. Ref: [ef4d2ba](https://github.com/savonet/liquidsoap/commit/ef4d2bac3adad72ed563b65362536f90e53bf71c), #3651.

That might have been me working against this bug. Duh of course I didn't add a commit comment.

vitoyucepi (approved):

src/lang/builtins_string.ml
```diff
@@ -29,6 +29,15 @@ let _ =
...
 
+let _ =
+  Lang.add_builtin ~base:string "digest" ~category:`String
+    ~descr:"Return an MD5 digest for the given string."
+    [("", Lang.string_t, None, None)]
+    Lang.string_t
+    (fun p ->
+      let data = Lang.to_string (List.assoc "" p) in
+      Lang.string Digest.(to_hex (string data)))
+
```
vitoyucepi, line 40: Why is this change related to the cover?

toots: Not at all. I tried something, realized we didn't have `string.digest` and the reverted the changes but kept the new function.

I was trying to store the cover with a digest of its content but then realized we also need to remove them so I reverted..

toots: I've been testing it and reviewed your remark about the temp file, you are correct, the file should always be either the default cover or a unique temp file.

toots: I'll add this directly to main and will rebase here.

vitoyucepi:
After merging this PR, I'd like to reopen #4005 to continue the investigation since you're rolling back fixes from a4cb8e2bc8f05b4b56d95e234ef7e607db77524c.

Issue #3752, opened 2024-02-19
Unable to build liquidsoap with gstreamer


vitoyucepi:
**Describe the bug**
Liquidsoap cannot be built with *gstreamer* enabled since #3665.

<details>
<summary>Build error</summary>

```
#=== ERROR while compiling liquidsoap-core.2.2.4-1 ============================#
# context     2.1.5 | linux/x86_64 | ocaml-base-compiler.4.14.1 | https://opam.ocaml.org#8b5a0a33
# path        ~/.opam/4.14.1/.opam-switch/build/liquidsoap-core.2.2.4-1
# command     ~/.opam/4.14.1/bin/dune build -p liquidsoap-core -j 5 --promote-install-files=false @install
# exit-code   1
# env-file    ~/.opam/log/liquidsoap-core-13391-20ab37.env
# output-file ~/.opam/log/liquidsoap-core-13391-20ab37.out
### output ###
# (cd _build/default && /root/.opam/4.14.1/bin/ocamlc.opt -w -40 -g -bin-annot -I src/core/.liquidsoap_gstreamer.objs/byte -I /root/.opam/4.14.1/lib/angstrom -I /root/.opam/4.14.1/lib/bigstringaf -I /root/.opam/4.14.1/lib/bytes -I /root/.opam/4.14.1/lib/camlp-streams -I /root/.opam/4.14.1/lib/camomile/lib -I /root/.opam/4.14.1/lib/cry -I /root/.opam/4.14.1/lib/curl -I /root/.opam/4.14.1/lib/dto[...]
# File "src/core/decoder/gstreamer_decoder.ml", line 314, characters 37-51:
# 314 |       (Decoder.test_file ~log ~mimes:mime_types#get
#                                            ^^^^^^^^^^^^^^
# Error: This expression has type string list
#        but an expression was expected of type Decoder.stream list option
```

</details>

**To Reproduce**
1. `docker run -it --rm debian:bookworm-slim bash`
1. ```sh
   OPAM_VERSION=2.1.5;
   OCAML_VERSION=4.14.1;
   ```
   ```sh
   apt-get update;
   apt-get dist-upgrade;
   apt-get install --no-install-recommends -y curl ca-certificates patch unzip bubblewrap git bzip2 gcc make libc6-dev g++;
   curl -sS -L -o /usr/local/bin/opam "https://github.com/ocaml/opam/releases/download/$OPAM_VERSION/opam-$OPAM_VERSION-x86_64-linux";
[...]

PR #3769, opened 2024-02-24
Fix gstreamer decoder after #3664

vbgl:
Fixes #3752

Issue #1020, opened 2019-11-03
Gstreamer RTMp freezes


Infernion:
**Describe the bug**
I'm trying to build the script that combine rtmp VA with http audio based on some rules. I've made it work, but the actual script doesn't work stable. I use gstreamer to receive AV and audio. After several minutes of work the AV un-sync become visible. The image freezes from time to time. And in the end the RTMP stream  stops. The log says `Restarting pipeline` but no restart happening actually.

**To Reproduce**
```set("frame.video.width", 1920)
set("frame.video.height", 1080)
set("frame.video.samplerate", 30)
set("frame.audio.samplerate", 48000)
set("gstreamer.add_borders", false)
set("clock.allow_streaming_errors", false)


origin_video_stream_address = "rtmp://0.0.0.0:1935/live_output/stream"
stream_to_server_address = "rtmp://0.0.0.0:1935/live_en/stream"
icecast_audio_address= "http://0.0.0.0:7800/en"

audio_trans = mksafe(
        input.gstreamer.audio(
          pipeline="uridecodebin uri=#{icecast_audio_address}"
    )
)

origin_video_stream = mksafe(input.gstreamer.audio_video(
    pipeline="rtmpsrc location=#{origin_video_stream_address} ! tee name=t",
    audio_pipeline="t. ! queue",
    video_pipeline="t. ! queue"
    )
)

stream = mux_video(
        video=drop_audio(origin_video_stream),
        add([drop_video(origin_video_stream), audio_trans])
)

output = output.gstreamer.audio_video(
        video_pipeline="videoconvert ! x264enc bitrate=5000 tune=zerolatency pass=17 ! video/x-h264,profile=main ! queue ! mux.",
        audio_pipeline="audioconvert ! voaacenc bitrate=128 ! queue ! mux.",
        pipeline="flvmux streamable=true name=mux ! rtmpsink location=\"#{stream_to_server_address} live=1\"",
        stream
)
```

**Expected behavior**
The script should work without AV sync diff and restart output if some error happens.

**Version details**
 - OS: Ubuntu 18.04
 - Version 1.4.0

**Install method**
OPAM installation

The actual log:
```
sudo su -l liquidsoap -s /bin/bash -c "TRANS_TO_LANG_CODE=en GST_DEBUG="*:2" liquidsoap /etc/nginx/scripts/stream_lang.liq"
2019/11/03 02:21:51 >>> LOG START
[...]

PR #4036, opened 2024-07-18
Remove gstreamer.

toots:
Gstreamer has been deprecated for years.

It has been historically very hard to integrate because its API assumes that the application using it functions like a GUI application with a main event loop directing the application.

As a result of the difficulty to maintain a working gstreamer, it has been disabled in our internal tests, resulting in long stretch of broken operators if not broken compilation altogether.

We have offered `ffmpeg` as an alternative for years now and this seems to be working well. We are constantly working on enhancing it and we believe that it offers pretty much the same functionalities if not more.

This fixes: #3752, #3769, #3752, #1020


toots:
Gonna merge this too. `gstreamer` had been marked as deprecated for a while and I would like to get feedback if people actually need it when the pre-release for `2.3.0` comes out.

PR #4037, opened 2024-07-18
Add `typeof`

toots:
This PR adds a `typeof <value>` type annotation.

It has been proposed and scrapped several times but, recently, a pretty valid use-case popped up: https://github.com/savonet/liquidsoap/discussions/3760#discussioncomment-10084540

The idea being that `json` parsing is driven by the type but sometimes, describing a complex type is complicated and also hard to mainain.

Instead, with this PR, it is possible to simply use your value and then ask to parse according to its type. 

For instance, if you are storing data in a ref and wants to fetch the latest saved value, you can do:

```liquidsoap
data = ref([])

try
  let json.parse (savedValue: typeof data()) = file.contents("/path/to/saved/data.json")
  data := savedValue
catch _ do
  ()
end

# Now use `data` and add value to it etc..
```

### Technical notes

The `typeof v` type annotation has to be converted to a type. It is done via a `lazy` value. The term referred to is kept in the closure of the lazy value and typechecked when the type is forced.

`typeof` type values are stored in the same facility as type variables. This is a bit hacky but it makes it safer as they can be dereferenced just like regular type variables, thus being part of the existing type API flow without much required change.


toots:
I'm gonna wait for @smimram's opinion on this. There's to rush to merge but I think that it would be a great user-friendly feature.

gAlleb:
It would be a great user-friendly feature indeed) Thanks!

PR #4039, opened 2024-07-19
Fix metadata cover cleanup.

toots:
I finally figured out the issue with #4005 and the double cover save hack: once cover file are pushed down to the request pipeline, the underlying file should not be removed!

Instead, we have to lift the various APIs to directly consume requests so we can pass cover as plain requests and let the normal request cleanup mechanism take care of removing the temporary files.

Fixes: #4005

### DIFF
--- a/src/libs/extra/metadata.liq
+++ b/src/libs/extra/metadata.liq
@@ -23,23 +23,6 @@ def file.cover.manager(
 
   current_cover_request = ref(default)
 
-  def save_cover_to_file(cover) =
-    mime = cover.mime
-    extension = mime_types[mime]
-
-    cover_file =
-      if
-        extension != ""
-      then
-        cover_file = file.temp("#{id}_", ".#{extension}")
-        file.write(cover_file, data=cover)
-        cover_file
-      else
-        null()
-      end
-    (cover_file, mime)
-  end
-
   def extract_cover_from_metadata(_metadata) =
     filename = _metadata["filename"]
     log.info(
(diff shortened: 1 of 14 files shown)


vitoyucepi:
```patch
diff --git a/src/libs/extra/metadata.liq b/src/libs/extra/metadata.liq
--- a/src/libs/extra/metadata.liq
+++ b/src/libs/extra/metadata.liq
@@ -50,7 +50,7 @@
     let (cover_file, mime) =
       null.case(cover, {(null(), "")}, save_cover_to_file)
 
-    next_cover =
+    current_cover_request :=
       if
         cover == null()
       then
@@ -79,7 +79,6 @@
         )
         request.create(temporary=true, cover_file)
       end
-    current_cover_request := next_cover
   end
 
   current_cover_request.{set=extract_cover_from_metadata}

```

src/libs/extra/metadata.liq
```diff
@@ -63,40 +30,42 @@ def file.cover.manager(
...
 
-    next_cover =
+    new_cover =
       if
-        cover == null()
-      then
-        log.important(
-          label=id,
-          "File #{string.quote(filename)} has no cover."
-        )
```
vitoyucepi, line 76: I'd like to preserve this log message.

src/libs/extra/metadata.liq
```diff
@@ -63,40 +30,42 @@ def file.cover.manager(
...
+              string.quote(cover_file)
+            }."
+          )
+
+          request.create(temporary=true, cover_file)
+        end
       end
-    set_current_cover(next_cover)
+
+    current_cover_request := (new_cover ?? default)
```
vitoyucepi, line 67: Does it make sense to return `null()`?

src/libs/extra/metadata.liq
toots: Done!

src/libs/extra/metadata.liq
toots: We need to reset to `default` when the file has no cover, right?

vitoyucepi: I'd like to point out that the previous if statement could return `default` instead of `null()`, so the whole if statement will assign a value to `current_cover_request`.

toots: It doesn't look like so. But if you find it more readable, we could return `default` from within the `if` statement:
```liquidsoap
    new_cover =
      if
        not null.defined(cover)
      then
        log.important(
          label=id,
          "File #{string.quote(filename)} has no cover."
        )
        null()
      else
        cover = null.get(cover)
        extension = mime_types[cover.mime]
        if
          extension == ""
        then
          log.important(
            label=id,
            "File #{string.quote(filename)} has unknown mime type #{
              string.quote(cover.mime)
            }."
          )
          null()
        else
          cover_file = file.temp("#{id}_", ".#{extension}")
          file.write(cover_file, data=cover)

          log.important(
            label=id,
            "Cover for #{string.quote(filename)} saved to #{
              string.quote(cover_file)
            }."
          )

          request.create(temporary=true, cover_file)
        end
      end
```

vitoyucepi: Never mind, it's fine.

PR #4087, opened 2024-08-12
Remove taglib, add more tags build config.

toots:
This PR removes support for `taglib`.

Taglib was historically important to support the myriad of formats out in the wild way back when. Nowadays, it seems that we only really need to support a core set, `id3v1/2`, `mp4`/`m4a` etc that is now natively supported via `ocaml-metadata`.

For exotic formats, `ocaml-ffmepg` should still provide support for those metadata.

Also, `taglib` requires a binary dependency on the C++ runtime library which is causing issues for the binary's portability and build. We also have had reports of straight crash with the (not yet supported) OCaml 5 compiler.

Lastly, the PR adds more details in the `Tags` section of the build config report to make it clear that we are still supporting a large variety of formats.


vitoyucepi:
Hi @toots,
Why would you want to change the bug report template? The `liquidsoap --build-config` output there is just a placeholder to show the reporter an example of the output.

toots:
> Hi @toots, Why would you want to change the bug report template? The `liquidsoap --build-config` output there is just a placeholder to show the reporter an example of the output.

I think that the bug report template should be as close as possible to the actual program output so that it will not confuse the user.

vitoyucepi:
> I think that the bug report template should be as close as possible to the actual program output so that it will not confuse the user.

I used the output of `savonet/liquidsoap:v2.2.5` in the placeholder, and I don't think it is necessary to update at all.
I can't say which is the most popular installation method, but I prefer
1. container;
1. repository package;
1. upstream package.

vitoyucepi (approved):

vitoyucepi (approved):

PR #4128, opened 2024-09-11
Dynamic source tracks.

toots:
There still are reports of issues with `source.tracks` forcing sources to compute their ` content_type` before we had a chance to assign their `frame_type`.

For reference:
* `frame_type` is a _type_ representing the source's content as computed during type unification, e.g. `{audio=pcm(stereo)}`, `{video=canvas, audio=pcm(5.1)}` etc. This value can be incomplete if no full restrictions are put on the source content. 

For instance, in:
```
s = blank()
output.dummy(s)
```
`s` can be any content because `output.dummy` does not care about the source's content. So, its `frame_type` is `'a`

* `content_type` is a  fully realized content value. It is calculated from the `frame_type`, adding default values.

For instance, a source with PCM audio without a specified number of channels, `{audio=pcm('a)}` is assigned a content type of `{audio=pcm(stereo)}`.

The problem is that there can be a competition between requesting the source's `content_type` and assigning the `frame_type`. If content type is requested before all the `frame_type` have been assigned then the result can be incompatible, e.g. assigning a `stereo` channel when the source should have had `5.1` channels.

The most typical case is when computing `source.tracks`. In the current implementation, the tracks are all extracted immediately at evaluation time which forces the computation of the source's content type right away.

In this PR, we introduce a _delayed_ `Source_tracks` runtime value. This value can be passed in-lieu of the actual record of source tracks and evaluated only when needed, e.g. in the muxer operator.

In order to make things work, we need to also introduce a notion of `dynamic_methods` so that we can do the following:

```liquidsoap
# s = blank();;
s :
  source('A)
  .{
...
  } where 'A is a set of internal tracks =
  <source(id=blank, frame_type=something that is a set of internal tracks>.{...}
# t = source.tracks(s);;
t :
  'A.{metadata : metadata, track_marks : track_marks}
  where 'A is a set of internal tracks = source.tracks(source=blank)
# s' = source({audio=t.audio});;
s' :
  source(audio='A)
  .{
...
  }
  where
    'A is a track and a track of type: pcm, pcm_s16, pcm_f32, canvas, metadata or track_marks =
  <source(id=source, frame_type={audio : 'A}
where
  'A is a track of type: pcm, pcm_s16, pcm_f32, canvas, metadata or track_marks and a track>.{...}
# t;;
- : 'A.{audio : 'B, metadata : metadata, track_marks : track_marks}
    where 'A is a set of internal tracks,
      'B is a track and a track of type: pcm, pcm_s16, pcm_f32, canvas, metadata or track_marks = source.tracks(source=blank)
```
It's a bit hacky but it seems to work fine.

### DIFF
--- a/src/lang/value.ml
+++ b/src/lang/value.ml
@@ -62,7 +62,7 @@ and t =
       pos : Pos.Option.t; [@hash.ignore]
       value : Custom.t; [@hash.ignore]
       methods : t Methods.t;
-      dynamic_methods : dynamic_methods option; [@hash.ignore]
+      dynamic_methods : dynamic_methods option;
       mutable flags : Flags.flags; [@hash.ignore]
     }
   | List of {
--- a/src/lang/value.ml
+++ b/src/lang/value.ml
@@ -57,6 +57,7 @@ and t =
       pos : Pos.Option.t; [@hash.ignore]
       value : Custom.t; [@hash.ignore]
       methods : t Methods.t;
+      hidden_methods : string list;
       dynamic_methods : (string -> t option) option; [@hash.ignore]
       mutable flags : Flags.flags; [@hash.ignore]
     }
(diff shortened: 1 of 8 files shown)


src/lang/value.ml
```diff
@@ -57,6 +62,7 @@ and t =
       pos : Pos.Option.t; [@hash.ignore]
       value : Custom.t; [@hash.ignore]
       methods : t Methods.t;
+      dynamic_methods : dynamic_methods option; [@hash.ignore]
```
toots, line 65:
```suggestion
      dynamic_methods : dynamic_methods option;
```

PR #4152, opened 2024-10-01
Fix append operator.

toots:


quentinmit:
Unfortunately this PR doesn't fix the append operator for me. With my simple repro (roughly what you have here as a test):

```
usic = chop(every=1., metadata=[("title", "music")], sine(amplitude=0.1, 440.))

def next(_) =
    print("next called")
    sine(amplitude=0.1, duration=.5, 880.)
end

s = append(music, next)
s.on_track(fun (t) -> print("on_track #{t}"))

output(s)
```

2.2.5 plays 1s of 440, then .5s of 880, then .5s of 880, etc.
2.3.0 just plays continuous 440 (although the log messages indicate that it's calling `next()`)
2.3.0 + this PR plays .5s of 880 immediately, then .5s of 440, then .5s of 880, etc. (i.e. it's not waiting for the first track boundary)

With a slightly more realistic test modeled on `source.say_metadata` (which I think is probably also broken), 2.3.0 with or without this PR completely skips the appended source:

```
# Sorry about this mess at the top, but I haven't been able to get any of the `say:` backends to work correctly, so I had to implement a new protocol for the `piper` TTS tool
def protocol_return(u) =
%ifversion >= 2.3
    u
%else
    [u]
%endif
end
PIPER_VOICES = environment.get("PIPER_VOICES")
if PIPER_VOICES != "" then
    def piper_protocol(~rlog=_,~maxtime=_,arg) =
        protocol_return(process.uri(extname="wav","echo #{process.quote(arg)} | piper -m $PIPER_VOICES/en/en_US/norman/medium/en_US-norman-medium.onnx -f $(output)"))
    end
    protocol.add("speak",piper_protocol,doc="Generate speech from text")
else
    protocol.add("speak",fun (~rlog=_,~maxtime=_,arg) -> protocol_return("say:#{arg}"))
end

music = chop(every=1., metadata=[("title", "music")], sine(amplitude=0.1, 440.))

def next(_) =
    print("next called")
    once(single("speak:in between songs"))
end

s = append(music, next)
s.on_track(fun (t) -> print("on_track #{t}"))

output(s)
```

On 2.2.5 that gives 1s of sine and then "in between songs" is spoken, etc. On 2.3.0 with or without this PR it just plays continuous sine.

PR #4156, opened 2024-10-03
source.dynamic improvements.

toots:


src/libs/request.liq
```diff
@@ -432,10 +432,20 @@ def request.player(~simultaneous=true) =
...
+      next_source := null()
+      s
+    end
+
+    s = source.dynamic(next)
 
     def play(r) =
-      s.set(request.once(r))
+      r = request.once(r)
+      s.prepare(s)
```
quentinmit, line 447: Did you mean `s.prepare(r)` here?

toots: Good catch thanks!
```suggestion
      s.prepare(r)
```

PR #4188, opened 2024-10-24
Change default values for fade out type and autocue loudness target

RM-FM:


Russsgithub:
I'm quite liking -16 personally , but can see the point of confirming to other standards.

RM-FM:
> I'm quite liking -16 personally , but can see the point of confirming to other standards.

If you prefer a -16 LUFS target, you can change it at any time by using `settings.autocue.internal.lufs_target := -16.0`

toots (approved):
Thanks!

PR #4216, opened 2024-11-19
Remove `source.{dump, drop}`, add `request.{dump, process}`.

toots:
This PR replaces the `source.{dump, drop}` mechanism with a safer one based on requests.

Now that these operators are being used more widely, via replaygain and autocue computations, it is clear that we need a better control over the potential for success before launching a CPU-burn clock on the underlying source.

With this PR, we introduce a new `request.{dump, drop}` set of operators that take a request and an optional processing function. From the request, a source is created using `request.dynamic` and passed to the process function. 

Then, before starting, the request source checks if the request can produce a decoder for its given content-type. If not, the processing is aborted right away. If it can, chances are pretty high that processing will sucesfully begin and terminate.

This makes these operators **fully synchronous** with no need for a special queue. They can also be launched from the CLI and terminate right away.

Lastly, the old `source.{dump, drop}` is make possible via the export of `clock.tick`. In this case, advanced users can implement the behavior if they know what they are doing:

```liquidsoap
      clock.assign_new(sync="passive", [s])

      stopped = ref(false)
      output.file(
        fallible=true, on_stop={stopped.set(true)}, %wav, file, once(s)
      )

      c = clock(s.clock)
      c.start()
      while not stopped() do c.tick() end
      c.stop()
```


Russsgithub:
@toots .
I have been running 2.3 on a stream that mixes 4 request dynamic sources dependent on metadata triggers and stream events, also uses sound touch and ladspa.
Just thought I'd report that all seems to be working and no longer collapsing to silence after a few songs ( which has been happening up until yesterday ).
:). 👍.

I'll report if anything unintended occurs.
Thanks.

toots:
> @toots . I have been running 2.3 on a stream that mixes 4 request dynamic sources dependent on metadata triggers and stream events, also uses sound touch and ladspa. Just thought I'd report that all seems to be working and no longer collapsing to silence after a few songs ( which has been happening up until yesterday ). :). 👍.
> 
> I'll report if anything unintended occurs. Thanks.

Awesome!

Issue #4248, opened 2024-12-09
YAML parsing and type matching


martinkirch:
### Description

YAML parsing is a great feature, allowing my Liquidsoap script to share the same configuration as my companion process. However I'm blocked by type checks at runtime.

### Steps to reproduce

Using the example from tests:
```
data = '---
record:
  a: 1
  b: ert
key_with_methods: value
bla: bar
baz: 3.14
'
print(data)

let yaml.parse (x :
    {
      bla: string,
      baz: float,
      key_with_methods: string,
      record: {a: float, b: string}
    }
  ) = data
# runtime error: Parse error: yaml value cannot be parsed as type: \n{\n  baz : float,\n  bla : string,\n  key_with_methods : string,\n  record : \n  {a : float, b : string\n  }\n}
print(x)
```

I tried a few variants, including without the [document separator](https://yaml.org/spec/1.2.2/#22-structures) and a simpler decomposition as in the documentation:
```
data = '
key_with_methods: value
bla: bar
baz: 3.14
'
print(data)

let yaml.parse ({bla, baz, key_with_methods} :
    {
      bla: string,
      baz: float,
      key_with_methods: string,
    }
  ) = data
# Parse error: yaml value cannot be parsed as type: {baz : float, bla : string, key_with_methods : string}
print(baz)
```

### Expected behavior

It would also be great to have a clearer error message, or being able to know which type has been parsed.

### Liquidsoap version

```text
Liquidsoap rolling-release-v2.3.x-52-g83173fa+dev
```

### Liquidsoap build config

```text
built on Fedora 41

 * Liquidsoap version  : rolling-release-v2.3.x-52-g83173fa+dev

 * Compilation options
   - Release build       : false
   - Git SHA             : (none)
   - OCaml version       : 5.2.1
   - OS type             : Unix
[...]

PR #4249, opened 2024-12-09
let go YAML's `not_found` exception

martinkirch:
this would avoid #4248, where I didn't notice that `yaml` was an optional dependency

### DIFF
--- a/src/lang/builtins_yaml.ml
+++ b/src/lang/builtins_yaml.ml
@@ -41,7 +41,8 @@ let _ =
       with exn -> (
         let bt = Printexc.get_raw_backtrace () in
         match exn with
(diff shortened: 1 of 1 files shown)


toots:
Thanks for the PR! I'm not sure if that would do anything. This file is simply ignored when the `yaml` optional opam package is not present so anything in its code is also silenced.

martinkirch:
I tested it by running `opam uninstall yaml`. It recompiled my pinned liquidsoap (that's how I usually test the `main` branch) and the exception was still raised.

toots:
My bad, I forgot this code was organized in a different way.. merging, thanks!

PR #4250, opened 2024-12-10
Reimplement audioscrobbler natively

toots:


gAlleb:
Very cool, @toots, thank you!

It uses a source all the way now unlike (on_metadata) in the past. Is there a way to avoid scrobbling jingles? Like I did with previous version:

```
def lastfm.submit(~user,~password,~source="broadcast",~length=false) = 

fun (m) ->

if (m["jingle_mode"] != "true") then

  audioscrobbler.submit(user=user,password=password,
                source=source,length=length,
                host="post.audioscrobbler.com",port=80,
                m)

end

end
```

toots:
> Very cool, @toots, thank you!
> 
> It uses a source all the way now unlike (on_metadata) in the past. Is there a way to avoid scrobbling jingles? Like I did with previous version:
> 
> ```
> def lastfm.submit(~user,~password,~source="broadcast",~length=false) = 
> 
> fun (m) ->
> 
> if (m["jingle_mode"] != "true") then
> 
>   audioscrobbler.submit(user=user,password=password,
>                 source=source,length=length,
>                 host="post.audioscrobbler.com",port=80,
>                 m)
> 
> end
> 
> end
> ```

Great idea!

I'm gonna add an optional metadata pre-processing callback. You should be able to bypass the process by returning an empty metadata list.

gAlleb:
Cool! Thanks!

gAlleb:
Thanks @toots! tried simple script:

```
settings.audioscrobbler.api_key := "My Key"
settings.audioscrobbler.api_secret := "Secret"

s = playlist("...")

s = audioscrobbler.submit(
  username="toots5446",
  password="hackme", 
 s
)

output.ao(fallible=true, s)
```

```ruby
 % liquidsoap lastfm.liq
At /Users/stef/.opam/liq/share/liquidsoap-lang/libs/extra/audioscrobbler.liq, line 101 char 4 - line 104 char 5:
    error.raise(
      error.invalid,
      "Invalid response: #{resp}, error: #{err}"
    )

Error 14: Uncaught runtime error:
type: invalid,
message: "Invalid response: <?xml version=\"1.0\" encoding=\"UTF-8\"?>\n<lfm status=\"ok\">\n  <session>\n    <name>Rock_at_omFM</name>\n    <key>8PbqAswYawEXAMPLEHfeZb5_7dRSq</key>\n    <subscriber>0</subscriber>\n  </session>\n</lfm>\n, error: error(kind=\"assertion\",message=\"Assertion failed.\",positions=\"at ./lastfm.liq, line 10 char 8 - line 14 char 1, at /Users/stef/.opam/liq/share/liquidsoap-lang/libs/extra/audioscrobbler.liq, line 423 char 4 - line 428 char 5, at /Users/stef/.opam/liq/share/liquidsoap-lang/libs/extra/audioscrobbler.liq, line 94 char 2 - line 105 char 5\")",
stack: at /Users/stef/.opam/liq/share/liquidsoap-lang/libs/extra/audioscrobbler.liq, line 101 char 4 - line 104 char 5,
at /Users/stef/.opam/liq/share/liquidsoap-lang/libs/extra/audioscrobbler.liq, line 101 char 4 - line 104 char 5,
at /Users/stef/.opam/liq/share/liquidsoap-lang/libs/extra/audioscrobbler.liq, line 94 char 2 - line 105 char 5,
at /Users/stef/.opam/liq/share/liquidsoap-lang/libs/extra/audioscrobbler.liq, line 423 char 4 - line 428 char 5,
at ./lastfm.liq, line 10 char 8 - line 14 char 1

```

toots:
Thanks for the feedback @gAlleb. I have removed the `assert` can you try now?

PS: You might want to disconnect your application here: https://www.last.fm/settings/applications and get a new session key.

gAlleb:
Thanks, @toots. Works now and scrobbles!

However I'm a bit stuck with `jingle_mode`. In order to bypass scrobbling what shall I do?

1. I can map matadata and remove `artist` and `title` in jingle-tracks before applying audioscrobbler. Is this a usecase?
It works but what if I need this metadata further down the code chain. (For example some functions for sending statistics.)

2. I can edit `audioscrobbler.api.apply_meta` function to accommodate check for m["jingle_mode']. 
This works but looks too hacky).

How can I disable scrobbling if m["jingle_mode"] metadata is found in a right way? (disregarding empty/not-empty artist/title)

toots:
I added a `metadata_preprocessor` to the `audioscrobbler.submit` function. If you return an empty list `[]` there it will disable the scrobbling process but won't affect your stream's metadata.

toots:
See: https://www.liquidsoap.info/doc-dev/reference-extras.html#audioscrobbler.submit

gAlleb:
Could you please give me an example, I'm failing at getting this to work.

Also I've encountered unexpected behavior: When you scrobble tracks — all of them are scrobbled at the exact time/time ago.

Снимок экрана 2024-12-14 в 20 20 26

All of the above tracks have been played through from start to finish - so the time of scrobble should be different.

gAlleb:
I could only think of

```
def jingle_check(m) =
   jingle = m["jingle_mode"]
      if jingle == "true" then
         []
      else
        m
     end
end

radio = audioscrobbler.submit(
  username="",
  password="", 
  metadata_preprocessor=(fun(m) -> jingle_check(m)),
 radio
)
```

Something like this?

<hr/>

Also I've encountered unexpected behavior: When you scrobble tracks — all of them are scrobbled at the exact same time/time ago. It is the time when first track has been scrobbled since the start of LS.

Снимок экрана 2024-12-14 в 20 20 26

All of the above tracks have been played through from start to finish - so the time of scrobble should be different.

toots:
Great catch on the optional arg. This is a weird historical thing that we evaluate optional argument values at script eval not at function eval. Something we could change, I'll ask David why we did that.

Yes, returning empty metadata disables the process. It needs at least track title and artist to proceed. Think that this is user-friend enough?

gAlleb:
Well, I got an idea right away but it took me some time to get the syntax right with `metadata_preprocessor=`

`metadata_preprocessor=(fun(m) -> if m["jingle_mode"] == "true" then [] else m end),`

It feels ok now :)

toots:
Ok cool. We could also allow the function to return `null()` which would be more explicit.

gAlleb:
I think [] is fine too.

Issue #4275, opened 2024-12-18
file.watch() does not work


brenc:
### Description

This code does not work:

```
file.watch(
  "/tmp/testfile",
  {print("testfile changed")},
)
```

I don't think this worked in 2.2 either (or earlier).

### Steps to reproduce

Run the provided code then `echo TEST >> /tmp/testfile`.

### Expected behavior

For the provided code to produce "testfile changed" when it's changed.

### Liquidsoap version

```text
2.3.0
```

### Liquidsoap build config

```text
* Liquidsoap version  : 2.3.0

 * Compilation options
   - Release build       : true
   - Git SHA             : (none)
   - OCaml version       : 4.14.2
   - OS type             : Unix
[...]

PR #4278, opened 2024-12-20
Fix inotify GC cleanup removing the watch function

toots:
Some of the changes to optimize liquidsoap's memory footprint did remove unused methods. This might have triggered a situation where we acutally removes the watch call during a GC of the `unwatch` method.

Instead, we should require the user to explicitely call `unwatch`.

Fixes: #4275

### DIFF
--- a/src/core/file_watcher.inotify.ml
+++ b/src/core/file_watcher.inotify.ml
@@ -82,7 +82,7 @@ let watch : watch =
            with exn ->
              let bt = Printexc.get_backtrace () in
              Utils.log_exception ~log ~bt
-               (Printf.sprintf "Error whole removing file watch handler: %s"
+               (Printf.sprintf "Error while removing file watch handler: %s"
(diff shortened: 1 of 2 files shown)


src/core/file_watcher.inotify.ml
```diff
@@ -77,17 +77,12 @@ let watch : watch =
...
-            handlers := List.remove_assoc wd !handlers)
-      in
-      Gc.finalise finalise unwatch;
-      unwatch)
+      Mutex_utils.mutexify m (fun () ->
+          (try Inotify.rm_watch fd wd
+           with exn ->
+             let bt = Printexc.get_backtrace () in
+             Utils.log_exception ~log ~bt
+               (Printf.sprintf "Error whole removing file watch handler: %s"
```
toots, line 85:
```suggestion
               (Printf.sprintf "Error while removing file watch handler: %s"
```

PR #4298, opened 2025-01-05
docs: Fix Global Settings in `migrating.md`

vitoyucepi:
## Description
This PR updates the `migrating.md` file to fix how to enable the old `on_air` metadata. The previous text did not mention the global `settings` object, which is the recommended way to configure the application.


toots:
Thanks!

PR #4308, opened 2025-01-08
Allow disabling normalize.

smimram:
We cannot simply use a switch to diable normalize because of the delay line...

Refs #4305.


toots:
Can you add a `CHANGES.md` entry for it?

PR #4518, opened 2025-06-15
Add warning when erasing top-level variables

toots:
This PR adds a new warning when top-level variables as defined in the whole standard library are overidden.

It works by taking all the environment variable names from the standard library and testing if any of these names are overridden.

This is done during the typing phase so that it can be printed (or raised) before the program starts.

The `--strict` option can be used to raise errors when these warnings are trigered.


Tampa:
Would it be possible to add an output that states exactly which variable it is that caused the error?

toots:
> Would it be possible to add an output that states exactly which variable it is that caused the error?

It does already. However, I have just pushed a commit that will make it also highlight the exact variable in a function definition that caused the override:

Screenshot 2025-07-06 at 10 28 29 AM

PR #4526, opened 2025-06-19
Advanced function argument syntaxes.

toots:
Before merging https://github.com/savonet/liquidsoap/pull/4518, I'd like to be able to accept a named argument for a function using one of the protected top-level names but rename it to a safe variable name inside the function. Something like:

```liquidsoap
def f(~request:r) =
  print(request.uri(r))
end
```

This PR cleans up the current function arguments. It:
* Deprecates `def f(~foo=_) = ... end`. This syntax is clashing with default values: `def f(~foo=123) = foo end`.
* Adds the combination of typed and renamed named argument
* Documents the full (?) list of possible syntaxes. Here it is:
```liquidsoap
# Ignored anonymous argument
def f(_) = 123 end
# f : ('a) -> int = fun (_) -> 123

# Typed anonymous argument
def f((foo:int)) = foo end
# f : (int) -> int = <fun>

# Anonymous argument with default value
def f(foo = 123) = foo end
# f : (?int) -> int = <fun>

# Typed ignored anonymous argument
def f((_:int)) = 123 end
# f : (int) -> int = fun (_) -> 123

# Typed anonymous argument with default value
def f((foo:int) = 123) = foo end
# f : (?int) -> int = <fun>

# Ignored anonymous argument with default value
def f(_ = 123) = 456 end
# f : (?int) -> int = fun (_=123) -> 456

# Typed ignored anonymous argument with default value
def f((_:int) = 123) = 456 end
# f : (?int) -> int = fun (_=123) -> 456

# Typed named argument
def f(~(foo:int)) = foo end
# f : (foo : int) -> int = <fun>

# Named argument with rename
def f(~foo:bla) = bla end
# f : (foo : 'a) -> 'a = <fun>

# Ignored named argument
def f(~foo:_) = 123 end
# f : (foo : 'a) -> int = fun (~foo=_) -> 123

# Named argument with default value
def f(~foo=123) = foo end
# f : (?foo : int) -> int = <fun>

# Typed named argument with rename
def f(~foo:(bla:int)) = bla end
# f : (foo : int) -> int = <fun>

# Typed named argument with default value
def f(~(foo:int)=123) = foo end
# f : (?foo : int) -> int = <fun>

# Typed named argument with rename and default value
def f(~foo:(bla:int)=123) = bla end
# f : (?foo : int) -> int = <fun>

# Typed ignored named argument with default value
def f(~foo:(_:int)=123) = 456 end
# f : (?foo : int) -> int = fun (~foo=123) -> 456

# Ignored argument with default value
def f(~foo:_=123) = 456 end
# f : (?foo : int) -> int = fun (~foo=123) -> 456
```

### DIFF
--- a/doc/content/language.md
+++ b/doc/content/language.md
@@ -753,6 +753,10 @@ def f((_:int)) = 123 end
 def f((foo:int) = 123) = foo end
 # f : (?int) -> int = <fun>
 
+# Ignored anonymous argument with default value
+def f(_ = 123) = 456 end
+# f : (?int) -> int = fun (_=123) -> 456
+
 # Typed ignored anonymous argument with default value
 def f((_:int) = 123) = 456 end
 # f : (?int) -> int = fun (_=123) -> 456
--- a/doc/content/language.md
+++ b/doc/content/language.md
@@ -725,6 +725,71 @@ corresponding label with "`?`", so that the type of the above function is
 (samples : float, ?duration : float) -> float
 ```
 
+### Advanced argument syntax
+
+Arguments can be ignored, typed, named (and renamed) and given default values
+and all these possibilities can be combined.
+
+Here is the syntax to do it:
+
+```liquidsoap
+# Ignored anonymous argument
+def f(_) = 123 end
+# f : ('a) -> int = fun (_) -> 123
+
+# Typed anonymous argument
+def f((foo:int)) = foo end
+# f : (int) -> int = <fun>
+
+# Anonymous argument with default value
+def f(foo = 123) = foo end
+# f : (?int) -> int = <fun>
+
+# Typed ignored anonymous argument
+def f((_:int)) = 123 end
+# f : (int) -> int = fun (_) -> 123
+
+# Typed anonymous argument with default value
+def f((foo:int) = 123) = foo end
+# f : (?int) -> int = <fun>
+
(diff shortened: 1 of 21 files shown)


toots:
@vitoyucepi any feedback on the syntax here?

vitoyucepi:
I'm not sure if the renaming feature is necessary, and the parentheses look unintuitive. I can't explain exactly what's wrong, but it looks awkward.
Why `def f(~(foo:int))` instead of `def f(~foo:int)`?

toots:
> I'm not sure if the renaming feature is necessary, and the parentheses look unintuitive. I can't explain exactly what's wrong, but it looks awkward. Why `def f(~(foo:int))` instead of `def f(~foo:int)`?

You definitely need the rewrite, typically you want to accept arguments with a meaningful name but use a shorthand inside the function's body to avoid erasing top-level modules, for instance:

```liquidsoap
# that's a rename
def f(~request:r) =
  request.uri(r)
end
```

You also need the type annotation for those cases where the function's code is not enough to infer the foll type like:

```liquidsoap
# that's a type annotation 
def f(~(arg:int)) =
  print(arg)
end
```

`(arg:type)` is already the syntactic construct for type annotations so it makes sense to keep it.

Lastly, all these combinations are probably not very likely to be used together but it's good to know that they can be used if needed..

PR #4541, opened 2025-06-29
Move insert_metadata to source method.

toots:
This PR deprecates the `insert_metadata` operator and moves the implementation to the source class, making `insert_metadata` a source method available with every source.


src/core/lang_source.ml
```diff
@@ -255,6 +255,24 @@ let source_methods =
        source is currently streaming, just that its resources are all properly \
        initialized.",
       fun s -> val_fun [] (fun _ -> bool s#is_ready) );
+    ( "insert_metadata",
+      ( [],
+        Lang.fun_t
+          [(true, "new_track", Lang.bool_t); (false, "", metadata_t)]
+          Lang.unit_t ),
+      "Dynamically insert metadata in a stream. Inserts a new track with the \
+       given metadata if `new_track is `true`.",
```
toots, line 264:
```suggestion
       given metadata if `new_track` is `true`.",
```

PR #4562, opened 2025-07-11
Allow destructuring function arguments!

toots:
This PR adds support for pattern matching inside function arguments:

```liquidsoap
  x = 123
  def f({foo}, [a, b], ~x:{gni}) =
    test.equal(foo, 123)
    test.equal(a, "aaa")
    test.equal(b, "bbb")
    test.equal(x, 123)
    test.equal(gni, 3.14)
  end
  f({foo=123}, ["aaa", "bbb"], x={gni=3.14})
```

### DIFF
--- a/src/lang/parser.mly
+++ b/src/lang/parser.mly
@@ -523,18 +523,18 @@ arg:
                    `Term {label = $3; as_variable = None; typ =  Some $5; default = $7; annotations = []; pos = $loc($3) }
                  }
   | TILD VAR GETS UNDERSCORE opt {
-                    `Term {label = $2; as_variable = Some "_"; typ = None; default = $5;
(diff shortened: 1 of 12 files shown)


src/lang/parser.mly
```diff
@@ -523,18 +523,18 @@ arg:
                    `Term {label = $3; as_variable = None; typ =  Some $5; default = $7; annotations = []; pos = $loc($3) }
                  }
   | TILD VAR GETS UNDERSCORE opt {
-                    `Term {label = $2; as_variable = Some "_"; typ = None; default = $5;
+                   `Term {label = $2; as_variable = Some { pat_pos = $loc($4); pat_entry = `PVar ["_"] }; typ = None; default = $5;
```
toots, line 526:
```suggestion
                    `Term {label = $2; as_variable = Some { pat_pos = $loc($4); pat_entry = `PVar ["_"] }; typ = None; default = $5;
```

Issue #4494, opened 2025-05-27
Unable to build Liquidsoap v2.3.3 with FFmpeg


vitoyucepi:
### Description

Liquidsoap v2.3.3 cannot be built with ffmpeg because errors occur during the build process.

### Steps to reproduce

1. `alpine.containerfile` 
   ```containerfile
   FROM alpine:3.21
   
   ARG OCAML_VERSION=4.14.2
   
   RUN apk add --no-cache \
         bash \
         curl \
         ca-certificates \
         patch \
         unzip \
         git \
         bzip2 \
         gcc \
         make \
         g++ \
         opam \
       ;
   
   RUN opam init --bare --disable-sandboxing -n;
   RUN opam switch create "$OCAML_VERSION";
   RUN git clone https://github.com/savonet/liquidsoap.git -b v2.3.3;
   RUN set -eux; \
         cd liquidsoap; \
         opam pin --no-action --yes .;
   RUN apk add --no-cache linux-headers curl-dev ffmpeg-dev;
   RUN opam install --confirm-level=unsafe-yes --no-depexts liquidsoap ffmpeg; \
       cat /root/.opam/log/liquidsoap*.out;
   ```
1. `ubuntu.containerfile`
   ```containerfile
   FROM ubuntu:24.04
   
   ARG OCAML_VERSION=4.14.2
   ARG DEBIAN_FRONTEND=noninteractive
   
   RUN set -eux; \
         apt-get update; \
         apt-get install --no-install-recommends -y \
           curl \
           ca-certificates \
           git \
           gcc \
           g++ \
           make \
           opam \
         ;
   
   RUN opam init --bare --disable-sandboxing -n;
   RUN opam switch create "$OCAML_VERSION";
   RUN git clone https://github.com/savonet/liquidsoap.git -b v2.3.3;
   RUN set -eux; \
         cd liquidsoap; \
         opam pin --no-action --yes .;
   RUN opam install --confirm-level=unsafe-yes liquidsoap ffmpeg; \
       cat /root/.opam/log/liquidsoap*.out;
   ```
1. `docker build --progress plain -f alpine.containerfile .`

### Expected behavior

I expect that liquidsoap builds with FFmpeg.

### Log extracts

<details>
<summary>Warnings</summary>

```
[...]

PR #4576, opened 2025-07-20
fix: Increase the minimum ocaml-ffmpeg version

vitoyucepi:
## Summary

Liquidsoap is not compatible with [`ffmpeg:1.2.4`](https://github.com/savonet/ocaml-ffmpeg), so the minimum version should be set to `1.2.5`.

Fixes: #4494.

## Additionally

Should I create a PR at https://github.com/ocaml/opam-repository as well?


toots:
Thanks for this. Yes, a PR with opam would be great too.

Issue #4194, opened 2024-10-27
interactive variables not updating due to no default values in list.assoc


Russsgithub:
### Description

Using interactive.bool results in 

```
Could not update variable  (not_found: no default value for list.assoc).


(no other information is given in the error to determine where this is occuring, but i have no calls to list.assoc in my script.)
```

When i update the variable through the :8000/interactive interface.

I believe this also occurs with the other interactive operators.

### Steps to reproduce

s =sine(440.)

bool = interactive.bool("toggle", false)

output.dummy(fallible=true, s)

### Expected behavior

bool respond to external call to change.

### Liquidsoap version

```text
Liquidsoap rolling-release-v2.3.x-54-g63aea1b+dev
Copyright (c) 2003-2024 Savonet team
Liquidsoap is open-source software, released under GNU General Public License.
See <http://liquidsoap.info> for more information.
```

### Liquidsoap build config

```text
* Liquidsoap version  : rolling-release-v2.3.x-54-g63aea1b+dev

 * Compilation options
   - Release build       : false
   - Git SHA             : (none)
   - OCaml version       : 4.14.2
   - OS type             : Unix
   - Libs versions       : angstrom=0.16.1 asetmap=0.8.1 asn1-combinators=0.3.2 astring=0.8.5 base=v0.16.3 base.base_internalhash_types=v0.16.3 base.caml=v0.16.3 base.shadow_stdlib=v0.16.3 base64=
3.5.1 bigarray=[distributed with Ocaml] bigarray-compat=1.1.0 bigstringaf=0.10.0 bos=0.2.1 bytes=[distributed with OCaml 4.02 or above] ca-certs=v1.0.0 camlp-streams camomile.lib=2.0 cohttp=5.3.1 
cohttp-lwt=5.3.0 cohttp-lwt-unix=5.3.0 conduit=7.1.0 conduit-lwt=7.1.0 conduit-lwt-unix=7.1.0 cry=1.0.3 cstruct=6.2.0 ctypes=0.23.0 ctypes-foreign=0.23.0 curl=0.9.2 digestif.c=1.2.0 domain-name=0.
4.0 domain_shims dtools=0.4.5 dune-build-info=3.16.0 dune-private-libs.dune-section=3.16.0 dune-site=3.16.0 dune-site.private=3.16.0 duppy=0.9.4 eqaf=0.10 ffmpeg-av=1.2.0 ffmpeg-avcodec=1.2.0 ffmp
[...]

PR #4592, opened 2025-07-28
fix: Prevent trailing '&' in request payload for when updating interactive variables

myeungdev:
First of all, thank you for this fantastic project! 

Currently, whenever an interactive variable is updated through the web interface, Liquidsoap throws the following error, even though the update is successful:

```
  [interactive.harbor:3] Could not update variable  (not_found: no default value for list.assoc).
```

This was very confusing for me, as it led me to believe my variables were not being updated successfully, and I ended up spending a lot of time debugging my script. After some investigation, I discovered that the error occurs because Liquidsoap attempts to find a variable with an empty string name after each update request, caused by a trailing `&` in the request body.

This pull request adds a conditional check to only append `&` if it’s not the last parameter, preventing this issue from happening again.

This error also caused confusion for one of the users:
https://github.com/savonet/liquidsoap/issues/4194

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -36,6 +36,7 @@ Fixed:
 
 - Fix error when loading script path having non-ascii characters in them
   (#4343)
+- Fix error with interactive variables (#4592, @myeungdev)
 - Prevent concurrent reload and request fetch in playlists (#4316)
 - Don't mark source as ready until their clock has started. (#4496)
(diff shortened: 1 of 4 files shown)


toots:
Thanks for this! I also added a fix to remove empty entries from our underlying API that is used to parse url arguments!

Issue #4589, opened 2025-07-28
http.transport.tls client certificate support


DelilahHoare:
### Description

I'd like to be able to use client certificates for https connections.  I'd be happy to implement this myself.

### Preferable solution

`http.transport.tls` has new arguments `client_certificate` (type `string?`, default `null`) and `client_key` (type `string?`, default `null`).

For a client, `client_certificate` is the client certificate to use, and `client_key` is its private key.

For a server, `client_certificate` contains acceptable client certificates/CAs, and `client_key` is ignored.  If `client_certificate` is non-`null`, clients are required to use a valid client certificate.

### Alternative solutions

It would be nice to provide client certificate details to the `input.harbor` `auth` callback, and allow it to accept connections without a valid client certificate, along the lines of `SSLVerifyClient optional` in Apache `mod_ssl`.  This could be added later.

### Additional context

It looks like ocaml-tls supports client certificates and ocaml-ssl does not.

toots:
Hi! Sorry for the late response. This would be a welcome new feature. I'm happy to help with the codebase if you would like to contribute it.

PR #4616, opened 2025-08-09
Add client certificate support to http.transport.tls

DelilahHoare:
Implements #4589

### DIFF
--- a/src/core/builtins/builtins_tls.ml
+++ b/src/core/builtins/builtins_tls.ml
@@ -312,11 +312,17 @@ let _ =
       ( "client_certificate",
         Lang.nullable_t Lang.string_t,
         Some Lang.null,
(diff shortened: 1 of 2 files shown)


src/core/builtins/builtins_tls.ml
```diff
@@ -279,6 +309,14 @@ let _ =
         Some
           "Path to certificate private key. Required in server mode, e.g. \
            `input.harbor`, etc. Unused in client mode." );
+      ( "client_certificate",
+        Lang.nullable_t Lang.string_t,
+        Some Lang.null,
+        Some "Path to client certificate file." );
```
toots, line 315: Could you explain what the client certificate does here?

toots:
Thanks!

PR #4628, opened 2025-08-15
Explicit source cleanup.

toots:


codeofdusk:
@toots Given a script like:

```
music1 = replaygain(playlist("path/to/music1",id="music1",reload_mode="watch"))
music2 = replaygain(playlist("/path/to/music2",id="music2",reload_mode="watch"))
security = single("/path/to/default.ogg")
# ...
radio = fallback(track_sensitive = true, [random([music1, music2]),security])
```

Before this change, at startup, `default.ogg` plays, then we go into music.

After this change, `default.ogg` does not play at startup.

toots:
@codeofdusk thanks for letting me know. Is that a problem for you?

codeofdusk:
@toots Not a major one I guess, but it is a noticeable change in behaviour and I'd say worth fixing if you can or at least documenting.

PR #4684, opened 2025-09-25
Unbreak notify_metadata

parnikkapore:
Using notify_metadata causes a typecheck error in Liquidsoap 2.4.0; this adds the now-required synchronous= argument.


toots:
So sorry I missed that. Thank you for contributing!

Issue #4708, opened 2025-10-23
Difficult to reproduce, possible heap corruption


xogium:
### Description

When running Liquidsoap after a somewhat random time (usually a very long) time, it crashes after trying to allocate terabytes of memory during garbage collection after reading an invalid value from the heap. My current investigation points to autocue being a possible cause for this bug, or at the very least a way to accelerate things. Despite this, it is still extremely difficult to reproduce and might not trigger for several weeks, even months in a row.

### Steps to reproduce

No clear steps to reproduce at the moment. I will be providing as much data as possible. Right now I'm testing running my configuration with short clips of songs rather than their full songs. This may or may not help.

### Expected behavior

Liquidsoap should not crash.

### Log extracts

May be completely unrelated. Both the script with use of autocue and with cueing metadata already embedded in have these lines, for every new track that plays:
```
2025/10/18 09:51:09 [request.4:3] End of track reached at 0.70 before cue-out point at 279.30!
2025/10/18 09:55:45 [request.8:3] End of track reached at 0.32 before cue-out point at 108.10!
2025/10/18 09:57:27 [request.12:3] End of track reached at 5.02 before cue-out point at 181.70!
```

### Script extracts

This script seems to cause the issue. It uses autocue along with crossfade.
```
music = playlist("/var/lib/liquidsoap/music",reload_mode="watch",prefix="autocue:")
def crossfading(old, new) =
    add(normalize=false,
      [fade.in(initial_metadata=new.metadata, duration=.1, new.source),
       fade.out(initial_metadata=old.metadata, duration=2.5, old.source)]
    )
end

def append_album(m) =
	title = m["title"]
	artist = m["artist"]
	album = m["album"]
	[("artist","#{artist}"),("title","#{album} - #{title}")]
end

music = map_metadata(append_album, music)
music = amplify(1.,override="replaygain_track_gain",music)
radio = mksafe(cross(duration=2.5, width=2.0, crossfading, music))

output.icecast(%fdkaac(channels=2, samplerate=44100, bitrate=192, afterburner=true, aot="mpeg4_aac_lc", sbr_mode=true),
transport = http.transport.tls(),
        host = "domain.tld",
       port = 8443,
	mount = "/music",
	password = "redacted",
radio)
```

This other script, which doesn't use autocue but does use crossfade, seems to be fine. It has not crashed since may. Perhaps it is still possible for it to do so, but that autocue agitates the bug just enough to make it a little bit more likely to happen?
```
[...]

PR #4714, opened 2025-10-27
Fix CAMLparam definitions in C stubs

Jookia:
Possible fix for https://github.com/savonet/liquidsoap/issues/4708

### DIFF
--- a/src/core/operators/defer_c.c
+++ b/src/core/operators/defer_c.c
@@ -6,7 +6,7 @@
 
 // Single bigarray not registered with the GC.
 CAMLprim value liquidsoap_alloc_managed_int16_ba(value _len) {
-  CAMLparam0();
+  CAMLparam1(_len);
   intnat out_size = Int_val(_len);
   void *data = malloc(out_size * caml_ba_element_size[CAML_BA_SINT16]);
   if (!data)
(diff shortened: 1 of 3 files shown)


src/core/operators/defer_c.c
```diff
@@ -6,7 +6,7 @@
 
 // Single bigarray not registered with the GC.
 CAMLprim value liquidsoap_alloc_managed_int16_ba(value _len) {
-  CAMLparam0();
+  CAMLparam1(_len);
```
toots: Integers don't need to be registered with the GC because they don't hold any allocated memory!

src/modules/duppy/duppy_stubs.c
```diff
@@ -151,7 +151,7 @@ CAMLprim value caml_poll(value _read, value _write, value _err, value _timeout)
 
 CAMLprim value ocaml_duppy_write_ba(value _fd, value ba, value _ofs, value _len)
 {
-  CAMLparam2(ba,_fd) ;
+  CAMLparam4(_fd, ba, _ofs, _len);
```
toots: Same!

Jookia:
I'm aware these won't fix the issue but is there a downside to making the code compliant with OCaml rules?

PR #4730, opened 2025-11-09
Add binary value tag, use it to hide cover metadata from printing.

toots:
This should solve a long-term pain point when working with coverart metadata.


smimram:
Nice! Maybe could it be useful to have functions to set the binary flag directly from liq? (something like `string.set_binary`)

toots:
> Nice! Maybe could it be useful to have functions to set the binary flag directly from liq? (something like `string.set_binary`)

Good idea.

Done!

PR #4732, opened 2025-11-11
Allow selecting portaudio device by name.

### LINKED ISSUE #4731: Support selecting PortAudio devices by name instead of numeric ID
### Description

Currently, the `input.portaudio` source in Liquidsoap only supports selecting devices by their numeric `device_id`.  
This is problematic in real-world setups where device IDs are not stable across reboots or hardware changes.

For example, the following works today:

```liquidsoap
source = input.portaudio(device_id=3)
```
…but the numeric ID 3 can change after a VM reboot, host migration, or even when a new audio device is added to the system.
The --list-portaudio-devices output shows user-friendly names like:

Device ID 1:
- name: TONOS COSQUIN (AXIA IP-Driver)
- host API: 0
- max input channels: 2
- default sample rate: 44100

However, there is no way to select a device by name, e.g.:
```
source = input.portaudio(device_name="TONOS COSQUIN (AXIA IP-Driver)")
```

This feature would make Liquidsoap scripts much more robust and portable.

### Preferable solution

Add a new optional argument to input.portaudio:
```
input.portaudio(?device_name : string?, ?device_id : int?, ...)
```

If device_name is provided, Liquidsoap should internally resolve it to the matching PortAudio device ID before initializing the stream.
If both are provided, device_id should take precedence.

Example use case
```
source = input.portaudio(device_name="TONOS COSQUIN (AXIA IP-Driver)")
output.icecast(%mp3, host="localhost", port=8000, password="hackme", mount="/mic")
```
This ensures that the script will work even if the PortAudio device order changes.

### Alternative solutions

_No response_

### Additional context

Why this is needed

Device IDs are volatile: On Windows, macOS, and virtualized environments (VMware, Parallels, etc.), PortAudio device IDs can change after every boot.

Automation & deployment: For automated setups (e.g. Docker, Kubernetes, or virtual studio pipelines), hardcoding numeric IDs is fragile.

User experience: For operators configuring multiple sources (e.g., "Mic 1", "Line In", "Axia IP Driver"), names are the natural way to identify devices.

Cross-platform stability: Device names are consistent across sessions, unlike IDs.

smimram:
Fixes #4731.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -11,6 +11,7 @@
 - Added `json.value` to make it possible to mix different value
   types when returning json values (#4712)
 - Allow referring to pulseaudio devices by name (#4732)
+- `output.shoutcast`: make `dj` argument a getter.
 
 ## Fixed:
 
--- a/src/core/request.ml
+++ b/src/core/request.ml
@@ -346,7 +346,7 @@ let conf_recode_excluded =
   Dtools.Conf.list
     ~d:["pic"; "apic"; "metadata_block_picture"; "coverart"]
     ~p:(conf_recode#plug "exclude")
-    "Exclude these metadata from automatic recording."
+    "Exclude these metadata from automatic recoding."
 
 let resolve_metadata ~initial_metadata ~excluded name =
   let decoders = get_decoders conf_metadata_decoders mresolvers in
--- a/src/libs/icecast.liq
+++ b/src/libs/icecast.liq
@@ -3,13 +3,13 @@
 # @category Source / Output
 # @argsof output.icecast[!method,!mount,!description,!protocol]
 # @param ~icy_reset Reset shoutcast source buffer upon connecting (necessary for NSV).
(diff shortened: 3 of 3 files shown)


smimram:
The only thing I am not sure of is how to properly raise the exception instead of the `failwith` which gives a badly formatted message.

toots:
Looking great!

For the error I would raise a `Lang.Invalid_value`. It takes the name value (for position) so it's a bit annoying to carry but shouldn't be too hard.

In fact, I would pick the winner earlier on during argument parameters parsing and pass the corresponding value along with a polymorphic variant `[`Id of int | `Name of string]` so you can raise an `Invalid_value` on any error during open regardless of wether id or name was used.

smimram:
I now raise `Error.Invalid_value`. This cannot be done during the parsing of arguments because we need portaudio to be initialized when computing the id from the name...

toots (approved):
Looking good thank you!

PR #4814, opened 2026-01-03
url.decode() the path served by harbor.http.static

martinkirch:
I'm not sure if that should be handled here or earlier in the HTTP stack, but at least that fixes serving files that contain url-encoded characters.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -50,7 +50,7 @@ body:
       render: text
       placeholder: |
         Liquidsoap 2.2.5
-        Copyright (c) 2003-2024 Savonet team
+        Copyright (c) 2003-2026 Savonet team
         Liquidsoap is open-source software, released under GNU General Public License.
         See <http://liquidsoap.info> for more information.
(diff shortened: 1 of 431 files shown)


toots:
Thanks for this! Could you elaborate a little bit more how this comes into play? Maybe a simple file fetching example so I can understand where it should be applied?

martinkirch:
I discovered this by serving files generated with something like `time.string("...%Y-%m-%d/%Hh%M_%S_%z...")` - the %z turns into +001 whose + gets url-encoded. 

example script would be (adapt your own paths)

```
entree = input.alsa()

output.file(%wav,
    { time.string("/home/martin/pigeama/pige/%Y-%m-%d/%Hh%M_%S_%z.wav") },
    entree,
    reopen_when = { 0m },
    fallible=true
)

harbor.http.static(
    port=8080,
    browse=true,
    "/home/martin/pigeama/pige"
)
```

martinkirch:
(that would probably happen to spaces, too)

toots:
I see where you got that `string.sub` from haha. Let me fix that first.

toots (approved):
This looks good thanks!

PR #4815, opened 2026-01-03
fix sub-folder browsing in harbor.http.static

martinkirch:
When browsing sub-folders via `harbor.http.static(browse=true)`, `request.path` does not have a trailing slash so we must add it.


toots:
Thanks!

PR #4835, opened 2026-01-16
Fix source dynamic leak.

toots:
I tried to do something clever for `source.dynamic` that ended up being a bad idea.

The whole problem started because `source.dynamic` has a ill-balanced activation API:
* Some sources can be passed already activated
* Some sources need to be activated prior to be passed to `source.dynamic` to make sure that they ready immediately. for instance `single`.

Because of that, we've been dealing with how to properly take down these sources.

In `2.4.1`, we introduced a separate token given at activation that can be used to deactivate the source.

This is a good idea because, if the token is not used anymore, for instance if the source and token as discarded, we can detect that and deactivate the source.

However, with `source.dynamic`, again, this becomes tricky:
* We need to hold the token while the source is in use in case it gets passed to the `source.dynamic`. Otherwise, the token may get collected and the source would be taken down and would not be ready.
* We want to automatically take down the source if it ends up not being passed to the `source.dynamic`.

I tried an implementation using a `WeakMap` which was a mistake: this ended up creating a dependency loop that prevented any source cleanup at all..

Instead, this PR does what should have been done from the beginning: explicitely pass a cleanup function and ask the user to deal with it...

### DIFF
--- a/src/core/base/operators/dyn_op.ml
+++ b/src/core/base/operators/dyn_op.ml
@@ -213,8 +213,9 @@ let _ =
                        "Source activation" );
                    ]) );
           descr =
-            "Prepare a source that will be returned later. Returns a cleanup \
-             function to call when the source is not needed anymore.";
+            "Prepare a source that will be returned later. Returned source has \
+             new activation attached to it and should be used to submit to the \
+             source later.";
           value =
             (fun s ->
               Lang.val_fun


src/core/base/operators/dyn_op.ml
```diff
@@ -223,16 +183,23 @@ let _ =
...
             (fun s ->
               Lang.val_fun
                 [("", "x", None)]
                 (fun p ->
                   let child = List.assoc "x" p |> Lang.to_source in
-                  s#prepare child;
-                  Lang.unit));
+                  let a = s#prepare child in
+                  Lang.val_fun [] (fun _ ->
+                      s#sleep a;
```
toots:
```suggestion
                      child#sleep a;
```

PR #4857, opened 2026-01-25
fix: prevent frame duplication during manual skip in cross operator

jpmarques19:
## Summary

When `abort_track` is called during an active transition (Before or After state), the previous implementation kept the old transition state while simultaneously starting to buffer for a new transition. This caused audio frame duplication and stuttering, particularly noticeable on resource-constrained devices like Raspberry Pi 4.

## Problem

On manual track skip with crossfade enabled:
- Stutter/glitch occurs at a consistent point during the crossfade
- Issue happens even without ALSA underruns
- Root cause: frame duplication in the audio data during transition creation

## Fix

The fix properly cleans up the current transition state by:
1. Calling `set_status \`Idle` to properly tear down the old transition
2. Calling `reset_analysis` to clear the old buffers  
3. Then proceeding with `abort_track` and `prepare_before`

This ensures that when a manual skip triggers a new crossfade, the old transition is fully cleaned up before the new one begins.

## Test Plan

- [x] Compiles successfully on x86_64
- [ ] Test on ARM64 (Raspberry Pi 4) with manual skip during crossfade
- [ ] Verify no audio stutter during skip
- [ ] Verify natural track transitions still work correctly

### DIFF
--- a/src/core/base/operators/cross.ml
+++ b/src/core/base/operators/cross.ml
@@ -590,9 +590,13 @@ class cross val_source ~end_duration_getter ~override_end_duration
     method abort_track =
       match status with
         | `Idle -> self#source#abort_track
-        | `Before s | `After s ->
+        | `Before _ | `After _ ->
+            (* Clean up the current transition before starting a new one.
+               This prevents frame duplication when abort_track is called
+               during an active transition (e.g., manual skip). *)
+            self#set_status `Idle;
+            self#reset_analysis;
             self#source#abort_track;
-            status <- `After s;
             ignore self#prepare_before
   end
 


toots:
Hi @jpmarques19 ! Thanks for this. Have you tried with the lastest `main`? I recently fixed a clock issue in `crossfade` that was also causing hard to track frames duplications..

jpmarques19:
Hi, I was trying to patch this locally and submitted the PR by mistake. Sorry about that! 
I will try with the latest main. Thank you so much for the help @toots

jpmarques19:
fixed! =D

toots:
I'm so glad to hear. I pulled my hair for a long time on this one!

We just did two releases in a short while so I'm gonna try to wait a little bit but I'll get this out in due time.

PR #4877, opened 2026-02-05
FFmpeg copy and raw content improvement.

toots:
Move ffmpeg copy and raw content params outside of the content list to be acessible even when there is no content.

This is useful when trying to reason about the produced content even when no frame or packets have been produced.

The trick is to preserve stream bondaries inside the data, which is why we're using a chunked structure similar to how `Content_base` works.

Up until now, raw and copy (encoded) ffmpeg content was represented as an array of:

```ocaml
{ length: <...>, data: [(<pos>, { stream_idx, time_base, <packet or frame>})] }
```

This format repeated `stream_idx` and `time_base` on each packet or frame even if there were from the same stream but supported an array containing adjacent streams.

This has worked well for non-sparse streams where the stream always starts with a first packet or frame. However, with subtitle streams, we are dealing with streams where there might be a very long initial content with no packet or frame. Think about a movie where the first subtitle comes after 20 min.

In this case, we really need to be able to handle chunks of content with no data. So in this PR, we switch the content structure to:
```ocaml
[
  { length: <..>, stream_idx: 123, time_base, data: [<content>]}, 
  { length: <..>, stream_idx: 234, time_base, data: [<content>]}
]
```
This way, we are able to account for chunks of content for e.g. stream `123` where no packets or frames are present, i.e. `data = []`.

I have known about this being a better approach for a while but this is a fairly large change due to the requirement to tackle all impacted code so I got helped from the robots on that part..

### DIFF
--- a/src/core/optionals/ffmpeg/ffmpeg_copy_content.ml
+++ b/src/core/optionals/ffmpeg/ffmpeg_copy_content.ml
@@ -22,7 +22,6 @@
 
 open Avutil
 open Avcodec
-include Content_video.Base
 
 let conf_ffmpeg_copy =
   Dtools.Conf.void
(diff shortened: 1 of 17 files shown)


src/core/optionals/ffmpeg/ffmpeg_copy_content.ml
```diff
@@ -43,31 +42,26 @@ type packet =
...
   [ `Audio of audio Avcodec.params
   | `Video of video_params
   | `Subtitle of subtitle_params ]
 
-type packet_payload = {
-  stream_idx : Int64.t;
-  time_base : Avutil.rational;
-  packet : packet;
-}
+(* params is codec_params option, content uses it directly *)
```
toots:
```suggestion
```

PR #4910, opened 2026-02-21
Add .nfo sidecar metadata resolver

oneingan:
## What
Adds optional support for XML `.nfo` sidecar files as a metadata source.

- New stdlib helpers in `src/libs/nfo.liq`:
  - `file.nfo.metadata(nfo_file)` parses a Kodi/Emby/Jellyfin-style XML `.nfo` and returns flattened `nfo.*` metadata.
  - `enable_nfo_metadata(...)` registers a `decoder.metadata.add` resolver that, for a media file `Foo.ext`, looks for `Foo.nfo` and merges the extracted keys.

## Mapping
- The root element name is exported as `nfo.root`.
- Direct textual children of the root element are exported as `nfo.<tag>=<value>`.
- Tag names and `uniqueid` types are normalized (lowercased, namespace prefixes ignored).
- `<uniqueid type=\"X\">VALUE</uniqueid>` is exported as `nfo.uniqueid.X=VALUE`.
- If `<uniqueid ... default=\"true\">VALUE</uniqueid>`, the ID is also exported as:
  - `nfo.uniqueid.default=VALUE`
  - `nfo.uniqueid.default.type=X`
- Typeless `<uniqueid>` or `type=\"default\"` (scraper quirk) is exported as `nfo.uniqueid.untyped=VALUE`.
- Only `nfo.*` keys are emitted (no standard `title`/`year`/`genre` mapping) to avoid collisions.

## Extensibility
- `file.nfo.rules(~extra_rules=[...])` builds a reusable rule table (prepended to built-ins).
- `file.nfo.metadata(~rules=..., ...)` accepts a prebuilt rule table (defaults to `file.nfo.rules()`).
- `enable_nfo_metadata(~extra_rules=[...], ...)` precomputes rules once and reuses them for each track.
- `enable_nfo_metadata(~rules=..., ...)` accepts a prebuilt rule table (overrides `extra_rules`).

## Tests
- `tests/streams/nfo_sidecar_metadata.liq` covers both `file.nfo.metadata` and resolver integration via `file.metadata` (including namespaces/casing, default uniqueid selection, `untyped` fallback, and custom rules).

## Notes / Limitations
- Conservative first step: it only extracts direct textual children (nested structures like actors are ignored for now).

### DIFF
--- a/src/libs/nfo.liq
+++ b/src/libs/nfo.liq
@@ -4,6 +4,7 @@
 # metadata resolver through `enable_nfo_metadata(...)`.
 
 # @category File
+# @flag hidden
 let file.nfo = ()
 
 # @flag hidden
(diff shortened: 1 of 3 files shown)


toots:
This is looking good! I merged a PR to make the tests requiring private assets optional on forks, feel free to rebase/merge..

src/libs/nfo.liq
```diff
@@ -0,0 +1,272 @@
+# Parse XML `.nfo` sidecar files (Kodi/Emby/Jellyfin style).
+#
+# This is meant to be used either directly via `file.nfo.metadata(...)` or as a
+# metadata resolver through `enable_nfo_metadata(...)`.
+
+# @category File
```
toots:
```suggestion
# @flag hidden
```

src/libs/nfo.liq
```diff
@@ -0,0 +1,272 @@
...
+    {without_type()},
+    fun (t) ->
+      # Some scrapers incorrectly emit `type="default"`. Kodi warns against it,
+      # so we treat it as untyped.
+      (t == "" or t == "default") ? without_type() : with_type(t)
+  )
+end
+
+# @flag hidden
+let file.nfo._tag_rules =
```
toots: I would add this one and `_uniqueid_rule` inside the function below's scope to avoid adding to the top-level (public) scope.

src/libs/nfo.liq
```diff
@@ -0,0 +1,272 @@
...
+# Special case:
+# - `<uniqueid type="X">VALUE</uniqueid>` is exported as `nfo.uniqueid.X=VALUE`.
+# - If `default="true"`, the ID is also exported as `nfo.uniqueid.default=VALUE` and
+#   the type as `nfo.uniqueid.default.type=X`.
+#   If multiple IDs are marked default, values are joined with `;`.
+# - Typeless `<uniqueid>` (or `type="default"`, a scraper quirk) is exported as
+#   `nfo.uniqueid.untyped=VALUE`.
+# - Repeated tags are joined with `;` (in order of appearance).
+# - Tag names and unique ID types are lowercased; namespace prefixes are ignored.
+# - Deprecated Kodi `<id>` is exported as `nfo.id`.
```
toots: I see that this is all tested except `<id>` right?

src/libs/nfo.liq
```diff
@@ -0,0 +1,272 @@
...
+  # Shape: [(tag_name, (node, text) -> [(key, value)])].
+  [("uniqueid", file.nfo._uniqueid_rule)]
+
+# Build a tag rule table for `.nfo` parsing.
+#
+# Returned rule tables can be reused across multiple calls to `file.nfo.metadata`.
+#
+# @category File
+# @param ~extra_rules Additional tag rules (prepended to built-ins).
+def file.nfo.rules(~extra_rules=[]) =
```
toots: Could `extra_rules` override the `"uniqueid"` rule if the user wanted to?

toots (commented):
This looks great. I like the explicit `rules`, this gives the user some additional flexibility.

Only minor code reorganization comments. This is definitely good to go otherwise. We'll need to add a `CHANGES.md` entry for it.

On the same topic, do you know how you'd plan to use it? This could be worth sharing or, if you feel like it, adding a `doc/content/nfo.md` entry so more people can know it's available and how to use it. (I'm also curious what you plan on doing with it!)

oneingan:
(Edited: markdown got mangled by my shell quoting; see the next comment for the proper summary.)

oneingan:
Addressed the inline review comments:
- _accum_add now uses list.fold
- inlined single-use helpers and kept the uniqueid handler local to file.nfo.rules
- extended tests/streams/nfo_sidecar_metadata.liq to cover overriding the uniqueid rule
- added a CHANGES.md entry

All review threads should be resolved now.

oneingan:
My use case: I generate an XMLTV EPG externally, read a programme uniqueid from it, and then choose the file to play by matching that uniqueid against the NFO sidecar metadata (e.g. nfo.uniqueid.* / nfo.uniqueid.default).

Once I have the full workflow wired up I can share a minimal example (and/or add a small doc page) showing: enabling the resolver, reading the XMLTV id, and selecting the next media file by matching IDs.

oneingan:
@toots re: the `~priority` docstring / your inline note about avoiding overriding media tags: right now the resolver only emits `nfo.*` keys, so it won’t override embedded tags like `title`/`artist` unless a user-provided rule emits non-namespaced keys.

Would you like to keep it strictly namespaced (and we just clarify that in the docstring / skip a priority-collision test), or do you envision an opt-in mapping helper that copies selected `nfo.*` fields to standard metadata keys (where low priority + a test would make sense)?

PR #4919, opened 2026-02-24
Reduce default buflen.

toots:
Default buffer length was set to the architecture's page size in https://github.com/savonet/liquidsoap/pull/915 in an attempt to blindly optimize memory allocation.

The assumption was that memory, in the non-OCaml heap, is allocated faster if it asks for a whole page at once.

However, this is not how OCaml memory allocation works. Memory on the OCaml side is allocated using a promoted minor head. When requesting memory, the minor heap is used until it runs out of space which then forces it to promote its unused values to major heap and so on.

Thus, requesting large amount of memory on the OCaml side may result in larger minor stacks and potentially more major heap promotions, which would impact overall performances.

This PR thus reduces this number.

### TODO

- [x] Can this be measured?
- [x] Can this be configurable?

### DIFF
--- a/src/core/base/tools/utils.ml
+++ b/src/core/base/tools/utils.ml
@@ -102,7 +102,7 @@ end
 (* Force locale *)
 external force_locale : string -> unit = "liquidsoap_set_locale"
 
-let buflen = 1024
+let buflen = try int_of_string (Sys.getenv "LIQ_BUFLEN") with _ -> 1024
 let () = force_locale "C"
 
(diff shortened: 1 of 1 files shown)


toots:
I ran a quick benchmark with https://github.com/toots/test-alloc-size and OCaml `4.14` and `1024` seems like the sweet spot:

results

toots:
Similar sweet spot with OCaml `5.4`:
results

Issue #4928, opened 2026-02-26
list.assoc.nullable causes incorrect type inference for list values


toots:
## Description

When using `list.assoc.nullable` on a list, the type inference incorrectly infers the list values as nullable, which then causes type errors when trying to access fields on those values directly.

## Expected type signature

```
('a, ['a * 'b]) -> 'b?
```

The function should return a nullable `'b`, but the input list should have non-nullable values `'b`.

## Actual behavior

The type inference seems to propagate the nullability back to the list type, making `'b` nullable in the list itself.

## Workaround

Use `list.assoc.mem` + `list.assoc` instead of `list.assoc.nullable`.

PR #4931, opened 2026-02-28
Change inner type for `null` to be non-nullable to allow its use to protect a type from being unified as nullable.

toots:
Fixes: #4928

### DIFF
--- a/src/lang/base/builtins_null.ml
+++ b/src/lang/base/builtins_null.ml
@@ -24,7 +24,7 @@ let null =
   let a = Lang.univ_t () in
   Lang.add_builtin "_null" ~category:`Programming
     ~descr:"Create a nullable value."
-    [("", Lang.nullable_t a, Some Lang.null, Some "Value to make nullable.")]
+    [("", a, Some Lang.null, Some "Value to make nullable.")]
     (Lang.nullable_t a)
     (fun p ->
       match Lang.to_option (List.assoc "" p) with
--- a/src/libs/list.liq
+++ b/src/libs/list.liq
@@ -218,7 +218,7 @@ end
 # @category List
 def list.assoc.nullable(key, l) =
   try
-    list.assoc(key, l)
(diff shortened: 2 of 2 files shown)


toots:
The only liquidsoap script API difference from this change is on the `nfo` parsing functions:

e.g. `file.nfo.metadata` becomes:

```
Extract metadata from an `.nfo` file.

Returned metadata keys are namespaced under `nfo.*`.

Currently, only direct children of the root element that contain textual content are extracted.

Special case:
- `<uniqueid type="X">VALUE</uniqueid>` is exported as `nfo.uniqueid.X=VALUE`.
- If `default="true"`, the ID is also exported as `nfo.uniqueid.default=VALUE` and   the type as `nfo.uniqueid.default.type=X`.   If multiple IDs are marked default, values are joined with `;`.
- Typeless `<uniqueid>` (or `type="default"`, a scraper quirk) is exported as   `nfo.uniqueid.untyped=VALUE`.
- Repeated tags are joined with `;` (in order of appearance).
- Tag names and unique ID types are lowercased; namespace prefixes are ignored.


Type: (?rules : [string *
           ((
             {
               xml_children : [string * {xml_text : string}],
               xml_params : [string * string?]
             }, string) -> [string * string])],
 string) -> [string * string]

Category: File

Arguments:

 * rules : [string *
 ((
   {
     xml_children : [string * {xml_text : string}],
     xml_params : [string * string] # <- HERE was xml_params : [string * string?] before.
   }, string) -> [string * string])] (default: [("uniqueid", <fun>)])
     Tag rules. Use `file.nfo.rules(...)` to build them.

 * (unlabeled) : string
     Path to the `.nfo` file.
```

This seems fine to me, this is saying the user should submit strings for `xml_params`. And I think that it parses as `string` anyways, it was typed as `nullable` probably because of the use of `null(param)` somewhere.

cc/ @oneingan

PR #5029, opened 2026-03-23
clock: fix sub_clock accumulation causing gradual CPU growth

toots:
## Problem

Sub-clocks created by `Child_support.base` operators (e.g. `noblank`, `pipe`, `resample`, ffmpeg filters) were registered in the output clock's `sub_clocks` queue at construction time and never removed. When such operators are used inside dynamic contexts (crossfade transitions, dynamic sources), each instantiation adds an entry to `output_clock.sub_clocks` that persists after the operator sleeps.

Since `_tick` iterates `Queue.elements clock.sub_clocks` and recursively ticks each entry on every frame, this causes O(N) extra work per tick after N instantiations — gradual CPU growth over time.

## Changes

- `clock.ml` / `clock.mli`: replace `create_sub_clock` with explicit `register_sub_clock` and `deregister_sub_clock`. `register_sub_clock` validates that the clock is passive and not already owned by another clock.
- `child_support.ml`: create the sub-clock without registering it at construction. Register in `on_wake_up` and deregister in `on_sleep`, tying the sub-clock's presence in the parent queue to the operator's active period.
- `builtins_ffmpeg_filters.ml`: register `input_clock` when the first graph output wakes up, deregister when the last one sleeps.

### DIFF
--- a/src/core/base/clock.ml
+++ b/src/core/base/clock.ml
@@ -796,24 +796,6 @@ let set_stack c stack =
 
 let register_sub_clock parent sub =
   let _sub = Unifier.deref sub in
-  (match Atomic.get _sub.state with
-    | `Stopped _ -> Atomic.set _sub.state (`Stopped `Passive)
-    | _ when _sync _sub = `Passive -> ()
-    | _ ->
-        failwith
-          (Printf.sprintf
-             "register_sub_clock: clock %s must be passive but has sync mode %s"
-             (_id _sub)
-             (string_of_sync_mode (_sync _sub))));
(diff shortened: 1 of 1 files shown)


charles-rc:
Ohhhh boy @toots! Are you about to make my day?

toots:
@charles-rc it's a bit of a blind fix but it's got decent chances. I found the smoking gun in your logs let me update your issue one sec.

charles-rc:
@toots thanks for the work into this! We'll let you know how the testing goes once there's a release we can consume.

Issue #5059, opened 2026-03-27
EBU R128 TP measurement and conformity safeguard


MikaSappi:
### Description

Setting track amplification by only the measured LUFS leads to a scenario where a low-level, unlimited audio would not be within the R128 specification. For example, a track could have integrated LUFS of -23.5, and a true peak at -0.4 dB. Amplifying purely based on LUFS alone leads to a true peak of 0.1 dB, which is more than what R128 allows and leads to distortion.

The standard consists of two values we need to keep an eye on: LUFS-I and TP. LUFS has a tolerance of ± 1 dB, while TP is a hard limit at -1 dB. This enhancement trades at most 0.5 LUFS for TP conformity.

I think checking the TP would make sense in this scenario. If the true peak is less than half of the tolerance (meaning 0.5 dB) over the limit, we can amplify the audio by `lufs_track_gain + [target-tp - real-tp]` to make sure the track is within the LUFS tolerance, but also that it doesn't have a true peak over -1 dB.

Another case would be tracks that, even after corrected attenuation, would still be too high in true peak. These tracks would benefit from limiting. An architectural questions, which I'm not capable of answering, is where this limiting should happen. The limiter in this case is meant to be the absolute last safeguard before any other processing. It's also unknown at this moment whether the built-in limiter works on sample peaks or true peaks.

I don't know how much resources the limiter in Liquidsoap consumes, but if negligible, I would suggest the following:
```
def normalize_track_gain(~id=null, s) =
  s = amplify(id=id, override=settings.normalize_track_gain_metadata(), 1., s)
  limit(id=id, ratio=100., threshold=-1., attack=5., release=50., s)
end
```

There's an upcoming PR coming for the gain correction (which handles smaller overshoots). The limiter idea would require a bit of thinking.

PR #5060, opened 2026-03-27
Measure true peak and modify track amplification based on the measured TP

MikaSappi:
This PR adds an argument for the `enable_lufs_track_gain_metadata()` operator to allow true peak computation. It also adds new settings entry `settings.lufs.true_peak_max` to set the TP target. 

To use TP computation, the user must use `enable_lufs_track_gain_metadata(true_peak=true)`. Default is false.

It follows the same flow how LUFS is measured and used.

The calculation is the minimum compliant implementation as set by the [ITU](https://www.itu.int/dms_pubrec/itu-r/rec/bs/r-rec-bs.1770-2-201103-s!!pdf-e.pdf). The oversampling is fixed at 4x at the time being, which is conservative for high sample rates. The specification allows 2x oversampling for 96 kHz.

The new function logs the following line to show what it's doing. In this example it tells that the correction value has been capped at 0.5 dB to stay within tolerance, but to do as much as possible to get the true peak closer to the -1 dB threshold. Values that fall in case 3, would need a limiter, which hasn't been implemented at the time being.
```
2026/03/27 17:20:59 [lufs:3] "/home/liq-user/tp-test-files/file.m4a": true peak after normalization would be -0.237829942188 dBTP (0.762170057812 dB over limit). Gain correction capped at 0.5 dB to preserve LUFS-I; a limiter is required for this track.
```
The user may also log the values:
```
def on_meta(meta) =
  gain = meta["liq_normalize_track_gain"]
  tp   = meta["lufs_true_peak"]
  log("gain : #{gain}")
  log("tp   : #{tp}")
end

s.on_metadata(synchronous=false, on_meta)
```
For users using `enable_lufs_track_gain_metadata()` and `normalize_track_gain(s)`, this update doesn't do anything unless they opt to use the TP computation.

This presents a partial resolution for #5059.


src/libs/lufs.liq
```diff
@@ -117,27 +161,118 @@ def replaces file.lufs(~id=null, ~compute=true, ~ratio=null, file_name) =
...
+# When `compute=false`, the existing LUFS metadata is used without any
+# computation (same behaviour as before).
+#
+# @param ~compute Compute LUFS if metadata tag is empty.
+# @param ~true_peak Also measure true peak and apply EBU R128 TP correction to the gain.
+# @param ~ratio Decoding ratio. A value of `50.` means try to decode the file `50x` faster than real time, if possible. Defaults to `settings.lufs.decoding_ratio` when `null`.
 # @category Liquidsoap
-def enable_lufs_track_gain_metadata(~compute=true, ~ratio=null) =
-  ratio = ratio ?? settings.lufs.decoding_ratio()
+def enable_lufs_track_gain_metadata(~compute=true, ~true_peak=false, ~ratio=null) =
```
toots: Should `true_peak` be `true` by default? Do you expect most user to assume this protection is on by default?

src/core/base/operators/lufs_c.c
```diff
@@ -104,3 +104,103 @@ CAMLprim value liquidsoap_lufs_process(value _stage1, value _stage2, value _x) {
...
+  TP_val(ans) = tp;
+  CAMLreturn(ans);
+}
+
+/* Unlike liquidsoap_lufs_process (the original IIR function), we validate that
+   the OCaml buffer has at least as many channel sub-arrays as the state expects.
+   The original code relies on the OCaml wrapper's assert for this guarantee;
+   we add a C-side guard as well because an out-of-bounds Field() access on a
+   format change would be a silent memory-safety violation, not a catchable
+   OCaml exception. */
```
toots: I think this comment can be removed, this check is not out of the ordinary and pretty straight forward.

toots (commented):
This looks great! Only minor comments..

src/libs/lufs.liq
MikaSappi: After I tested the CPU cost of this operation, it should be `true` by default.

MikaSappi:
The push also includes a bug fix in the TP calculation, before it only calculated the inter-sample peaks while ignoring the original sample peaks, now it actually calculates the max between these two types of peaks.

Measurement agrees with me:
```
2026/03/30 13:45:29 [lang:3] RMS: 0.707106781187
2026/03/30 13:45:29 [lang:3] LUFS integrated: -0.69
2026/03/30 13:45:29 [lang:3] LUFS momentary:  -0.682439549862
2026/03/30 13:45:29 [lang:3] True peak:       0.0
```
Since $RMS = 1 / \sqrt 2$, for TP of 0.0, dB = $20log_{10}(1) \rightarrow 0$, RMS from 1.0 is 0.707... is correct.

toots (approved):
This looks great! I think one code comment should be removed. Then we should rebase the branch against `main` and let this auto-merge nicely!

toots:
The CI won't share secrets so docker won't pass the rest looks good. Let me see if I can make the change to make this pass.

PR #5077, opened 2026-04-04
switch: per-source composition as a handoff between leaving and entering sources

toots:
## Why this matters

Source switching in Liquidsoap has always been powerful but hard to get right. The old API (`transitions`, `transition_length`, `override`, `track_sensitive`, `replay_metadata`) was a flat list of parallel parameters that every source in a switch had to share. You could not say "the live input may interrupt a song, but the jingle playlist may not" — the setting applied to the whole operator.

This PR replaces that model with **per-source composition methods**. Each source says how it wants to be handed over to, and the switch just follows. The mental model is a handoff between a **leaving** source and an **entering** source: every decision belongs to one of the two sources actually involved, never to the switch as a whole.

The rest of this description is what that looks like in practice. The new [source composition](doc/content/composition.md) page covers the same ground in more detail.

---

### A live show interrupting the music

```liquidsoap
live  = input.harbor("live")
music = playlist("~/music")

radio = fallback([live, music])
```

- Nobody connected: `live` is unavailable, `music` plays.
- The DJ connects mid-song: a live input does not wait for track boundaries, so it cuts in right away and the song is faded out underneath it.
- The DJ disconnects: `live` is gone, so there is nothing to fade and the music returns immediately. Because its song had been interrupted, the playlist starts a **fresh track** rather than resuming half-way through.

No parameters. `input.harbor` is live, `playlist` is file, and each behaves accordingly. Before this PR you needed `track_sensitive=false` plus a `transitions` list to get the same result.

### Jingles between songs

```liquidsoap
radio = rotate([music.{weight = 3}, jingles.{weight = 1}])
```

Both are file sources, so neither will interrupt the other: when the jingle's turn comes up the switch waits for the current song to end, then starts the jingle cleanly with no fade — nothing was interrupted, so there is nothing to fade.

### An announcement that cannot wait

Sometimes a file source *should* interrupt:

```liquidsoap
announcements = request.queue()

radio = fallback([announcements.{track_sensitive = false}, music])
```

As soon as something lands in the queue the song is faded out and the announcement plays.

This is the general rule:

> A switch cuts into a song that is still playing only if **at least one** of the two sources involved has `track_sensitive = false`. Otherwise it waits for the end of the track.

Both sides get a say, so **adding a live input to a `fallback` never changes how two playlists hand over to each other**.

### A relay that carries a playlist

`input.http` is live by default and cuts in immediately. If the stream being relayed is itself a playlist of songs, you would rather wait for a boundary:

```liquidsoap
relay = input.http("https://relay.example.com/stream")
relay.composition_type := "file"
```

### Other per-source controls

```liquidsoap
# Never pick this source for two consecutive tracks.
radio = rotate([music, jingles.{single = true}])

# This source announces its own metadata; don't replay anything.
radio = fallback([s1.{replay_metadata = false}, s2])
```

### Writing your own transition

Replacing the old `transitions=[...]` list — the transition belongs to the source being entered:

```liquidsoap
def my_transition({ending, starting, replay_metadata = _}) =
  if null.defined(ending) then
    old = max_duration(3., null.get(ending))
    (add([fade.out(duration=3., old), fade.in(duration=3., starting)]) : source)
  else
    starting
  end
end

# Only s1 gets the custom transition; s2 uses the default.
radio = fallback([s1.{on_select = my_transition}, s2])
```

`ending` is the source being left, `null` when nothing was interrupted. The `max_duration` matters: whatever you return keeps pulling from `ending` until you stop it, and the leaving source is only cleaned up once nothing pulls from it any more.

---

## The methods

Every source exposes these in its **Composition methods** section (`liquidsoap -h <operator>`):

| Method | Purpose |
|---|---|
| `composition_type` | `"file"` or `"live"` — which set of defaults this source gets |
| `track_sensitive` | Whether this source insists on track boundaries |
| `replay_metadata` | Replay latest metadata when selected (default `true`) |
| `single` | Forbid selection for two consecutive tracks |
| `on_select` | Method of the **entering** source: builds the transition |
| `on_leave` | Method of the **leaving** source: cleanup once released |

`on_leave` receives `{source, track_sensitive}`, where `track_sensitive` reports whether the source finished its track naturally (`true`) or was interrupted (`false`).

Defaults live in two global profiles, `source.composition.file` and `source.composition.live`, each holding `on_select`, `on_leave`, `track_sensitive` and `replay_metadata`. `source.composition.default_on_select` is the default transition and `source.composition.legacy_on_select` restores the previous no-fade behaviour. `settings.source.composition.max_fade` controls the default fade length.

### What is removed

`transitions`, `transition_length`, `override`, `track_sensitive` and `replay_metadata` are removed **as switch parameters**, with no backwards-compatibility shim. `track_sensitive`, `replay_metadata` and `single` live on as per-source methods. See `doc/content/migrating.md` for the mapping.

---

## Implementation notes

**Composition tag on every source** (`source.ml`) — each source carries a `` [ `File | `Live | `Passthrough ] `` tag. `Passthrough` delegates to `effective_source`, so transform operators inherit from what they wrap; `resolved_composition` follows the chain with a self-loop guard. `Lang.add_operator` infers the tag from `~category`: active inputs are live, passive inputs and synthesis sources are file, everything else passes through, and outputs get no composition methods.

**Selection is decided per handoff** (`switch.ml`) — whether we are at a track boundary is derived from the source being left: it could not fill the frame past the requested position, or it is no longer ready. That alone decides whether `on_select` receives an `ending` to transition out of. Mid-track takeovers are gated on the two sources involved rather than on all children, which is what keeps one live source from making the whole switch non-track-sensitive. The switch-wide flag remains only for deciding whether the selection must be re-evaluated mid-track at all.

**Metadata replay no longer clobbers the source's own metadata** — `insert_metadata` gains an `override` parameter (default `true`, preserving existing behaviour). The composition profiles replay with `override=false`, so a source starting a fresh track keeps its own metadata and the replay only fills in what is missing.

**Documentation** — new `doc/content/composition.md` linked from the index; `migrating.md` corrected (`replay_metadata` moved rather than removed, and the `ending` / `track_sensitive` semantics stated precisely); operator help no longer reuses the title `Composition:` for two different sections.

## Tests

`tests/regression/` covers:
- `switch_per_source_on_select`: custom per-source `on_select` callbacks
- `switch_per_source_on_leave`: `on_leave` firing on preemption and natural track end
- `switch_per_source_single`: `single = true` prevents back-to-back selection
- `switch_per_source_replay_metadata`: metadata replay on selection
- `switch_per_source_no_reselect_during_fade`: no re-selection while a transition is in progress
- `switch_autodj_jingle_metadata`: a delayed jingle source reports its own metadata on re-selection, not the previous jingle's
- `GH3239`: regression for a related switch ordering bug

Fixes #5074


doc/content/migrating.md
```diff
@@ -81,6 +81,71 @@ The `add` operator now relays metadata from all sources being summed (see above)
...
+```liquidsoap
+def my_on_select({ending, starting}) =
+  if null.defined(ending) then
+    let old = null.get(ending)
+    (sequence([(fade.out(duration=3., old) : source), starting]) : source)
+  else
+    starting
+  end
+end
+s = fallback([s1.{on_select = my_on_select}, s2])
```
martinkirch: this example does not passes type-checking:
```
At test.liq, line 23, char 2-7:
  music.{

Error 5: this value has type
  _
  .{
    on_select : ('A
                 .{
                   ending : source(audio='C, 'B)?,
                   starting : source(audio='C, 'B)
                 }, ...) -> _
  }
  where 'C is a track and a track of type: pcm, pcm_s16 or pcm_f32,
    'B is an orderable type and a set of tracks to be muxed into a source
but it should be a subtype of the type of the value at test.liq, line 19 char 20 - line 26 char 1
  _.{on_select : ({ending? : source('A), starting : source('A)}, ...) -> _} (inferred at test.liq, line 20, char 2-8)
```

however it works when writing

```
def on_stream_select(state) =
  if null.defined(state?.ending) then
    ...
  else
    fade.in(state.starting)
  end
end
```

martinkirch (changes requested):
we're getting close !

for a start, I migrated the test with a track-insensitive fallback that crossfades between a local playlist and an http stream. It fades, but it does not skip

````
server.harbor() # go to http://localhost:8000/telnet , type stream.start() / .stop()
stream = buffer(input.http(id="relay", "https://live.campusgrenoble.org/rcg112", start=false))
music = playlist("/home/martin/Musique")

def on_stream_select(state) =
  if null.defined(state?.ending) then
    let old = null.get(state?.ending)
    (add([
      (fade.out(duration=2., old) ), 
      (fade.in(duration=2., state.starting) ), 
      ]) 
    )
  else
    fade.in(state.starting)
  end
end

programs = fallback([
  stream.{
    track_sensitive = false,
    on_select = on_stream_select,
  },
  music.{
    track_sensitive = false,
    on_leave = fun(_) -> begin log("music.leave"); music.skip(); end
  }
])

output(programs)
````

no `music.leave` in logs and after the stream the track resumes to where it stops.

**Comment by toots** [comment]
Hi @martinkirch ! The API is not yet fully tested nor implemented.

I just pushed some commits that seem to make the fallback between `playlist` and i`nput.http` work.

This looks like a big chunk of work but I think it's exciting because this could potentially make switch, fallback etc work the way 99% of the users would expect them too while keeping a lot of flexibility under the hood.

**Comment by martinkirch** [comment]
> Hi @martinkirch ! The API is not yet fully tested nor implemented.

Sorry I'm eager to see that happen, too :sweat_smile: 

> I just pushed some commits that seem to make the fallback between `playlist` and i`nput.http` work.

using the test script above, still no `music.leave` in logs :/ 

it also seem that `input.http` is a kinda scattered, many micro-silences are inserted.

> This looks like a big chunk of work but I think it's exciting because this could potentially make switch, fallback etc work the way 99% of the users would expect them too while keeping a lot of flexibility under the hood.

That's right, defaults are perfect. tag me when you'd like some tests !

**Comment by toots** [comment]
@martinkirch it's getting there!

What's your opinion on leaving live inputs like `input.http`? 

Take this simple switch:
```
fallback([
  (play_live, input.http("...")),
  ({true}, playlist("..."))
 ]
```
Currently, with the new changes, the playlist will "play it nice" and say okay my data is not live I can wait for a nice track mark to switch.

But the live source may or may not have track marks. A LOT of icecast sources don't. 

I'm not sure if there's a sensitive user-facing default.

Users who use a `input.http` with a URL that has them will expect the playlist to wait and user who use a `input.http` with a URL that doesn't have track marks will get tripped and say: "my `play_live` implementation says `false` why did it not switch".

I could of course implement a finer-grained logic live the live source could say "switch off as soon as my predicate is done" but, again, it's not obvious to see if there would be a sensible user default..

**Comment by martinkirch** [comment]
That example is not so simple, because it assumes you also have implemented `play_live`. At that level of implementation I'd advise for using `input.http.start() / .stop` instead.

Another problem in your example is that the `input.http` will overflow its buffer too easily. When leaving because `play_live = false` it'd be nice to have a `.skip()` method that would flush last frames in the buffer.

But are you sure it should be track-sensitive ? All cases I know of http/harbor sources are supposed to play when asked to, at most within a few seconds. Tracks marks are interesting to switch between "file" sources, but as soon as there's a "live" one we should probably default to track-insensitive. That's _the_ great improvement of `_composition` markers as file/live/passthrough : they'd allow to bubble up such default behaviors in a source/operators tree.

**Comment by toots** [comment]
@martinkirch I think that we're ready for testing now!

File: doc/content/migrating.md [hunk 2]
```diff
@@ -97,6 +97,82 @@ The `add` operator now relays metadata from all sources being summed (see above)
...
+s = fallback(transitions=[my_transition], transition_length=3., [s1, s2])
+```
+
+**After:**
+
+```liquidsoap
+def my_on_select({ending, starting, replay_metadata}) =
+  if null.defined(ending) then
+    let old = null.get(ending)
+    (sequence([(fade.out(duration=3., old) : source), starting]) : source)
```
martinkirch, line 147: it seems that `: source` anotations are not even needed \o/

martinkirch (commented):
looking great! I'll need a longer test session to check how transitions sound. I'll try to get back to you before Monday.

martinkirch (changes requested):
Found a bug with this test script: https://gist.github.com/martinkirch/09fbae895fd12fff2bd132421be04329

 - when sending stream.start, it cross-fades as expected 👌 
 - when stopping the stream, it lasts a few more seconds until the buffer is emptied (OK) then fades out... to silence

_but_ it kinda keeps playling: if you leave it running, the log tells it's still playing files. In Radiotomate's main script the "live" view was still working and counting remaining time correctly, however it seemed that it did not skip the left track (or it was repeating previous metadata).

toots:
> Found a bug with this test script: https://gist.github.com/martinkirch/09fbae895fd12fff2bd132421be04329
> 
> * when sending stream.start, it cross-fades as expected 👌
> * when stopping the stream, it lasts a few more seconds until the buffer is emptied (OK) then fades out... to silence
> 
> _but_ it kinda keeps playling: if you leave it running, the log tells it's still playing files. In Radiotomate's main script the "live" view was still working and counting remaining time correctly, however it seemed that it did not skip the left track (or it was repeating previous metadata).

Yes I think I see why this fails. 

It's really tricky to know _when_ the transition is over, to issue `source.skip` and to allow re-selection.

In order to fix this, the new functionalities here require that you properly dispose of the ending source in your transition so this can be detected and acted upon.

The two operators that can properly dispose of the ending source are: `source.dynamic` and `sequence`.

You can see here how it's done in the default transition: https://github.com/savonet/liquidsoap/pull/5077/changes#diff-a3f4eda9df94466811d98b85712824e1803a81ac019b46457ffe592471bb5a19R135

In your case, I would rewrite the transition as:
```liquidsoap
def on_stream_select({ending, starting, replay_metadata}) =
  if null.defined(ending) then
    let old = null.get(ending)
    old = fade.out(
      duration=2.,
      max_duration(2., sequence([old switch([])]))
    )
    
    add([
      old,
      fade.in(duration=2., starting), 
      ])
  else
    fade.in(starting)
  end
end
```

I'll definitely add something in the doc and perhaps some helpers in the library too to that effect.

martinkirch:
seems that in that case we need a type annotation on `old` ; so the test script becomes 

```
server.harbor() # go to http://loccalhost:8000/telnet , type stream.start / .stop
stream = buffer(input.http(id="stream", "https://live.campusgrenoble.org/rcg112", start=false))
music = playlist("/home/martin/Musique")

def on_stream_select({ending, starting, replay_metadata}) =
  if null.defined(ending) then
    let old = (null.get(ending) : source)
    old = fade.out(duration=2.,
      max_duration(2., sequence([old, switch([])]))
    )
    add([old, fade.in(duration=2., starting)])
  else
    fade.in(starting)
  end
end

programs = fallback([
  stream.{
    track_sensitive = false,
    on_select = on_stream_select,
  },
  music.{
    track_sensitive = false,
    on_leave = fun(_) -> begin log("music.leave"); music.skip(); end
  }
])

output(programs)
```

however when ending the stream it crashed: log [in this gist](https://gist.github.com/martinkirch/977cfdd136bd7750ac20c8100e631b20)

oranjebloom:
Apologies if this comes across as a stupid question, but I have been following this thread and want to know if the changes that are being tested here with `fallback` will also be applicable to `switch` when this eventually rolls out? I am hoping that this is the case.

toots:
@oranjebloom Absolutely! What do you think of the current proposal?

@martinkirch Sorry I've been busy with other fires but I do plan on coming back to this very soon!

oranjebloom:
> @oranjebloom Absolutely! What do you think of the current proposal?
> 

@toots from what I can see, it allows for a lot more flexibility and control and I think it is an improvement. This may be premature, but I've been testing time-based scheduling with switch using track_sensitive = false per source and on_select = custom_fade. I'm seeing the same silence issue after a transition completes that @martinkirch describes. My use case is purely file-based playlists with a time-based schedule — no live sources. The on_leave skip is firing correctly but the starting source doesn't play after the fade completes.

oranjebloom:
I've been testing Liquidsoap 2.5.0+git@4b1dff7d9 and I have been having a lot of joy with file-based playlists and a time-based schedule. Thank you for providing the updates! I hope the same can be said for @martinkirch 's tests with live sources.

I have even been able to successfully manage the two scenarios below:
- Tracks that are shorter than my set fade out duration of 10 seconds (such as short jingles)
- Tracks that have less than 10 seconds remaining to play when the scheduled switch requires it to fade out. 

My preference in both scenarios is that the track of the old outgoing source be allowed to play out the remainder of its audio before the next track from the incoming new source begins. 

minimal test script: https://pasteview.com/5YjtsGO

toots:
@oranjebloom I made the `max_fade` a `settings.source.composition.max_fade`. If there is more than `max_fade` during transition, the track ends without a fade-out!

@martinkirch bug should be fixed now thanks for reporting!

martinkirch:
> @martinkirch bug should be fixed now thanks for reporting!

It is, thanks for the fix ! First tests look good. Now I'm trying to skip the leaving track.

Starting with [the script above](https://github.com/savonet/liquidsoap/pull/5077#issuecomment-4290704408) :

````ocaml
music = playlist("/home/martin/Musique")
...
programs = fallback([
...
  music.{
    track_sensitive = false,
    on_leave = fun(_) -> begin log("music.leave"); music.skip(); end
  }
````

That plays, but it seems `on_leave` is not called at all : no log message and the track resumes where it was when the sources is back on air.

So I tried another syntax, inspired by the previous examples of  `on_stream_select`, but that doesn't type check:

````ocaml
def on_music_leave({ending, track_sensitive}) =
  log("music.leave")
  ending.skip()
end
...
    on_leave = on_music_leave,
````

````
At test.liq, line 22 char 20 - line 32 char 1:
programs = fallback([
  stream.{
...
  }
])

Error 5: this value has type
  [_
   .{
     on_leave : (_.{ending : _, source : _, track_sensitive : _}, ...) -> _,
     on_select : _,
     track_sensitive : _
   }]
but it should be a subtype of
  [_
   .{
     on_leave? : ({source : _, track_sensitive : _}, ...) -> _,
     on_select? : _,
     track_sensitive? : _
   }]
````

Surprisingly, writing `def on_music_leave(ending, track_sensitive) =` works, but ignores the function entirely.

sidenote: using a complex on_stream_select seem to create many operators, and each one logs a lot, on every transition they printed:
````
026/05/12 21:29:07 [switch:3] Switch to buffer.producer with transition.
2026/05/12 21:29:07 [source.2:3] Content type is {audio=pcm(stereo)}.
2026/05/12 21:29:07 [audio.add:3] Content type is {audio=pcm(stereo)}.
2026/05/12 21:29:07 [fade_out:3] Content type is {audio=pcm(stereo)}.
2026/05/12 21:29:07 [track_audio_amplify.1:3] Content type is {audio=pcm(stereo)}.
2026/05/12 21:29:07 [source.dynamic.1:3] Content type is {audio=pcm(stereo)}.
2026/05/12 21:29:07 [max_duration:3] Content type is {audio=pcm(stereo)}.
2026/05/12 21:29:07 [sequence:3] Content type is {audio=pcm(stereo)}.
2026/05/12 21:29:07 [blank:3] Content type is {audio=pcm(stereo)}.
2026/05/12 21:29:07 [fade_in.1:3] Content type is {audio=pcm(stereo)}.
2026/05/12 21:29:07 [track_audio_amplify.2:3] Content type is {audio=pcm(stereo)}.
2026/05/12 21:29:07 [source.dynamic.2:3] Content type is {audio=pcm(stereo)}.
2026/05/12 21:29:07 [buffer.producer.proxy:3] Content type is {audio=pcm(stereo)}.
````
It could be worth decreasing the log level of `Content type is...`

**Comment by toots** [comment]
> > @martinkirch bug should be fixed now thanks for reporting!
> 
> It is, thanks for the fix ! First tests look good. Now I'm trying to skip the leaving track.
> 
> Starting with [the script above](https://github.com/savonet/liquidsoap/pull/5077#issuecomment-4290704408) :
> 
> ```ocaml
> music = playlist("/home/martin/Musique")
> ...
> programs = fallback([
> ...
>   music.{
>     track_sensitive = false,
>     on_leave = fun(_) -> begin log("music.leave"); music.skip(); end
>   }
> ```
> 
> That plays, but it seems `on_leave` is not called at all : no log message and the track resumes where it was when the sources is back on air.
> 
> So I tried another syntax, inspired by the previous examples of `on_stream_select`, but that doesn't type check:
> 
> ```ocaml
> def on_music_leave({ending, track_sensitive}) =
>   log("music.leave")
>   ending.skip()
> end
> ...
>     on_leave = on_music_leave,
> ```
> 
> ```
> At test.liq, line 22 char 20 - line 32 char 1:
> programs = fallback([
>   stream.{
> ...
>   }
> ])
> 
> Error 5: this value has type
>   [_
>    .{
>      on_leave : (_.{ending : _, source : _, track_sensitive : _}, ...) -> _,
>      on_select : _,
>      track_sensitive : _
>    }]
> but it should be a subtype of
>   [_
>    .{
>      on_leave? : ({source : _, track_sensitive : _}, ...) -> _,
>      on_select? : _,
>      track_sensitive? : _
>    }]
> ```
> 
> Surprisingly, writing `def on_music_leave(ending, track_sensitive) =` works, but ignores the function entirely.
> 
> sidenote: using a complex on_stream_select seem to create many operators, and each one logs a lot, on every transition they printed:
> 
> ```
> 026/05/12 21:29:07 [switch:3] Switch to buffer.producer with transition.
> 2026/05/12 21:29:07 [source.2:3] Content type is {audio=pcm(stereo)}.
> 2026/05/12 21:29:07 [audio.add:3] Content type is {audio=pcm(stereo)}.
> 2026/05/12 21:29:07 [fade_out:3] Content type is {audio=pcm(stereo)}.
> 2026/05/12 21:29:07 [track_audio_amplify.1:3] Content type is {audio=pcm(stereo)}.
> 2026/05/12 21:29:07 [source.dynamic.1:3] Content type is {audio=pcm(stereo)}.
> 2026/05/12 21:29:07 [max_duration:3] Content type is {audio=pcm(stereo)}.
> 2026/05/12 21:29:07 [sequence:3] Content type is {audio=pcm(stereo)}.
> 2026/05/12 21:29:07 [blank:3] Content type is {audio=pcm(stereo)}.
> 2026/05/12 21:29:07 [fade_in.1:3] Content type is {audio=pcm(stereo)}.
> 2026/05/12 21:29:07 [track_audio_amplify.2:3] Content type is {audio=pcm(stereo)}.
> 2026/05/12 21:29:07 [source.dynamic.2:3] Content type is {audio=pcm(stereo)}.
> 2026/05/12 21:29:07 [buffer.producer.proxy:3] Content type is {audio=pcm(stereo)}.
> ```
> 
> It could be worth decreasing the log level of `Content type is...`

Yeah I want to be better at that but for now, your problem.

This is another subtle trick with `on_leave`. As I said earlier, we want to make sure the source is properly left before calling `on_leave` to avoid interfering with live data.

In your example (which was also the current `stdlib` implementation), if the source keeps going in:
```
max_duration(2., sequence([old, switch([])]))
```
then `switch` keeps relaying it and (before the latest changes), `max_duration` too.

I have fixed `max_duration` so that it properly releases its source now. You can use it and don't need to use `switch` anymore. Your script should be working fine now.

**Comment by oranjebloom** [comment]
> @oranjebloom I made the `max_fade` a `settings.source.composition.max_fade`. If there is more than `max_fade` during transition, the track ends without a fade-out!

@toots thanks for making this part of the core implementation. I've been very happy with the way the new switch behaves on it's own. Fading occurs when it should and short tracks / tracks with less time remaining than the max-fade limit play out when they should. So good!!!

However I have noticed something that occurs when I pair the switch with a fallback in a hierarchical structure (see the minimal example below). Not sure if this is the appropriate thread to post it in or whether it requires a new one specific to this issue (if so please let me know). 

Everything behaves as it should in the minimal script below, but I notice an audio artifact (a pop-like sound) at the start of a new source (the delayed jingle) only when the fallback is coupled with the switch. Not sure if it is the fallback, delay, or switch that is the culprit. It seems similar to the kinds of audio issues issues mentioned in [#4828](https://github.com/savonet/liquidsoap/issues/4828) and [#5019](https://github.com/savonet/liquidsoap/issues/5019).

```
settings.source.composition.max_fade := 10.

# SOURCE A
a_jingles = playlist("/path/to/files")
a_music = playlist("/path/to/files")
a_jingle = delay(300.,b_jingles)
source_a = fallback([a_jingle.{track_sensitive = true},a_music.{track_sensitive = true}])

# SOURCE B
b_jingles = playlist("/path/to/files")
b_music = playlist("/path/to/files")
b_jingle = delay(300.,b_jingles)
source_b = fallback([b_jingle.{track_sensitive = true},b_music.{track_sensitive = true}])

# SCHEDULE
radio = switch(
  [
    ({21h00-22h00}, source_b.{track_sensitive = false}),
    ({true}, source_a.{track_sensitive = false})
  ]
)

```

**Comment by martinkirch** [comment]
> I have fixed `max_duration` so that it properly releases its source now. You can use it and don't need to use `switch` anymore. Your script should be working fine now.

It does :tada: 

It even works if `music` is a `request.queue`, and that queue can be modified even when it's offline and that is a neat improvement.

The final boss will be metadata. I'll need a few days to port "the big script" and check how it goes.

**Comment by oranjebloom** [comment]
I've done some further tests and can confirm the following behaviour below.

- No pop sound artifact occurs on the first transition from source_a to source_b
- A pop sound artifact occurs on every subsequent transition (either when leaving source_b to source_a or vice versa)
- This does NOT happen with simple playlist sources directly in the switch (without nested fallbacks).



I set the schedule to switch between sources every minute for the purpose of testing.

If it was never intended that fallbacks be nested within the switch, and this is not advisable, then what would the best practice be?

 ```
settings.source.composition.max_fade := 10.
 
 # SOURCE A
 a_jingles = playlist("/path/to/files")
 a_music = playlist("/path/to/files")
 a_jingle = delay(300.,b_jingles)
 source_a = fallback([a_jingle.{track_sensitive = true},a_music.{track_sensitive = true}])

# SOURCE B
b_jingles = playlist("/path/to/files")
b_music = playlist("/path/to/files")
b_jingle = delay(300.,b_jingles)
source_b = fallback([b_jingle.{track_sensitive = true},b_music.{track_sensitive = true}])
 
# SCHEDULE
radio = switch(
 [
 ({21h01-21h02}, source_b.{track_sensitive = false}),
 ({21h03-21h04}, source_b.{track_sensitive = false}),
 ({21h05-21h06}, source_b.{track_sensitive = false}),
 ({true}, source_a.{track_sensitive = false})
  ]
 )
 ```

**Comment by martinkirch** [comment]
Hello,
Still testing. I noticed the pop, too. Probably the last sample from the previous song, that should have been skipped on_leave. Happened in a more complex setting, like
```bash
fallback([ # all track_sensitive=false
    harbor,
    http.input,
    request.queue,
    fallback([  # both track_sensitive=true
        delay(request.queue),
        request.queue # skip on_leave
    ]),
    blank()
])
```

I'm still trying to find the best `repeat_metadata` combination in that stack. The script relies on `on_metadata` and I've seen weird cases, like metadata was one track late, or the delayed queue's md was flushed entirely. I'll try to write a targeted test case.

martinkirch:
So I instrumented a bit the common case example given initially, to keep an eye on metadata by the way:
```bash
server.harbor() # go to http://localhost:8000/telnet , type stream.start / .stop

stream = buffer(input.http(id="stream", "https://live.campusgrenoble.org/rcg112", start=false))
music  = playlist("/home/martin/Musique")
backup = single("/home/martin/radiotomate/tests/assets/GTA3Theme.mp3")

radio = fallback([stream, music, backup])

def log_md(md) =
  log("-----METADATA!------------")
  log(md["artist"] ^ " - " ^ md["title"] ^ " - " ^ md["source"] ^ " - " ^ md["initial_uri"])
end

radio.on_metadata(synchronous=false, log_md)

output(id="o", radio)
output.file(%wav, "test_default_output.wav", radio)
```
what happened:
 - music track 1: all OK
 - music track 2: all OK (glad it picked a loud example)
 - `stream.start` fades out then fades in
 - stream plays (this one helps the test because it has a lower level)
 - `stream.stop` does not fade, but OK
 - third track plays, but showed two metadata entries. log says, in a flash:
   * `METADATA! [track2]`
   * `Could not update timestamps for skipped samples.` (not sure if linked)
   * `METADATA! [track3]`
  
and indeed the waveform confirms there's one last sample from track 2:
Capture d’écran du 2026-05-19 23-27-24-annotated
(silence in the center is me editing the waveform)

oranjebloom:
@martinkirch glad you hear it too and good work on visually representing the problem.

toots:
Thanks for the detailed report! Skipping has always been tricky and part of the goal for this work is to make it work more naturally so I'm not surprised I'm tripping here hehe.

I have pushed what should be a fix. I'm still reviewing it but meanwhile this should be testable with it now!

martinkirch:
Yeah we are now reaching the hardest layers: metadata & skips. And my first tests are OK ! Soon enough to be announced during the liquidshop ;)

I'll test this against TheBigScript in the upcoming days.

martinkirch:
Caught a metadata bug: updated to cdd5b4c60, and added delayed jingles to last week's example so we now have a fallback in a fallback:

```
stream = buffer(input.http(id="stream", "https://live.campusgrenoble.org/rcg112", start=false))
music  = playlist("/home/martin/Musique")
backup = single("/home/martin/radiotomate/tests/assets/GTA3Theme.mp3")

jingles = delay(180., playlist("/home/martin/radiotomate/radio_data/carts/1Jingles"))
autodj = fallback([jingles, music])

radio = fallback([stream, autodj, backup])

def log_md(md) =
  log("-----METADATA!------------")
  log(md["title"] ^ " - " ^ md["source"] ^ " - " ^ md["initial_uri"])
end

radio.on_metadata(synchronous=false, log_md)

output(radio)
```

The auto-dj works, inserts jingles, but jingles' metadata is not correct: it repeats the first track's metadata.

Alsto tried with [TheBigScript](https://foss.heptapod.net/radiotomate/radiotomate/-/blob/topic/devel/ls2.5/playout/radiotomate.liq?ref_type=heads#L146), where everything is a `request.queue` : the first jingle logs correct metadata, then all others log empty MD.

oranjebloom:
@martinkirch are you still getting those pop sounds as well?

martinkirch:
> @martinkirch are you still getting those pop sounds as well?

no but in that case I'm queuing sounds that start and end silently, so if the phantom sample is still there it would be just a few ms of silence.

I'll be far from my test computer for a few days so I can't test with more adversarial sounds right now

toots:
@martinkirch: could you try again with the latest commit and make sure to set `synchronous=true` to the on_metadata handler to make sure order is preserved? I cannot reproduce here..

martinkirch:
@toots I've simplified the script and used some generated waveforms to check if a phantom sample was hiding: sound is correct, but metadata is not. You'll find the script, basic wavs and recorded debug log + output at http://radiotomate.org/tmp/liquidsoap_test_01468999f.zip

The log contains:

```
2026/06/21 17:06:09 [main:3] Liquidsoap 2.5.0+git@01468999f
2026/06/21 17:06:12 [lang:3] ---------- ON_METADATA:::::::::::::::: 
2026/06/21 17:06:12 [lang:3] Test Radio - Jingle Sine 220hz - jingles/Jingle Sine 220hz.mp3
2026/06/21 17:06:16 [decoder:2] Decoding "jingles/Jingle Sine 220hz.mp3" ended: Ffmpeg_decoder.End_of_file.
2026/06/21 17:06:16 [lang:3] ---------- ON_METADATA:::::::::::::::: 
2026/06/21 17:06:16 [lang:3] Test Radio - Track Triangle 440 - music/Track Triangle 440.mp3
2026/06/21 17:06:46 [decoder:2] Decoding "music/Track Triangle 440.mp3" ended: Ffmpeg_decoder.End_of_file.
2026/06/21 17:06:46 [lang:3] ---------- ON_METADATA:::::::::::::::: 
2026/06/21 17:06:46 [lang:3] Test Radio - Jingle Sine 220hz - jingles/Jingle Sine 220hz.mp3
2026/06/21 17:06:50 [decoder:2] Decoding "jingles/Jingle Sine 880hz.mp3" ended: Ffmpeg_decoder.End_of_file.
```

no "Sine 220hz" played at 17:06:46, the last `on_metadata` is not correct :/

toots:
@martinkirch metadata bug should be fixed now!

toots:
Ok tests are now fully passing. More testing would be nice but otherwise we should consider merging this!

martinkirch:
@toots sorry for the bad news but it seems the metadata bug is still here, and we even have a regression: with the simplest test script, when the last source in the main fallback is `blank()` (instead of `backup = single(..)`), the stream does not even start. I've put a copy of the debug log of this case in [this gist](https://gist.github.com/martinkirch/7879bede47cc222cc342473b8bfe2036) (the app  is stuck at `Queued 1 request(s)`)

toots:
> @toots sorry for the bad news but it seems the metadata bug is still here, and we even have a regression: with the simplest test script, when the last source in the main fallback is `blank()` (instead of `backup = single(..)`), the stream does not even start. I've put a copy of the debug log of this case in [this gist](https://gist.github.com/martinkirch/7879bede47cc222cc342473b8bfe2036) (the app is stuck at `Queued 1 request(s)`)

Yeah well there's something I had not thought about: how to handle the last source in a `fallback` or `switch` when it's supposed to be a fallback.

In the current context, this means the last source should be of `live` composition:

```liquidsoap
b = blank()
b.composition_type := "live"
radio = fallback([autodj, b])
```

I'm not sure if that is gonna be super user-friendly and I'm not sure if most users would expect the last source to be a fallback...

Something to think about.

martinkirch:
> In the current context, this means the last source should be of `live` composition:

why ?

I'm confused because `single` works fine but it's a `file` composition, right?

> I'm not sure if most users would expect the last source to be a fallback...

That's hard to tell, but as outputs want a safe source I'd bet it's a common pattern.

toots:
> > In the current context, this means the last source should be of `live` composition:
> 
> why ?
> 
> I'm confused because `single` works fine but it's a `file` composition, right?
> 
> > I'm not sure if most users would expect the last source to be a fallback...
> 
> That's hard to tell, but as outputs want a safe source I'd bet it's a common pattern.

One side-effect that I did not realize in the original design is that the composition type influences how to switch *to* and *from*. De-factor, a `live` composition sources currently forces `track_sensitive` when switching away from it too.

This is actually reasonable: a live source may never have track marks, for instance a talk show, so if there is a source higher in the switch/fallback priority, you might want to also switch immediately.

This makes `blank` work because `blank` does not have track mark so, when switching between blank and a file-base sources, there is never an opportunity for a track mark to switch away..

I'm thinking on how to clarify all these, I want to continue thinking about the other use-case to make sure that we can cover all reasonable simple user needs. I think the direction is good just need to think it through a bit more.

toots:
@martinkirch I just finished another pass, fixed some bugs and clarified the behavior. I think it's worth giving it another try! Thanks!

martinkirch:
> @martinkirch I just finished another pass, fixed some bugs and clarified the behavior. I think it's worth giving it another try! Thanks!

That's an interesting approach (again) ! However, be careful that documentation's examples only cover `fallback(something_live, something_file)`.

Current branch passes basic tests, so I switched to the complete script, [who looks simpler now](https://foss.heptapod.net/radiotomate/radiotomate/-/commit/2367cd129f19d5f0f0cabcc544a55bf9f23f9667) but found a bug. The script can be summarized as:
```
fallback(
   harbor.input,
   http.input(start=false),
   request.queue(id="programs"),
   fallback(id="autodj")
       delay(1000., request.queue(id="jingles")),
       request.queue(id="music")
   )
   blank()
)
```
To workaround blank's `live` composition,  I've set `autodj.{track_sensitive=true}` (this used to be `fallback(track_sensitive=true`).

An external program always ensures jingles and music always have something (and they do), but I noticed that sometimes one last frame of music plays during the handover from `programs` to `autodj`, and it that case auto-DJ's fallback does not handover to jingle's delay, although it should have played right after a long enough program, but it does after one or two tracks. For now I can hardly be more precise than "sometimes", but if that helps note that the complete script wraps all file-based sources in `blank.eat(start_blank=true`.

The upside is that metadata is always correct, both from `on_metadata` and fetched regularly from the main output :+1:

toots:
Thanks for checking @martinkirch. I pushed some changes that should have fixed that, wanna try again?

martinkirch:
That's an interesting version, because it highlights some quirks in _my_ script. In particular it fallbacks too easily to blank, logging `Ending blank.proxy.1 with a 1.0s fade out..` and that second delay sounds long. I'll improve that and test again, meanwhile I noticed other details:

*  Blank's fade-out was preceded by `[autodj_music.proxy.6:3] Switching to autodj_music.proxy.6: inserting latest metadata` although I've put `repeat_metadata=false` all around. It *sometimes* results in an unwanted `on_metadata` call, and I'm sure I heard a phantom frame from the previous track once.
* when a live source takes over, it seems the previous track is not skipped right after leaving, as if it was stuck in the auto-DJ queue - I'm polling `autodj_queue.current()` to know when it's time to push the next track, and I noticed that it happened only when the live source handed back to the AutoDJ.
   - as a sidenote: I'm very interested in the possibility to poll/update queues even when they're not playing, to ensure auto-DJ queues are always ready to play a track that should be playing at the current time. But "what should be playing" may evolve while they're silenced by a long live stream.
* the latter problem could be hiding behind jingles, but it seems that `delay(queue(id="jingles"))` is not re-enabling the source fast enough.

toots:
@martinkirch thanks for you continued feedback. What script should I use for reference for your last comment?

martinkirch:
I'm kinda glad it doesn't take the whole requests-and-polling stack :sweat_smile: , I found a problem in a more basic script:

```
server.harbor()

programs = request.queue(id="programs")

music  = playlist("music")
jingles = delay(20., playlist("jingles"))
autodj = fallback([jingles, music])

radio = fallback([programs, autodj.{track_sensitive = false}, blank()])

def log_md(md) =
  log("---------- ON_METADATA:::::::::::::::: ")
  log(md["artist"] ^ " - " ^ md["title"] ^ " - " ^ md["initial_uri"])
end
radio.on_metadata(synchronous=true, log_md)

output(id="o", radio)
output.file(%mp3, "output.mp3", radio)
```

 - start the script
 - open http://localhost:8000/telnet
 - during some music, type `programs.push short_program.wav`
 - right after the program, jingles should play, but only a short slice (and its metadata) are outputted

Log starting from my `push`:
```
2026/07/29 21:43:11 [decoder.ffmpeg.description:3] Analyzing container for format: "wav", uri: "/home/martin/radiotomate/radiotomate_data/carts/5shorttest/chimes.wav"
2026/07/29 21:43:11 [decoder.ffmpeg:3] FFmpeg recognizes "/home/martin/radiotomate/radiotomate_data/carts/5shorttest/chimes.wav" as audio: {codec: pcm_s16le, 48000Hz, 2 channel(s)}
2026/07/29 21:43:11 [decoder.ffmpeg.description:3] Requested content-type: {audio=pcm(stereo)}
2026/07/29 21:43:11 [decoder.ffmpeg.description:3] Decoded content-type: {audio=pcm(stereo)}
2026/07/29 21:43:11 [programs:3] Prepared "/home/martin/radiotomate/radiotomate_data/carts/5shorttest/chimes.wav" (RID 7).
2026/07/29 21:43:11 [switch.1:3] Switch to programs with transition.
2026/07/29 21:43:11 [switch.proxy:3] Ending switch.proxy with a 1.0s fade out..
2026/07/29 21:43:11 [source.dynamic.3:3] Content type is {audio=pcm(stereo)}.
2026/07/29 21:43:11 [fade_out.1:3] Content type is {audio=pcm(stereo)}.
2026/07/29 21:43:11 [track_audio_amplify.1:3] Content type is {audio=pcm(stereo)}.
2026/07/29 21:43:11 [source.dynamic.2:3] Content type is {audio=pcm(stereo)}.
2026/07/29 21:43:11 [max_duration.1:3] Content type is {audio=pcm(stereo)}.
2026/07/29 21:43:12 [switch.proxy:3] Source left in the middle of a track: skipping current track
2026/07/29 21:43:12 [programs.proxy:3] Switching to programs.proxy: no latest metadata to insert
2026/07/29 21:43:12 [programs.proxy:3] Content type is {audio=pcm(stereo)}.
2026/07/29 21:43:12 [decoder.ffmpeg.internal:3] Frame format change detected!
2026/07/29 21:43:12 [lang:3] ---------- ON_METADATA::::::::::::::::
2026/07/29 21:43:12 [lang:3]  -  - /home/martin/radiotomate/radiotomate_data/carts/5shorttest/chimes.wav
2026/07/29 21:43:35 [decoder:2] Decoding "/home/martin/radiotomate/radiotomate_data/carts/5shorttest/chimes.wav" ended: Ffmpeg_decoder.End_of_file.
2026/07/29 21:43:35 [switch:3] Switch to delay with track-sensitive transition.
2026/07/29 21:43:35 [delay.proxy:3] Switching to delay.proxy: inserting latest metadata
2026/07/29 21:43:35 [delay.proxy:3] Content type is {audio=pcm(stereo)}.
2026/07/29 21:43:35 [switch.1:3] Switch to switch with track-sensitive transition.
2026/07/29 21:43:35 [switch.proxy.1:3] Switching to switch.proxy.1: no latest metadata to insert
2026/07/29 21:43:35 [switch.proxy.1:3] Content type is {audio=pcm(stereo)}.
2026/07/29 21:43:35 [music:3] Prepared "music/Track Square 440-660.mp3" (RID 5).
[mp3float @ 0x7f60d80c1500] Could not update timestamps for skipped samples.
2026/07/29 21:43:35 [switch:3] Switch to music with track-sensitive transition.
2026/07/29 21:43:35 [music.proxy.2:3] Switching to music.proxy.2: inserting latest metadata
2026/07/29 21:43:35 [music.proxy.2:3] Content type is {audio=pcm(stereo)}.
2026/07/29 21:43:35 [switch:3] Source created multiple tracks in a single frame! Sub-frame tracks are not supported and are merged into a single one..
2026/07/29 21:43:35 [lang:3] ---------- ON_METADATA::::::::::::::::
2026/07/29 21:43:35 [lang:3] Test Radio - Jingle Sine 880hz - jingles/Jingle Sine 880hz.mp3
2026/07/29 21:43:35 [lang:3] ---------- ON_METADATA::::::::::::::::
2026/07/29 21:43:35 [lang:3] Test Radio - Track Square 440-660 - music/Track Square 440-660.mp3
```
metadata is correct: in the recording there's a short slice of the 880 sine before the "music" track. But it should have played entirely (`Jingle Sine 880hz.mp3` lasts 4 seconds).

toots:
@martinkirch Looks like your last feedback was another `delay` issue.. Think it should be fixed now if you want to check again..

martinkirch:
not yet :grimacing: 

it seems that pushing requests while the script is running makes a difference. Using this script:
````ocaml
programs = request.queue(id="programs")
music  = request.queue(id="music")
jingles = request.queue(id="jingles")
DELAY = 20. # the two test tracks last 30s
autodj = fallback([delay(DELAY, jingles), music])

radio = fallback([
  programs.{
    track_sensitive = false,
  },
  autodj.{
    track_sensitive = false,
  },
  blank().{
    track_sensitive = false,
  }]
)

jingles.push(request.create("jingles/Jingle Sine 220hz.mp3"))
music.push(request.create("music/Track Square 440-660.mp3"))

# push before the end of the first track:
thread.run(delay=DELAY - 5., { jingles.push(request.create("jingles/Jingle Sine 880hz.mp3")) })
thread.run(delay=DELAY - 5., { music.push(request.create("music/Track Triangle 440.mp3")) })

def log_md(md) =
  log("---------- ON_METADATA:::::::::::::::: ")
  log(md["artist"] ^ " - " ^ md["title"] ^ " - " ^ md["initial_uri"])
end

radio.on_metadata(synchronous=true, log_md)

output(id="o", radio)
output.file(%mp3, "output.mp3", radio)
````

 - the second jingle is not played between the two tracks, but it should have
 - on_metadata is called twice when tarting the second track

**Comment by toots** [comment]
@martinkirch looks like we're iterating on `delay` issues here.. 😆

Found and fixed.. Mind trying again?

**Comment by martinkirch** [comment]
I think we're good with `delay` ! Then there's `blank.eat` (yeah).

Test of the day: I firstly noticed that `replay_metadata = true` (default) mights be buggy for music playlists, because they're skipped when a higher-priority source plays, but the operator sill logs `[autodj_music.proxy.1:3] Switching to autodj_music.proxy.1: inserting latest metadata`. This results in two consecutive calls to `on_metadata`. That doesn't happen when setting `replay_metadata = false`, so OK... then I heard a phantom sample, probably from a previous track :hear_no_evil: 

so, back to the test script, apparently the key change is wrapping programs and music in `blank.eat` :

```
BLANK_MAX_DURATION = 1.
BLANK_THRESHOLD = -45.

programs_q = request.queue(id="programs_q")
programs = blank.eat(id="programs",
  start_blank=true,
  max_blank=BLANK_MAX_DURATION,
  threshold=BLANK_THRESHOLD,
  stereo( # If there's a mono file
    programs_q
  )
)

music_q  = request.queue(id="music_q")
autodj_music = blank.eat(id="autodj_music",
  start_blank=true,
  max_blank=BLANK_MAX_DURATION,
  threshold=BLANK_THRESHOLD,
  stereo( # If there's a mono file
    music_q
  )
)

jingles = request.queue(id="jingles")
DELAY = 20. # the two test tracks last 30s
autodj = fallback([delay(DELAY, jingles), autodj_music])

radio = fallback([
  programs.{
    track_sensitive = false,
  },
  autodj.{
    track_sensitive = false,
  },
  blank().{
    track_sensitive = false,
  }]
)

jingles.push(request.create("jingles/Jingle Sine 220hz.mp3"))
music_q.push(request.create("music/Track Square 440-660.mp3"))

thread.run(delay=DELAY - 10., { programs_q.push(request.create("/home/martin/radiotomate/radiotomate_data/carts/5shorttest/chimes.wav")) })

thread.run(delay=DELAY - 5., { jingles.push(request.create("jingles/Jingle Sine 880hz.mp3")) })
thread.run(delay=DELAY - 5., { music_q.push(request.create("music/Track Triangle 440.mp3")) })

def log_md(md) =
  log("---------- ON_METADATA:::::::::" ^ md["initial_uri"])
end

radio.on_metadata(synchronous=true, log_md)

output(id="o", radio)
output.file(%mp3, "output.mp3", radio)

```

The bug is after the "program":
 - a jingle plays, that's OK
 - a phantom sample from the skipped track is visible on the waveform, and raises on_metadata
 - main fallback transitions to `blank`
 - the track that should have played after the jingle starts after 2s (blank fade-in + out)

martinkirch (approved):
I think the core of the change is now correct.

But last tests showed it impacts some operators, potentially more than the ones I'm using. For example  I've read on Discord someone asked about `source.seek() `, that'd  be a great test, and there are probably a few others.

This could be worth a major version increment, because it will probably require most users to change their script. At least I think it's worth a pre-release (with packages and Dockers images) so more users can test, improving operators coverage.

But overall that's a significant improvement ! Thanks for keeping it up.

toots:
Awesome! I fixed your last issue too. Will let this sit a bit more and will merge if no other issue arises.

Issue #5138, opened 2026-05-04
`input.external.wav` panics when facing unexpected EOF


MikaSappi:
### Description

When `input.external.wav()` uses `socat` to read a non-existent file, LS panics. While this is a reasonable outcome, this would make Liquidsoap dependent on external programs and their lifecycle. 

*in this case, a whep/whip server restarting caused LS to crash*

### Steps to reproduce
```
s_in =
  mksafe(
    stereo(
      input.external.wav(
        id="aircore.edge.mic",
        buffer=0.1,
        max=1.,
        restart=true,
        restart_on_error=true,
        #restart_delay=10.,
        "socat -u UNIX-CONNECT:/tmp/path_to/non_existent/mic.pcm"
      )
    )
  )
```
### Expected behavior

Do not panic, use `restart_on_error`, and a `restart_delay`.

### Log extracts

```
2026/05/04 22:04:50 socat[14328] E connect(, LEN=24 AF=1 "/tmp/path_to/onair.pcm", 24): No such file or directory
2026/05/04 22:04:50 [aircore_edge_mic:4] 2026/05/04 22:04:50 socat[14329] E exactly 2 addresses required (there are 1); use option "-h" for help
2026/05/04 22:04:50 [aircore_edge_mic:4]
2026/05/04 22:04:50 [aircore_edge_mic:4]
2026/05/04 22:04:50 [aircore_edge_mic:3] Error while running process: End_of_file
2026/05/04 22:04:50 [aircore_edge_mic:3] Raised at Wav_aiff.callback_ops.really_input in file "src/core/base/tools/wav_aiff.ml", line 46, characters 22-39
2026/05/04 22:04:50 [aircore_edge_mic:3] Called from Wav_aiff.read_header.read_string in file "src/core/base/tools/wav_aiff.ml", line 96, characters 4-27
2026/05/04 22:04:50 [aircore_edge_mic:3] Called from Wav_aiff.read_header in file "src/core/base/tools/wav_aiff.ml", line 100, characters 10-26
2026/05/04 22:04:50 [aircore_edge_mic:3] Called from External_input_audio.(fun).read_header in file "src/core/base/sources/external_input_audio.ml", line 134, characters 21-68
2026/05/04 22:04:50 [aircore_edge_mic:3] Called from External_input.base.(fun).on_stdout in file "src/core/base/tools/external_input.ml", line 47, characters 24-42
[...]

PR #5139, opened 2026-05-04
Do not panic & crash when external input finds an EOF (#5138)

MikaSappi:
Exceptions in `process_handler.ml` go through `on_stop` instead of raising them again, and the drain loop logs errors instead of escaping. Added `restart_delay` to `External_input.base` (default `0.`), exposed on `input.external.wav` and `input.external.rawaudio` with a `1.` default. 

Video operator was left untouched because I had no reliable way to test it.

Prevents the crash reported in #5138.

### DIFF
--- a/src/core/base/tools/external_input.ml
+++ b/src/core/base/tools/external_input.ml
@@ -24,8 +24,8 @@ open Extralib
 
 (* {1 External Input handling} *)
 
-class virtual base ~name ~restart ~restart_on_error ~on_data ?read_header
-  command =
+class virtual base ~name ~restart ~restart_on_error ?(restart_delay = 0.)
+  ~on_data ?read_header command =
   let no_header, read_header =
     match read_header with
       | None -> (true, fun _ -> assert false)
@@ -60,9 +60,14 @@ class virtual base ~name ~restart ~restart_on_error ~on_data ?read_header
           let on_stop status =
             Generator.add_track_mark self#buffer;
             header_read <- false;
-            match status with
-              | `Status (Unix.WEXITED 0) -> restart
-              | _ -> restart_on_error
+            let should_restart =
+              match status with
+                | `Status (Unix.WEXITED 0) -> restart
+                | _ -> restart_on_error
+            in
+            if should_restart && restart_delay > 0. then
+              Thread.delay restart_delay;
+            should_restart
           in
           let log s = self#log#important "%s" s in
           process <-
--- a/src/core/base/tools/external_input.ml
+++ b/src/core/base/tools/external_input.ml
@@ -65,9 +65,7 @@ class virtual base ~name ~restart ~restart_on_error ?(restart_delay = 0.)
                 | `Status (Unix.WEXITED 0) -> restart
                 | _ -> restart_on_error
             in
-            if should_restart && restart_delay > 0. then
-              Thread.delay restart_delay;
-            should_restart
+            if should_restart then restart_delay else -1.
           in
           let log s = self#log#important "%s" s in
           process <-
--- a/src/core/base/tools/external_input.ml
+++ b/src/core/base/tools/external_input.ml
@@ -24,8 +24,8 @@ open Extralib
 
 (* {1 External Input handling} *)
 
-class virtual base ~name ~restart ~restart_on_error ?(restart_delay = 0.)
-  ~on_data ?read_header command =
+class virtual base ~name ~restart ~restart_on_error
+  ?(restart_delay_on_error = 0.) ~on_data ?read_header command =
   let no_header, read_header =
     match read_header with
       | None -> (true, fun _ -> assert false)
@@ -60,12 +60,9 @@ class virtual base ~name ~restart ~restart_on_error ?(restart_delay = 0.)
           let on_stop status =
             Generator.add_track_mark self#buffer;
             header_read <- false;
-            let should_restart =
-              match status with
-                | `Status (Unix.WEXITED 0) -> restart
-                | _ -> restart_on_error
-            in
-            if should_restart then restart_delay else -1.
+            match status with
+              | `Status (Unix.WEXITED 0) -> if restart then 0. else -1.
+              | _ -> if restart_on_error then restart_delay_on_error else -1.
           in
           let log s = self#log#important "%s" s in
           process <-
--- a/src/core/base/tools/process_handler.ml
+++ b/src/core/base/tools/process_handler.ml
@@ -145,7 +145,7 @@ let run ?priority ?env ?on_start ?on_stdin ?on_stdout ?on_stderr ?on_stop ?log
   let env = with_default (Unix.environment ()) env in
   let log = with_default (fun _ -> ()) log in
   let on_start = with_default (fun _ -> `Continue) on_start in
-  let on_stop = with_default (fun _ -> false) on_stop in
+  let on_stop = with_default (fun _ -> -1.) on_stop in
   let mutex = Mutex.create () in
   let create () =
     log "Starting process";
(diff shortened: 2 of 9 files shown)


src/core/base/tools/process_handler.ml
```diff
@@ -331,20 +335,16 @@ let run ?priority ?env ?on_start ?on_stdin ?on_stdout ?on_stderr ?on_stop ?log
...
-          match e with
-            | Wrapped e ->
-                f e;
-                raise e
-            | _ ->
-                f e;
-                restart_decision (on_stop (`Exception e)))
+          let e = match e with Wrapped e -> e | e -> e in
+          f e;
+          restart_decision (on_stop (`Exception e))
```
toots: The logic for `Wrapped` is to force the re-raise and bypass the `restart` logic

src/core/base/tools/external_input.ml
```diff
@@ -24,8 +24,8 @@ open Extralib
 
 (* {1 External Input handling} *)
 
-class virtual base ~name ~restart ~restart_on_error ~on_data ?read_header
-  command =
+class virtual base ~name ~restart ~restart_on_error ?(restart_delay = 0.)
```
toots: Should this be a tiny bit by default? Maybe `0.1`?

src/core/base/tools/external_input.ml
```diff
@@ -60,9 +60,14 @@ class virtual base ~name ~restart ~restart_on_error ~on_data ?read_header
...
-            match status with
-              | `Status (Unix.WEXITED 0) -> restart
-              | _ -> restart_on_error
+            let should_restart =
+              match status with
+                | `Status (Unix.WEXITED 0) -> restart
+                | _ -> restart_on_error
+            in
+            if should_restart && restart_delay > 0. then
+              Thread.delay restart_delay;
```
toots: Have you tested that? Will this impact the main streaming thread?

toots (commented):
Thanks for this!

I think that regarding restart delay, we should implement the delay at the process handler to use the duppy delayed task to prevent blocking whichever thread is running the `on_stop` callback.

I'm suggest to change `on_stop` to:
* negative: don't restart
* Otherwise delay by `d` and then restarts

I think it would be ok to use `0.` delay, it'll give the scheduler a change to run and process other tasks in case we're in a crash loop.

src/core/base/tools/external_input.ml
MikaSappi: `0.` or `0.1`? (The top-level comment says `0.`). For my use case, `0.` would be optimistic, but my use case isn't all of the use cases 😄 ... and to be fair, if we expose a parameter to set the restart time, the user has a chance to think about how often they want to try to restart.

src/core/base/tools/external_input.ml
MikaSappi: It probably will, switched to Duppy and testing it today. Didn't notice blocking in the original while testing since there was nothing to block.

src/core/base/tools/process_handler.ml
MikaSappi: I'm trying to preserve this in the revision I'm testing, let's see how it turns out.

src/core/base/tools/external_input.ml
MikaSappi: I'm starting to lean towards *at least* `0.1`, since if a process exits quickly, such as `socat`, we would be reading a non-existent file thousands of times a second, which feels not elegant at all. Could be even higher, and I think it comes down to whether it's more relevant to have a source ready as soon as possible, or accept e.g. a 500ms delay in the source becoming online, but not having to abuse the OS.

MikaSappi:
Looking nice:
```
__aircore_edge_mic =
  mksafe(
    stereo(
      input.external.wav(
        id="aircore.edge.mic",
        buffer=0.1,
        max=1.,
        restart=true,
        restart_delay=5.,
        restart_on_error=true,
        "socat -u - UNIX-CONNECT:#{__aircore_edge_in_socket}"
      )
    )
  )

[...]

def aircore_edge.monitor(~mix_minus=true, s) =
  feed =
    if
      mix_minus
    then
      add(normalize=false, [s, amplify(-1., __aircore_edge_mic)])
    else
      s
    end
  output.external(
    id="aircore.edge.onair",
    fallible=true,
    flush=true,
    register_telnet=false,
    reopen_delay=2.,
    reopen_on_error=fun (_) -> 2.,
    self_sync=false,
    start=true,
    %wav(channels = 2, header = true),
    {
      "socat -u - UNIX-CONNECT:#{__aircore_edge_out_socket}"
    },
    feed
  )
end
```
External service down:
```
2026/05/05 15:53:43 [_mic:4]
2026/05/05 15:53:43 [_mic:3] Cannot drain pipe: End_of_file
2026/05/05 15:53:43 [_mic:4]
2026/05/05 15:53:43 [_mic:3] Process exited with code 1
2026/05/05 15:53:45 socat[86348] E connect(, AF=1 "/tmp/ac-e/onair.pcm", 29): No such file or directory
sem_wait returned -1, errno = 4
2026/05/05 15:53:45 [_onair:3] Error while streaming: Lang.Runtime_error { kind: "system", msg: "Sys_error(\"Broken pipe\")", pos: [at src/core/base/outputs/pipe_output.ml, line 434, char 15-15] }, will re-open in 2.00s
2026/05/05 15:53:47 socat[86354] E connect(, AF=1 "/tmp/ac-e/onair.pcm", 29): No such file or directory
2026/05/05 15:53:47 [_onair:3] Error while streaming: Lang.Runtime_error { kind: "system", msg: "Sys_error(\"Broken pipe\")", pos: [at src/core/base/outputs/pipe_output.ml, line 434, char 15-15] }, will re-open in 2.00s
2026/05/05 15:53:48 [_mic:3] Starting process
2026/05/05 15:53:48 [_mic:4] 2026/05/05 15:53:48 socat[86360] E connect(, AF=1 "/tmp/ac-e/mic.pcm", 27): No such file or directory
2026/05/05 15:53:48 [_mic:4]
2026/05/05 15:53:48 [_mic:3] Cannot drain pipe: End_of_file
2026/05/05 15:53:48 [_mic:4]
2026/05/05 15:53:48 [_mic:3] Process exited with code 1
2026/05/05 15:53:48 [main_HLS_out:4] Terminating current segment 41 on stream lofi to make expected length
2026/05/05 15:53:49 socat[86363] E connect(, AF=1 "/tmp/ac-e/onair.pcm", 29): No such file or directory
sem_wait returned -1, errno = 4
2026/05/05 15:53:49 [_onair:3] Error while streaming: Lang.Runtime_error { kind: "system", msg: "Sys_error(\"Broken pipe\")", pos: [at src/core/base/outputs/pipe_output.ml, line 434, char 15-15] }, will re-open in 2.00s
```
No crashing, blocking, nor sadness.

toots (commented):
This looks great! Let me know when it's ready for a final review.

MikaSappi:
Right, the formatter had strong opinions, but now the tests pass. 

Notes: 

The class-default for `restart_delay` in `src/core/base/tools/external_input.ml` is still `0.` to not propagate the changes to the video side of things. For audio, the default (`0.1`) is declared in the Proto in `src/core/base/sources/external_input_audio.ml`. 

The `Wrapped` behavior is preserved (and now somewhat understood). Now it just doesn't raise an `e` and kill the whole process.

`src/core/base/operators/pipe.ml` bools mapped to floats.

Tests changed accordingly.

src/core/base/tools/external_input.ml
toots: Let's do `0.` actually unless you want to split up restart delay on error and restart delay on legit case. The happy path should be happy and if it's an expected restart, for instance if you need to restart the encoder to insert metadata (as with ogg format) then you want `0.` by default.

toots:
This is good to me. Let me know if you want to do `0.` or split restart delay on error/legit cases and to `0.1` for errors and `0.` for normal. We could also just have no params for normal restarts in fact. Prolly the best solution.

MikaSappi:
Now, for exit status 0, it doesn't take in the parameter, and for any other, it delays by `restart_delay_on_error`. I took the liberty of renaming that parameter to not let the users think it applies to all kinds of restarts. Let me know if it's too verbose. 

Class-default is `0.`, in Proto it's `0.1`.

toots (approved):
Looks awesome. You were right on the rename. Thanks!

Issue #5149, opened 2026-05-10
Calling `s.skip()` from a scheduler thread (harbor http server) races `cross`'s clock


MikaSappi:
### Description

This might or might not be the correct deduction, but I was able to reproduce said issue.

When a scheduler calls `s.skip()`, it drives cross' own clock and causes a race where `source.dynamic()` fails assertion.

From convo #5148

### Steps to reproduce

```
# Regression for https://github.com/savonet/liquidsoap/discussions/5148

def f() =
  iterations = ref(0)

  def next() =
    request.create("../media/@wav[stereo].wav")
  end

  s = request.dynamic(id="dyn", synchronous=true, next)

  s = cross(duration=0.5, fun (a, b) -> sequence([a.source, b.source]), s)

  clock.assign_new(sync="none", [s])
  output.dummy(fallible=true, s)

  # Hammer skip from another thread
  def hammer() =
    s.skip()
    ref.incr(iterations)
    if 1000 < iterations() then test.pass() end
    0.001
  end

  thread.run.recurrent(fast=true, delay=0.5, hammer)
end

test.check(f)

```

### Expected behavior

Run calls in cross' own thread.

### Log extracts

Full logs in convo #5148.

```
Error -1: Exception raised: Clock.Has_stopped
Raised at Clock.check_stopped in file "src/core/base/clock.ml" (inlined), line 295, characters 63-80
Called from Clock._tick in file "src/core/base/clock.ml", line 566, characters 2-18
Called from Cross.cross#child_get in file "src/core/base/operators/cross.ml", lines 248-254, characters 6-37
Called from Cross.cross#buffer_before in file "src/core/base/operators/cross.ml", line 363, characters 24-60
Called from Cross.cross#prepare_before in file "src/core/base/operators/cross.ml", line 336, characters 6-42
Called from Cross.cross#abort_track in file "src/core/base/operators/cross.ml", line 596, characters 19-38
Called from Stdlib__List.iter in file "list.ml", line 114, characters 12-15
Called from Stdlib__List.iter in file "list.ml", line 114, characters 12-15
Called from Stdlib__List.iter in file "list.ml", line 114, characters 12-15
Called from Builtins_source.(fun) in file "src/core/builtins/builtins_source.ml", line 66, characters 6-52
Called from Liquidsoap_lang__Evaluation.apply.f in file "src/lang/base/evaluation.ml", line 113, characters 8-12
[...]

PR #5194, opened 2026-06-03
Defer cross abort_track transition to its own clock thread

MikaSappi:
Calling `source.skip()` from a harbor or `thread.run` handler on a `request.dynamic` inside `cross`/`crossfade` crashes with `Assert_failure` at `request_dynamic.ml:145`. 

The cause is `cross#abort_track` running `prepare_before` on the caller's thread, racing the clock thread. The fix defers it to the cross's own clock thread via `Clock.on_tick`. 

A regression test (`GH5148.liq`) reproduces the assertion on current `main` and passes with the fix.

#5148 #5149

### DIFF
--- a/src/core/base/operators/cross.ml
+++ b/src/core/base/operators/cross.ml
@@ -475,13 +475,16 @@ class cross val_source ~override_duration ~duration_getter ~persist_override
         | `Idle -> self#source#effective_source
         | `Before (_, s) | `After (_, s) -> s#effective_source
 
+    val pending_abort_track = Atomic.make false
+
     method abort_track =
-      match status with
-        | `Idle -> self#source#abort_track
-        | `Before s | `After s ->
-            self#source#abort_track;
-            status <- `After s;
-            ignore self#prepare_before
+      self#source#abort_track;
+      if not (Atomic.exchange pending_abort_track true) then
+        Clock.on_tick self#clock (fun () ->
+            Atomic.set pending_abort_track false;
+            match status with
+              | `Idle -> ()
+              | `Before _ | `After _ -> ignore self#prepare_before)
   end
 
 let _ =
--- a/src/core/base/operators/cross.ml
+++ b/src/core/base/operators/cross.ml
@@ -477,14 +477,16 @@ class cross val_source ~override_duration ~duration_getter ~persist_override
 
     val pending_abort_track = Atomic.make false
 
-    method abort_track =
-      self#source#abort_track;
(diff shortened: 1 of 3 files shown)


toots:
Ho man great catch thank you. Can we try `on_before_streaming` instead of `on_tick`? This way the track should be aborted before the next streaming cycle.

I did something very simiar for `request_queue.ml` recently actually, this would looke like this:

```ocaml
    val should_skip = Atomic.make false

    val pending_abort_track = Atomic.make false

    initializer
      self#on_before_streaming_cycle (fun () ->
          if Atomic.pending_abort_track false then (
            match status with
              | `Idle -> ()
              | `Before _ | `After _ -> ignore self#prepare_before)

    def abort_track =
      self#source#abort_track;
      Atomic.set pending_abort_track true
```

MikaSappi:
Pushing, let's see does the CI agree with me. 😄

toots:
Looking great. I need to have a pass at flakey tests. Thanks!

toots:
The backport to `v2.4.x-latest` failed:
```
The process '/usr/bin/git' failed with exit code 1
```
To backport manually, run these commands in your terminal:
```bash
# Fetch latest updates from GitHub
git fetch
# Create a new working tree
git worktree add .worktrees/backport-v2.4.x-latest v2.4.x-latest
# Navigate to the new working tree
cd .worktrees/backport-v2.4.x-latest
# Create a new branch
git switch --create backport-5194-to-v2.4.x-latest
# Cherry-pick the merged commit of this pull request and resolve the conflicts
git cherry-pick -x --mainline 1 156e67e9a6da3d548358c931b5ce3c477fa92293
# Push it to GitHub
git push --set-upstream origin backport-5194-to-v2.4.x-latest
# Go back to the original working tree
cd ../..
# Delete the working tree
git worktree remove .worktrees/backport-v2.4.x-latest
```
Then, create a pull request where the `base` branch is `v2.4.x-latest` and the `compare`/`head` branch is `backport-5194-to-v2.4.x-latest`.

MikaSappi:
Want me to do the backport, or are you gonna take care of it? I see there's a few conflicts in dune.inc, where I used the format as seen in `main`, and `v2.4.x` uses a different format. Actually, since it's only the dune.inc I could just patch it into the `v2.4.x`-style.  Let me know. 

Thanks either way!

toots:
@MikaSappi yeah if you want to backport that would be great! I would put a CHANGES entry there b/c we consider the bug fixed first in the 2.4.x stable release. Appreciate.

PR #5200, opened 2026-06-04
Backport #5194 to v2.4.x-latest

MikaSappi:
Calling source.skip() from a harbor or thread.run handler on a request.dynamic inside cross/crossfade crashes with Assert_failure at request_dynamic.ml:145.

The cause is cross#abort_track running prepare_before on the caller's thread, racing the clock thread. The fix defers it to the cross's own clock thread via Clock.on_tick.

A regression test (GH5148.liq) reproduces the assertion on current main and passes with the fix.

https://github.com/savonet/liquidsoap/discussions/5148 https://github.com/savonet/liquidsoap/issues/5149


MikaSappi:
`build_no_depopts / build_no_depopts` hangs. I can't open it so can't see is it because of what I've done or is it just being  cranky.

toots:
Thanks so much!

PR #5206, opened 2026-06-05
Reduce the use of finalizers.

toots:


toots:
cc/ @CharlesCarey

charles-rc:
@toots great find! Let me know when it's back ported so we can test it.

toots:
@charles-rc It's in! https://github.com/savonet/liquidsoap/releases/tag/rolling-release-v2.4.x

PR #5239, opened 2026-06-18
cry: drop obsolete bytes findlib dependency

attilagyorffy:
The `cry` library only uses the `Bytes` module, which has been part of the OCaml standard library since 4.07. The separate `bytes` findlib package is a compatibility shim for older compilers and is no longer needed; listing it breaks builds on toolchains that no longer ship it.

This removes `bytes` from `src/modules/cry/dune`. Verified `dune build src/modules/cry/` passes on OCaml 5.4.1 / dune 3.23.1.

Context: this was being worked around downstream in the Homebrew formula via an `inreplace`; submitting upstream so the workaround can be dropped (Homebrew/homebrew-core#273636).

### DIFF
--- a/src/modules/cry/dune
+++ b/src/modules/cry/dune
@@ -1,6 +1,6 @@
 (library
  (name cry)
- (libraries bytes unix stdlib_utils)
+ (libraries unix stdlib_utils)
  (foreign_stubs
   (language c)
   (names cry_stubs))


toots:
Sorry I'm just seeing this. This makes a lot of sense, thanks for sending!

attilagyorffy:
> Sorry I'm just seeing this. This makes a lot of sense, thanks for sending!

No worries @toots. This means the workaround in the [Homebrew formula](https://github.com/Homebrew/homebrew-core/pull/273636) can be dropped at the next release.

PR #5255, opened 2026-07-08
Fix harbor output scale

toots:


toots:
I did some testings, including using `wrk` and this showing some really good promises.

PR #5263, opened 2026-07-11
HLS: recover from transient output errors

### LINKED ISSUE #5259: output.file.hls crashes the process when the HLS directory becomes unwritable
I ran into this while investigating an outage: if the directory used by `output.file.hls` becomes unwritable or full while streaming, Liquidsoap exits instead of just losing/retrying the HLS output. Other outputs in the same process go down with it.

In my repro this fails at the atomic rename in `src/core/base/outputs/hls_output.ml:675` on the v2.4.5 tag, then reaches
`Clock.wrap_errors` and is re-raised. A full tmpfs gives the same class of failure with `Unix.ENOSPC` on write.

What surprised me is that `output.file` has `reopen_on_error` / `reopen_delay`, and `output.url` has `on_error` / `restart_delay`, but `output.file.hls` has no equivalent recovery hook. The operator is registered under the `output.file` namespace, but it does not inherit the piped output recovery behavior.

This came out of a real deployment issue here: https://github.com/oszuidwest/zwfm-liquidsoap/issues/100

### Steps to reproduce

Run Liquidsoap from the official v2.4.5 container with `/hls` writable by the `liquidsoap` user, using this script:

```liquidsoap
log.level := 4

s = sine(440.0)

output.dummy(id="primary", mksafe(buffer(s)))

aac = %ffmpeg(format = "mpegts", %audio(codec = "aac", b = "48k"))

output.file.hls(
  id="hls",
  segment_duration=2.0,
  segments=3,
  "/hls",
  [("aac", aac)],
  mksafe(buffer(s))
)
```

Once it is streaming:

```sh
chmod 555 /hls
```

I also reproduced the same kind of process-level failure by mounting `/hls` as a tiny tmpfs, for example 16K. That fails with `ENOSPC` on write.

### Expected behavior

A recoverable filesystem error in the HLS output should not take down the whole Liquidsoap process. Ideally `output.file.hls` would have an `on_error` / `reopen_on_error` style policy similar to `output.file` or `output.url`, or at least a way for the script to decide when to retry.

The unrelated `primary` output should keep running.

### Log extracts

```text
--- chmod 0555 /hls ---
Error: Unix.Unix_error(Unix.EACCES, "rename", "/hls/liq958fb6tmp")
Called from Hls_output.hls_output#open_out.object#close.(fun) in file "src/core/base/outputs/hls_output.ml", line 675, characters 19-45
Called from Hls_output.hls_output#close_segment in file "src/core/base/outputs/hls_output.ml", line 704, characters 10-18
Called from Hls_output.hls_output#reopen_segment in file "src/core/base/outputs/hls_output.ml", line 790, characters 10-30
Called from Hls_output.hls_output#write_pipe in file "src/core/base/outputs/hls_output.ml", line 1124, characters 12-43
Called from Clock.wrap_errors in file "src/core/base/clock.ml", line 514, characters 6-10

2026/07/10 22:38:15 [clock:2] Source hls failed while streaming: Unix.Unix_error(Unix.EACCES, "rename", "/hls/liq958fb6tmp")!
2026/07/10 22:38:15 [threads:3] Thread "Clock hls" aborts with exception Unix.Unix_error(Unix.EACCES, "rename", "/hls/liq958fb6tmp")!

Thread 2 killed on uncaught exception Unix.Unix_error(1, "rename", "/hls/liq958fb6tmp")
Re-raised at Clock.wrap_errors in file "src/core/base/clock.ml", line 525, characters 9-45

2026/07/10 22:38:15 [clock.primary:3] Clock stopping
2026/07/10 22:38:15 [clock.primary:3] Clock thread has stopped: no more sources to process, global stop.

Fatal error: exception Unix.Unix_error(Unix.EACCES, "rename", "/hls/liq958fb6tmp")
LIQUIDSOAP_EXIT=2
```

### Script extracts

```liquidsoap
log.level := 4

s = sine(440.0)

output.dummy(id="primary", mksafe(buffer(s)))

aac = %ffmpeg(format = "mpegts", %audio(codec = "aac", b = "48k"))

output.file.hls(
  id="hls",
  segment_duration=2.0,
  segments=3,
  "/hls",
  [("aac", aac)],
  mksafe(buffer(s))
)
```

### Liquidsoap version

```text
Liquidsoap 2.4.5
Copyright (c) 2003-2026 Savonet team
Liquidsoap is open-source software, released under GNU General Public License.
See <http://liquidsoap.info> for more information.
```

### Liquidsoap build config

```text
 * Liquidsoap version  : 2.4.5

 * Compilation options
   - Release build       : true
   - Git SHA             : (none)
   - OCaml version       : 4.14.2
   - OS type             : Unix
   - architecture        : amd64
   - system              : linux

 * Configured paths
   - mode              : posix
   - standard library  : /usr/share/liquidsoap/libs

 * Supported output formats
   - FDK-AAC           : yes
   - FFmpeg            : yes
   - MP3               : yes
   - Opus              : yes
   - Vorbis            : yes
   - WAV/AIFF          : yes (native)

This is from the official savonet/liquidsoap:v2.4.5 container. I can paste the
full `liquidsoap --build-config` output if needed; it is mostly the long libs
version list.
```

### Installation method

From official container image.

### FFmpeg CLI reproduction

N/A.

### Additional Info

I also checked the current `main` source at `de68528`. The relevant HLS write path and the lack of an `output.file.hls` recovery hook still appear to be present, with shifted line numbers.

rmens:
## Summary

- add `reopen_on_error` support to `output.file.hls`, with a default two-second retry delay
- abort incomplete temporary segments after an error while preserving published HLS state and marking the resumed stream as discontinuous
- share `reopen_on_error` parsing and retry scheduling with the existing piped file outputs
- add a regression test that makes the HLS output path unavailable and verifies that segment generation resumes

## Testing

- `GH5259`
- `GH2756`
- `GH2756-2`

Fixes #5259

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -37,6 +37,8 @@
 
 ## Fixed:
 
+- Fixed `output.file.hls` terminating the process on transient filesystem
+  errors. The output now supports a `reopen_on_error` callback (#5259).
 - Fixed ID3 timed metadata missing from first mpegts HLS segment: metadata is
   now registered before the segment is opened and before data is encoded (#5084).
 - Make active `stereotool` really be active.. (#4882)
--- a/src/core/base/outputs/hls_output.ml
+++ b/src/core/base/outputs/hls_output.ml
@@ -31,6 +31,9 @@ let default_name =
   Lang.eval ~cache:false ~typecheck:false ~stdlib:`Disabled
     {|fun (metadata) -> "#{metadata.stream_name}_#{metadata.position}.#{metadata.extname}"|}
 
+let default_reopen_on_error =
+  Lang.eval ~cache:false ~stdlib:`Disabled ~typecheck:false "fun (_) -> 2."
+
 let hls_proto frame_t =
   let main_playlist_writer_t =
     Lang.fun_t
@@ -98,6 +101,15 @@ let hls_proto frame_t =
   in
   Output.proto
   @ [
+      ( "reopen_on_error",
+        Lang.fun_t
+          [(false, "", Lang.nullable_t Lang.error_t)]
+          (Lang.nullable_t Lang.float_t),
+        Some default_reopen_on_error,
+        Some
+          "Callback called when there is an error. The error is raised when \
+           returning `null`; otherwise, the output is reopened after the \
+           returned delay, in seconds." );
       ( "playlist",
         Lang.string_t,
         Some (Lang.string "stream.m3u8"),
@@ -175,7 +187,8 @@ let hls_proto frame_t =
     ]
 
 type atomic_out_channel =
-  < output_string : string -> unit
+  < abort : unit
+  ; output_string : string -> unit
   ; output_substring : string -> int -> int -> unit
   ; position : int
   ; truncate : int -> unit
@@ -347,6 +360,16 @@ class hls_output p =
   let autostart = Lang.to_bool (List.assoc "start" p) in
   let infallible = not (Lang.to_bool (List.assoc "fallible" p)) in
   let register_telnet = Lang.to_bool (List.assoc "register_telnet" p) in
+  let reopen_on_error = List.assoc "reopen_on_error" p in
+  let reopen_on_error ~bt exn =
+    let error = Lang.runtime_error_of_exception ~bt ~kind:"output" exn in
+    match
+      Lang.to_valued_option Lang.to_float
+        (Lang.apply reopen_on_error [("", Lang.error error)])
+    with
+      | Some delay when 0. <= delay -> delay
+      | _ -> -1.
+  in
   let prefix = Lang.to_string (List.assoc "prefix" p) in
   let main_playlist_writer =
     Option.map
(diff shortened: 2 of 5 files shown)


toots:
Thanks for this one. Do you think that 2 seconds is a sensible default for all users? What would someone using `output.file.hls` expect when errors happen if they haven't thought about it yet?

rmens:
Thinking it through, I'd actually defend recovery by default, and I think the codebase already agrees.

The pattern across existing outputs is that live-delivery outputs retry by default while file outputs fail hard: `output.icecast` reconnects after 3 seconds unless a callback overrides it ("Will try to reconnect in 3.00 seconds"), and `output.url` defaults `restart_delay` to `2.`. Only `output.file` defaults `reopen_on_error` to `null`, which makes sense for archiving, where losing data silently would be worse than stopping. HLS is live delivery that happens to go through the filesystem, so I'd argue it belongs with `icecast`/`url`, not with archiving.

For someone who hasn't thought about it yet, the current default is also not "this output fails loudly", it's "the whole process exits". Taking down every other output with it (that's how I hit #5259: one unwritable HLS directory silenced outputs that had nothing to do with HLS). For broadcast use, a down HLS rendition with a log line every 2 seconds seems like the better surprise than dead air on everything.

The legitimate concern with retry-by-default is that a permanent error (disk full, bad permissions) becomes a quiet retry loop. I'd be happy to address that by also adding an `on_error` callback like `output.icecast` and `output.url` have, so scripts can alert on every failure rather than it only living in the logs.

On the amount of seconds itself I have no strong attachment: 2 matched `output.url`, but I'm equally fine aligning with icecast's 3, or another value you prefer.

toots:
Ok let's go with it thanks!

Issue #5227, opened 2026-06-09
2.4.4 hangs under certain conditions when crossfade is used


MikaSappi:
EDIT September 2, 2026: **this bug is not present in `2.5.0`**. If you've got a chance, and your script relies on mechanics that incur the described behavior below, you could benefit from using `2.5.0` instead of `v2.4.x-latest`.

### Description

In LS 2.4.x, the `crossfade` operator can start looping its buffer indefinitely. It can act two ways: either the whole source stalls, with `output.file` looping, OR only `output.icecast` is locked in a few-hundred-ms chunk.

I'm pointing my finger at `crossfade`, since as soon as that was replaced by `metadata.map`, the stall didn't happen (at least in \~15 minutes).

### Steps to reproduce
```
settings.log.level.set(4)

# required: container's 1Hz status poller connects here
settings.server.socket := true
settings.server.socket.path := "/tmp/liquidsoap.sock"

audio_dir = "/repro/audio"

track_i = ref(0)
def next_track_request() =
  path = "#{audio_dir}/track#{1 + (track_i() mod 8)}.mp3"
  track_i := track_i() + 1
  # sizes cross's buffer window; crossfade()'s duration=0. below is unused
  cross_duration = 0.5 + random.float(min=0., max=2.0)
  request.create("annotate:liq_cross_duration=#{cross_duration}:#{path}")
end

jingle_i = ref(0)
def next_jingle_request() =
  path = "#{audio_dir}/jingle#{1 + (jingle_i() mod 2)}.mp3"
  jingle_i := jingle_i() + 1
  request.create(path)
end

tracks = request.queue(id="tracks", prefetch=0)
jingles = request.queue(id="jingles", prefetch=0)

tracks_cross = crossfade(id="tracks_cross", fade_in=0., fade_out=0., duration=0., tracks)

def refeed(_, _) =
  thread.run(
    fun () -> begin
      ignore(process.read("sleep 0.2"))
      tracks.push(next_track_request())
    end
  )
end
tracks.on_position(remaining=true, position=5., synchronous=false, allow_partial=true, refeed)

def push_jingle(_) =
  thread.run(delay=0.5, fun () -> jingles.push(next_jingle_request()))
end
tracks_cross.on_metadata(synchronous=true, push_jingle)

def status(_) =
  "#{tracks_cross.is_ready()}"
end
server.register(usage="status", description="Status of the repro", "status", status)

[...]

PR #5267, opened 2026-07-11
Clock cleanup phase 2

toots:
## Why

`Child_support` and `Producer_consumer` were two overlapping mechanisms for the same job: letting an operator animate a child source in its own clock and consume it at a different pace than it produces. `Producer_consumer` opted out of the generic child clock animation by toggling `output_enabled` around each pull, which meant that when several operators read from the same child source, only the one currently pulling recorded the data and the others silently skipped those frames.

This reworks the ideas of #4637, which could not be rebased: it predates push-based sync source propagation (#5133) and the `controller` concept.

## What changed

**One mechanism.** `Producer_consumer` is removed. `Child_support.base` wraps its child in a single `child_output` living in the child clock; each tick of that clock processes the child's frame into a generator that the operator reads from through `child_get_frame`. `Child_support.producer` replaces `Producer_consumer.producer` for the ffmpeg inline encoder, decoder and bitstream filters, and a `process_frame` hook replaces the `write_frame` closure for operators that transform the data rather than buffer it verbatim (`stretch`, `soundtouch`, the ffmpeg builtins).

**Ticks carry whether they were requested by a reader.** `Clock.tick ~pull:true` marks a tick as coming from a source reading from that clock, and `child_output` only processes a frame on such a tick. A parent clock still ticks its sub-clocks, which is needed because a child clock is not purely pull-driven — inline encoding puts a real output inside it and an active source below that, and neither is pulled by anyone — but that tick now animates them without buffering data nobody asked for. Conversely, a pull fills every `child_output` on the clock, so operators sharing a child source all see the same data instead of skipping each other's frames.

**Readers consuming at diverging rates raise instead of buffering without bound.** A lone reader cannot accumulate: it stops ticking as soon as it holds a frame worth of data. Readers sharing a child clock can, if one consumes slower than the others. Rates legitimately diverge for a while, e.g. while encoders fill their lookahead, so this is caught by bounding each buffer rather than by comparing readers: past `settings.clock.child.max_buffer` (10 seconds by default) an error names the operator that fell behind and points at the setting.

**Buffered data is drained rather than dropped.** Readiness goes through `child_is_ready`, which also covers data the child already produced, so an operator still holding a remainder — `cross` in particular, which consumes only up to the next track mark — finishes it instead of dropping it once the child source goes down.

Two smaller fixes: a producer only reports its child's sync source while the child is actually producing, matching what `Clock_base.self_sync` does, so an idle child does not take the sync role in the parent clock; and `Generator.slice` gained a `peek` argument so a reader can look at a frame worth of data and decide to consume less of it, e.g. to stop at a track mark.

## Tests

`tests/core/child_support_test.ml` covers the shared child clock: a tick issued by one reader fills the other reader's buffer, a tick that is not a pull buffers nothing, `process_frame` receives `Flush` when the producer sleeps, and two readers consuming at diverging rates raise.


MikaSappi:
Hi @toots , any chance of getting a backport of this to `v2.4.x-latest`? #5227 seems it could be fixed with the logic you did in this PR. I tried a few workarounds, but ended up breaking more stuff. I confirmed just now that the bug described in #5227 still occurs in `v2.4.x-latest`.

toots:
> Hi @toots , any chance of getting a backport of this to `v2.4.x-latest`? #5227 seems it could be fixed with the logic you did in this PR. I tried a few workarounds, but ended up breaking more stuff. I confirmed just now that the bug described in #5227 still occurs in `v2.4.x-latest`.

Will have a look. Any chance you might be able to move to `2.5.x`? Really hoping to get things switched over there and focus on new dev for that version..

toots:
@MikaSappi reason being: this is a large refactor. Usually larger than what I would consider for a bugfix cycle..

MikaSappi:
Moving to 2.5.0 would probably be the most sensible thing here. 

After all, the bug is an edge case from a very unusual configuration I don't believe many users would use, and probably the impact of fixing it in 2.4.x would be less than the cost in time and effort it'd take to backport.

PR #5296, opened 2026-08-02
Fix sequence dropping the first chunk of the source it moves to.

toots:
When `sequence` is done with a source mid-frame it advances to the next one but carries the `After_position` reselect constraint over. That position was computed against the source we just finished with and does not apply to a source that has not been read yet during this streaming cycle, so a source with no more samples available than the one it follows gets rejected and its first chunk is silently dropped, along with the metadata it carries.

This shows up with `cross`/`crossfade` transitions built with `sequence([before, after])`: when the outgoing track sets a cross duration of a single frame, `create_after` hands both sides buffers of the same length, the incoming side is rejected and the stream keeps advertising the outgoing track's metadata until the next track boundary.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -43,6 +43,11 @@
 
 - Fixed `delay` getting stuck after its first track and dropping the following
   track's metadata (#5282).
+- Fixed `sequence` dropping the first chunk of a source, along with its
+  metadata, when the source it follows leaves no room in the current frame.
+  This showed up as the track after a `cross`/`crossfade` transition built with
+  `sequence` keeping the previous track's metadata until the next track
+  boundary (#TODO).
 - Fixed `output.file.hls` terminating the process on transient filesystem
   errors. The output now supports a `reopen_on_error` callback (#5259).
 - Fixed `output.file.hls` accepting unwritable output, temporary and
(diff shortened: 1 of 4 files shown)


toots:
The backport to `v2.4.x-latest` failed:
```
The process '/usr/bin/git' failed with exit code 1
```
To backport manually, run these commands in your terminal:
```bash
# Fetch latest updates from GitHub
git fetch
# Create a new working tree
git worktree add .worktrees/backport-v2.4.x-latest v2.4.x-latest
# Navigate to the new working tree
cd .worktrees/backport-v2.4.x-latest
# Create a new branch
git switch --create backport-5296-to-v2.4.x-latest
# Cherry-pick the merged commit of this pull request and resolve the conflicts
git cherry-pick -x --mainline 1 905c57fda31127bff36e2ea3226838153c06a407
# Push it to GitHub
git push --set-upstream origin backport-5296-to-v2.4.x-latest
# Go back to the original working tree
cd ../..
# Delete the working tree
git worktree remove .worktrees/backport-v2.4.x-latest
```
Then, create a pull request where the `base` branch is `v2.4.x-latest` and the `compare`/`head` branch is `backport-5296-to-v2.4.x-latest`.

PR #5328, opened 2026-08-09
hls: wait for a keyframe before splitting a segment (2.4.x)

toots:
Backport of #5327 to `v2.4.x-latest`, re-verified against this branch. `CHANGES.md` is updated under 2.4.6.

The one difference: `wait_for_keyframe` defaults to `false` here, so this branch's behaviour is unchanged unless you opt in.

## Reproducing it

Keyframes every 4s, segments asked every 1s, so most splits land with no keyframe available:

```liquidsoap
enc = %ffmpeg(format="mpegts", %audio(codec="aac", b="64k"),
              %video(codec="libx264", g=100, b="300k", preset="ultrafast"))
output.file.hls(segment_duration=1., segments=10, ".", [("s", enc)], s)
```

20s of that logs `Splitting segment without a new keyframe!` 13 times, and the segments it writes are not media:

```
$ ffprobe -v error -select_streams v:0 -count_frames s_12.ts
non-existing PPS 0 referenced
no frame!
Stream #0:2: Video: h264, none, 25 fps          <- no resolution
```

The playlist is also invalid on its own terms, and was before this change: `#EXT-X-TARGETDURATION:1` while carrying `#EXTINF:1.300`. Segments already overshoot at frame boundaries, and RFC 8216 §4.3.3.1 requires every segment to fit within the announced target duration.

`ffmpeg -hls_time 1` on the same encode, as the reporter described:

```
#EXT-X-TARGETDURATION:4
#EXTINF:4.000000,
```

100 video frames per segment, clean decode.

## Changes

**`wait_for_keyframe`, default `false`.** When the duration target is reached with no keyframe to split on, wait for the next one instead of splitting anyway. The machinery was already there: `should_reopen` returns whether a split may proceed without a segmentable position, and metadata and extra-tag splits already decline to. This only changes what happens in the case that is currently broken; when keyframes line up with `segment_duration`, nothing moves.

On `main` this defaults to `true`; here it stays `false` to keep backward compatibility.

**`EXT-X-TARGETDURATION` is the longest segment in the playlist** rather than `segment_duration`. Required once segments may run long, and a fix on its own for the `1.300` case above.

With `wait_for_keyframe=true`: 0 forced splits, `EXT-X-TARGETDURATION:5`, `EXTINF:4.020`, and segments start on a keyframe.

## Test

`tests/streams/hls_wait_for_keyframe.liq` runs the setup above and asserts, from the playlist, that segments reach the keyframe interval rather than being cut at `segment_duration`, and that the target duration covers the longest one. It fails with `wait_for_keyframe=false` ("longest segment is 1.3s") and passes with `true`. The existing HLS tests stay green.

## Something else this turned up, not fixed here

Splitting on a keyframe is necessary but not sufficient. With `wait_for_keyframe` on, the segment does start on a keyframe, and ffmpeg still decodes exactly one frame out of each segment past the first:

```
$ ffmpeg -v error -i s_3.ts -map 0:v -f null -
[h264] non-existing PPS 0 referenced
frame=    1
```

against 100 for an `ffmpeg -hls_time` segment of the same encode. Concatenate our segments and the stream is perfect — 401 frames, 16.04s, 25fps — so the data is all there and only per-segment independence is broken.

I could not pin the cause and do not want to guess in a commit message. Two things I ruled out: the parameter sets are present (`trace_headers` finds 2 SPS and 2 PPS per segment, same as ffmpeg's), and the segments are correctly TS-aligned (size a multiple of 188, first byte `0x47`). Worth its own issue; it is arguably the larger half of what the reporter is seeing, and this PR does not close it.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,15 @@
 
 ## Fixed:
 
+- HLS: `EXT-X-TARGETDURATION` now covers the longest segment in the playlist
+  instead of always announcing `segment_duration`. A segment that overshoots,
+  which already happened at frame boundaries, made the playlist violate the HLS
+  specification (#5266)
+- HLS: added `wait_for_keyframe`. When reaching `segment_duration` with no
+  keyframe to split on, the output waits for the next one instead of splitting
+  anyway, which is what the `ffmpeg` HLS muxer does with `-hls_time`. Segments
+  then last longer than `segment_duration` but start on a keyframe. Defaults to
+  `false` on this branch, keeping the existing behaviour (#5266)
 - Fixed `file.watch` losing track of a file that gets replaced rather than
   written in place: an inotify watch follows the inode, so a writer doing the
   usual write-to-temporary-then-rename silenced the watcher for good. The
--- a/doc/content/hls_output.md
+++ b/doc/content/hls_output.md
@@ -37,8 +37,13 @@ Keyframes are very common in video codecs and can exist in audio codecs.
 In order to make sure that a HLS playlist can be decoded starting from any segment, liquidsoap tries to split segments on keyframe boundaries. When not possible,
 you will see a warning in the logs.
 
-Segment split is forced when reaching the value specified by `EXT-X-TARGETDURATION` to follow the HLS specifications. For metadata and extra tags, segment split will
-occur at the next keyframe.
+Segment split is forced when reaching `segment_duration`, so that segments stay close to the requested length. When no keyframe is available at that point, the
+segment is split anyway and a warning is logged: the resulting segment does not start with a keyframe and cannot be decoded on its own. For metadata and extra tags,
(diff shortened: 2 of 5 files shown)


toots:
Closing: this belongs on `main` only.

RFC 8216 §6.2.1 says "The value of the EXT-X-TARGETDURATION tag in the Media Playlist MUST NOT change", and §4.3.3.1 requires every segment's `EXTINF`, rounded to the nearest integer, to be at most that value. Waiting for a keyframe makes segment length depend on the source's GOP structure, which is only known at runtime, so the sliding-window maximum this PR computes is itself a §6.2.1 violation.

The fix needs a declared upper bound instead, which is a new parameter and a behaviour change — not something to backport. Reworking on `main` in #5327.

PR #5335, opened 2026-08-11
source: allow on_frame callbacks to be deregistered

charles-rc:
### The problem

`source#on_frame` appends to a list with no removal path:

```ocaml
val mutable on_frame : on_frame list = []
method on_frame fn = on_frame <- on_frame @ [fn]
```

So a callback lives as long as the source it was registered on. That is fine
when an operator registers on a source it created, but `fade.in`, `fade.out`
and `fade.skip` register on the source they are *given*:

```liquidsoap
s = source.methods(s)
s.on_track(synchronous=true, reset_overrides)
s.on_metadata(synchronous=true, update_fade)
```

A `switch`/`fallback` transition calls those on the same long-lived source on
every switch, so two or three closures accumulate on it permanently.

This is a regression from 2.2.5, where the equivalents were wrapper functions
(`s = source.on_track(s, fn)` returned a new source that died with the wrapper).
The 2.4.0 callback rework turned them into methods that mutate the receiver,
which silently changed callback lifetime from "as long as the wrapper" to
"forever".

It is easy to miss in production because nothing looks wrong: the source count
stays flat (the closures attach to an existing source) and the objects are
tiny, so RSS barely moves. But major GC marking cost scales with live *block*
count, so CPU climbs steadily instead. It took us months to track down as an
unexplained "CPU leak".

### The fix

Three coordinated changes, following the deregistration pattern already used
for `on_sync_source_change` and extended in #5163 / #5164:

1. **`source.ml`** — `on_frame` entries carry an id and registration returns a
   deregistration closure.
2. **`lang_source.ml`** — script-level callbacks return `unit` decorated with a
   `remove` method. This is deliberately not a bare handle: `can_ignore` calls
   `demeth` before checking for unit, so **every existing script that discards
   the return value keeps typechecking unchanged**, while new code can hold the
   handle and release the callback.
3. **`fades.liq`** — the three fade operators bind their registrations to the
   per-invocation source they return, releasing on shutdown and re-acquiring on
   wake up.

The `register` field of the `callback` record now returns `unit -> unit`. Where
the underlying registration has no deregistration path of its own
(`on_wake_up`, `on_sleep`, `on_blank`, `on_connect`, …) a `disarmable` adapter
guards the callback with a flag, so `remove` still does what it promises.

### Measurements

Three arms on a production streaming script, switching between playlist sources
in a `switch`. Slope fitted after warm-up; live blocks accumulated per switch,
95% CI:

| arm | OCaml | `fades.liq` | blocks/switch | switches |
| --- | --- | --- | --- | --- |
| release build | 4.14.2 | stock | +32.5 ± 7.3 | 22,000 |
| control | 5.4.0 | stock | +24.5 ± 0.8 | 41,000 |
| patched | 5.4.0 | patched | **−0.018 ± 0.088** | 130,000 |

The control arm is the same binary and runtime as the patched one, differing
only in whether `fades.liq` releases its registrations — so the effect is
attributable to the fix rather than to the runtime. After the patch the
confidence interval contains zero and excludes anything above 0.07
blocks/switch. An idle control (no switches) is flat to ±20 blocks over 5
minutes.

### Tests

Adds `tests/regression/on-frame-deregistration.liq`, covering removal,
double-removal being a no-op, and unremoved callbacks still firing.

Every existing fade and crossfade test passes against the patched build:
`streams/cross`, `streams/cross-override`, `streams/cross-persist-override`,
`streams/cross-sequence-transition`, `streams/crossfade`, `streams/crossfade2`,
`streams/fades-overrides`, `streams/fades-persistent-override`,
`regression/GH3239`, `regression/GH3239-2`, `regression/AC5109` and
`regression/switch_per_source_no_reselect_during_fade`.

### Backport

Could this get the `backport v2.4.x-latest` label? 2.4.x is affected and we
would like to fix this before bumping to 2.5.x.

Heads-up that the automatic cherry-pick will not apply: `main` has since moved
`src/core/base/` to `src/core/source/`. I have the same change already
written against `v2.4.x-latest`, and am happy to open it as a manual backport
PR on request:
https://github.com/charles-rc/liquidsoap/tree/rc/callback-deregistration

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -63,6 +63,12 @@
 
 ## Fixed:
 
+- Fixed callbacks registered via `source.on_frame` and friends accumulating for
+  the lifetime of the source they were registered on. `fade.in`, `fade.out` and
+  `fade.skip` register on the source they are given, so a switch transition
+  calling them on the same long-lived source leaked closures on every switch.
+  Callback registration now returns a `remove` method (#TODO)
+
 - An FFmpeg filter graph is no longer finished off by an input that runs dry.
   A source is unavailable for all sorts of passing reasons, a `buffer` filling
   up or a queue waiting on a request, and liquidsoap has no way to tell those
--- a/src/core/operators/cross.ml
+++ b/src/core/operators/cross.ml
@@ -183,7 +183,10 @@ class cross val_source ~override_duration ~duration_getter ~persist_override
               self#set_main_duration
             with _ -> ())
 
-    initializer self#on_frame (`Metadata self#process_override_metadata)
+    initializer
+      ignore
+        (self#on_frame (`Metadata self#process_override_metadata)
+          : unit -> unit)
 
     method private append mode buf_frame =
       let l = Frame.get_all_metadata buf_frame in
--- a/src/core/operators/noblank.ml
+++ b/src/core/operators/noblank.ml
@@ -284,7 +284,11 @@ let _ =
           descr = "when detecting a blank.";
           register_deprecated_argument = false;
           arg_t = [];
-          register = (fun ~params:_ s f -> s#on_blank (fun () -> f []));
+          register =
+            (fun ~params:_ s f ->
+              let f, remove = Lang_source.disarmable f in
+              s#on_blank (fun () -> f []);
+              remove);
         };
         {
           name = "on_noise";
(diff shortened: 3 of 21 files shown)


charles-rc:
One unrelated commit is included: `tests: reformat rotate.liq per liquidsoap-prettier`.

`tests/streams/rotate.liq` is not formatted according to
`liquidsoap-prettier@1.8.2`, the version pinned in `.pre-commit-config.yaml` —
it wants `{weight = 3}` rather than `{weight=3}`. This reproduces against a
pristine `main` checkout, so it is not caused by this PR.

Runs on `main` pass because they restore a cached pre-commit environment; the
cache key only covers the Python version and the hash of
`.pre-commit-config.yaml`, so a stale hook environment keeps being reused. Any
run that builds the environment from scratch — such as a PR from a fork — picks
up 1.8.2 and fails. It will affect the next contributor too.

Happy to split that commit into its own PR if you would rather keep this one
focused.

toots:
Thanks for reporting. Closing in favor of https://github.com/savonet/liquidsoap/pull/5338 which addresses the issue with a broader scope.

PR #5336, opened 2026-08-12
Bound Unifier retention with union by rank

charles-rc:
`Unifier.( <-- )` always attaches the left tree's root to the right one's. Links point from the attached root to the surviving one, so a node retains everything it points at, and `find_root` only compresses paths that are actually dereferenced. When unification repeatedly goes through the most recently created node, nothing ever dereferences from the head — its path is never compressed, and it retains every node created after it.

This is reachable in normal operation. `Lang_source.check_content` runs on every runtime operator instantiation and, for each source argument, does `Typing.(s#frame_type <: fresh_copy)`, reaching `Content_base.merge`. Scripts that build operators per track therefore accumulate live `Unifier` nodes on every track change — `fade.in`/`fade.out` wrap their argument in `source.dynamic`, and `cross.simple` builds a fresh fade pair per transition.

### Changes

- `src/core/utils/unifier.ml`: nodes carry a rank; `( <-- )` attaches the shallower tree under the deeper one, so a fresh singleton is linked under an established node rather than the reverse. The contract that the right-hand value survives is preserved — when rank forces attaching `r'` under `r`, `r'`'s value moves into `r` first, so both still dereference to it.
- `tests/core/unifier_test.ml`: assert that retention does not scale with the number of unifications.

### Measurements

Compacted `live_blocks` against track count, on a looping playlist with `crossfade(duration=1.)` and a `%wav` output:

| | blocks/track |
| --- | --- |
| before | +24.4 |
| after | flat |

The new test retains 5 words per unification before the change and a constant ~19 words after.

### DIFF
--- a/src/core/utils/unifier.ml
+++ b/src/core/utils/unifier.ml
@@ -22,48 +22,61 @@
 
 (* Simple unification module for variables with no unknown value *)
 
-type 'a t = [ `Value of 'a | `Link of 'a t Atomic.t ]
+(* A node retains whatever it links to, and [find_root] only compresses paths
+   that are actually walked. Ranks let [<--] attach the shallower tree under
+   the deeper one, so a node never ends up pointing at one created after it. *)
+type 'a t = [ `Value of 'a | `Link of 'a link ]
+and 'a link = { atom : 'a t Atomic.t; mutable rank : int }
 
-let make v = `Link (Atomic.make (`Value v))
+let make v = `Link { atom = Atomic.make (`Value v); rank = 0 }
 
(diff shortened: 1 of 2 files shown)


charles-rc:
@toots I think this should cleanly go into 2.4.x once ready. Should it get the `backport v2.4.x-latest` label?

toots:
> @toots I think this should cleanly go into 2.4.x once ready. Should it get the `backport v2.4.x-latest` label?

Sorry for the delay. The changes look great and, yeah, will backport too.

PR #5356, opened 2026-08-24
Don't discard existing RG tags, and prevent double calculation

MikaSappi:
This changes the behavior of the combination of `enable_replaygain_metadata()` and `replaygain:`. Now, received metadata is not discarded but is used. This prevents calculations happening twice for one track.

The same applies to `enable_lufs_track_gain_metadata()` combined with `lufs_track_gain:`, which did the same thing.

Fixes: #5355

PR #5358, opened 2026-08-24
Don't discard existing RG tags, and prevent double calculation (#5356)

MikaSappi:
(cherry picked from commit 03d4a9bd003978bb2864d1d9a117278c3898da15)


toots:
Sorry I let this slide, will fix and merge asap

PR #5359, opened 2026-08-25
Run duppy on a pool of domains

toots:
Stacked on #5364.

Duppy ran every task on a single core through a CPS monad that OCaml 4 required. It now dispatches onto a pool of domains, so requests, harbor clients, script callbacks and `thread.run` handlers run in parallel and heavy work in a callback no longer stalls playback. Computations park on an effect and carry on where they left off instead of being split by the monad, and the event loop watches descriptors through a set the kernel keeps rather than rescanning.

`settings.scheduler.generic_queues`, `fast_queues` and `non_blocking_queues` are deprecated — the scheduler sizes itself from the core count. `settings.scheduler.blocking_tasks` caps how many slow tasks run at once.

`settings.scheduler.legacy` is a fail-safe for a script that concurrent execution breaks: Duppy runs a pool of threads on the main domain instead, one per configured queue, each taking only the priorities its queue accepts, so nothing runs in parallel as before. The queue settings only apply in that mode and warn when set without it. It goes away in a later version.

Migration notes cover the change and the fail-safe, and a new "sharing state between tasks" section documents `atomic` and `exchange`: a group of writes is only seen whole by a reader that groups its reads the same way, and one reference holding a record needs neither.

Running script code on several cores exposed four races that CI hit one per run: a switch reading its predicates one at a time, playlist parser registration where claiming a flag did not wait for the registering, a playlist reload racing `next()` into a permanent stopped state, and a harbor request whose port closed under it, which killed the scheduler.

Tests: `scheduler_legacy.liq` asserts three handlers on one legacy queue never overlap, and fails without the setting. The duppy suite gains a thread-pool check: every task reports the main domain and each priority stays on its own thread. The three new doc examples are type-checked by the doc rules.

CI was green before the legacy and doc commits: 303 tests ok, 0 failed.


toots:
Notes from @smimram:
* Is there a possibility of preemption for blocking tasks on the new model?
* Can we have a legacy mode?
* How does this compare with eio or lwt or lwt_eio?

martinkirch:
The documentation looks good !

I've tested this branch for more than 24h, using a script that sends and receives HTTP requests and relies on request queues: CPU and RAM levels were stable.

toots:
> The documentation looks good !
> 
> I've tested this branch for more than 24h, using a script that sends and receives HTTP requests and relies on request queues: CPU and RAM levels were stable.

Nice!! About to merge.

PR #5372, opened 2026-08-30
Native font

smimram:


src/core/operators/video_text_native.ml
```diff
@@ -25,19 +25,20 @@ open Extralib
...
 
-(* The font's character map is shared, and rendering happens on whichever
-   domain runs the source. *)
-module Font = Image.Bitmap.Font.Make (Lazy.Mutexed)
+module Font = Image.Bitmap.Font
 
 let warned_custom_font = ref false
 
+(* The native font, computed on first use. *)
+let native_font = Lazy.Mutexed.from_fun Font.native
```
toots: This would make mm with with ocaml 5.5 or above only. Are you cool with that?

toots (approved):

PR #5386, opened 2026-09-02
Switch from yamlx to miniyaml.

smimram:


toots (commented):
Let's make miniyaml an internal synced module. I'll get onto it. Thanks for this!!

PR #5388, opened 2026-09-02
ffmpeg: spread decoding, encoding and scaling over every core

toots:
Closes #5014.

## Why

libavcodec opens a codec context with `thread_count = 1` unless the codec overrides it, and ffmpeg's own tools raise that to automatic before opening (`fftools/ffmpeg_dec.c`, `fftools/ffmpeg_enc.c`). liquidsoap never set it, so every decoder ran on a single thread, and so did every encoder except `libx264` — the one codec shipping `{"threads", "0"}` in its own defaults. That is why the `threads=2` in the report looked ignored: x264 was already threading, and the single-threaded 4K decode was the actual bottleneck.

## What changed

Every codec is now opened through one helper that defaults the thread count to automatic. `avcodec_open2` applies the caller's option dictionary itself, so an explicit `%ffmpeg(%video(codec="...", threads=N))` still wins over the default, and codecs that thread badly keep opting out through their own defaults.

Scaling has its own threads, which only `sws_scale_frame` reaches: `sws_scale` deliberately falls back to `slice_ctx[0]` whatever the context is set to. Conversion therefore goes through frames wrapping the planes the caller owns, in buffers that free nothing, and `Swscale.Make.create` takes the count. `settings.ffmpeg.scaling_threads` sets it, one per core by default, as ffmpeg does through its filter graphs.

## Measurements

8-core arm64, 20s of 3840x2160 h264 transcoded to 1280x720, measuring the content produced during a fixed 20s window (higher is better):

| pipeline | before | after |
|---|---|---|
| `libx264`, ultrafast | 23.6s @ 108% CPU | 32.5s @ 209% CPU |
| `libx265`, slow | 9.4s @ 252% CPU | 25.8s @ 418% CPU |

The x265 pipeline goes from 0.47x realtime, which stutters by construction, to 1.3x. The scaling half is the smaller share of that: scaling is roughly 15% of the serial cost of a video pipeline, so threading it is worth a few percent on its own, and the extra cores buy little once the codecs are already using them.

## Testing

No new automated test. The verification that could have said the opposite is an output comparison: a 5s 4K to 720p decode, scaled and written out as rawvideo, is byte-identical across the pre-change build, the new frame-based path at `scaling_threads=1`, and the same path at `0`.

Individual ffmpeg media tests fail identically with and without this changeset on my machine, the harness being unable to write its generated encoder scripts there, so CI is the real signal for those.

This touches `src/modules/synced/ffmpeg`, so the ocaml-ffmpeg mirror picks it up.

PR #5391, opened 2026-09-03
Stop and release active inputs replaced through source.dynamic

### LINKED ISSUE #5389: Active inputs replaced by source.dynamic are not stopped or collected
### Description

An active input returned by `source.dynamic` is detached when it is replaced, but it is not actually stopped. With `input.ffmpeg`, the decoder/container and its asynchronous polling task then remain reachable indefinitely.

This causes resources to grow with every replacement. It is especially visible with H.264 inputs because every retained decoder keeps its FFmpeg worker threads.

I reproduced this on both:

- `2.5.0+git@7978836`
- current `main` at `99c30465d7f845a274a3e9322a7191d897c54c7a`

### Minimal reproducer

This is self-contained and uses FFmpeg's `lavfi` input:

```liquidsoap
log.level := 4
settings.log.stdout := true
settings.log.file := false

stopped = ref(false)
collected = ref(false)

current = ref(source.fail())
dynamic = source.dynamic(track_sensitive=false, init=current(), fun () -> current())
output.dummy(mksafe(dynamic))

def replace() =
  current := input.ffmpeg(
    id="cleanup_second",
    start=true,
    format="lavfi",
    "sine=frequency=880:sample_rate=48000"
  )
end

def install() =
  first = input.ffmpeg(
    id="cleanup_first",
    start=true,
    format="lavfi",
    "sine=frequency=440:sample_rate=48000"
  )
  first.on_stop(
    synchronous=true,
    fun () -> begin stopped := true; print("STOPPED") end
  )
  first.on_collect(
    synchronous=true,
    fun () -> begin collected := true; print("COLLECTED") end
  )
  first.on_connect(
    synchronous=true,
    fun () -> thread.run(delay=0.2, replace)
  )
  current := first
end

thread.run(delay=0.1, install)
thread.run(
  delay=4.0,
  fun () -> begin
    runtime.gc.compact()
    print("RESULT stopped=#{stopped()} collected=#{collected()}")
    shutdown()
  end
)
```

### Actual result

`source.dynamic` does call `sleep` on the old activation, and the log contains:

```text
Source cleanup_first gets down.
RESULT stopped=false collected=false
```

Neither full major GC nor compaction changes the result.

With a local H.264/AAC MP4 and 12 replacements (13 inputs total), resources grew linearly:

```text
threads: 26 -> 206
file descriptors: 17 -> 53
socket/control descriptors: 10 -> 34
connected=13 disconnected=0 stopped=0 collected=0
```

A real HTTP run showed one TCP connection per old input, including old `ESTABLISHED` and `CLOSE-WAIT` sockets with unread receive queues.

### Root cause

There appear to be two independent cleanup gaps:

1. `Start_stop.active_source` registers `on_sleep` as `base#transition_to \`Stopped`. Since `transition_to` became asynchronous in #4849, the stop is queued. `Source.actual_sleep` has already detached the active source from its clock, so it never receives another streaming cycle in which to flush that queued transition.
2. Even when an explicit stop is allowed to complete before replacement, `input.ffmpeg` leaves its `Duppy.Async` `connect_task` registered. The task retains the source and its control socket pair, so `on_collect` still never runs.

Stopping while `Av.open_input` is polling also needs to participate in the FFmpeg interrupt callback; otherwise a canceled source can finish connecting after it has gone down.

### Tested fix

I tested a local patch that:

- executes the mandatory stop synchronously from `on_sleep` using `execute_transition`;
- stops and clears the FFmpeg input's `Duppy.Async` task after a stop;
- interrupts an in-progress `Av.open_input` when the source is stopping;
- treats stale task wakeups as no-ops instead of asserting that the status is always `Starting`.

Results with the patch:

- 100 H.264/AAC replacements over local HTTP: memory stayed around 95–100 MiB after warm-up, with 26 steady threads, 17 FDs and 11 socket/control descriptors.
- Every retired input was disconnected/stopped and became collectible; only the current input remained active.
- A complete HLS pipeline with 40 two-second source replacements stayed around 143–147 MiB RSS after warm-up, with stable thread/FD counts; the resulting six-segment fMP4 playlist decoded as H.264/AAC.
- Explicit `stop()` / `start()` cycles still reconnect correctly.
- A 3200-second `input.ffmpeg.seek()` still returned 3200 and resumed at the requested PTS under a 1 GiB cgroup limit.

I also have a regression test that fails on current `main` (`stopped=false`, `collected=false`) and passes with the patch. I can submit the patch and test as a PR if this diagnosis and fix direction look correct.

toots:
Fixes #5389.

An active input replaced through `source.dynamic` was detached from its clock but never actually stopped, and with `input.ffmpeg` never became collectable either. Every replacement leaked the previous input's connection, decoder and FFmpeg worker threads for the lifetime of the script.

Two independent gaps were involved. `Start_stop.active_source` registered its mandatory stop through `transition_to`, which has queued the transition for the next streaming cycle since #4849. A source that has just gone to sleep is detached from its clock and never gets that cycle, so the stop never ran. Outputs were already switched to the synchronous `execute_transition` in #4849; this does the same on the input side, which covers every active input, not only `input.ffmpeg`.

Separately, `input.ffmpeg` never stopped its `Duppy.Async` connect task on disconnect. The task closure kept the source and its wake-up socket pair reachable from the scheduler, so `on_collect` could not fire even after an explicit stop. The task is now stopped and cleared the way `input.srt` already does it. Since `Duppy.Async.stop` does not interrupt a task that is mid-`open_input`, `disconnect` also has to interrupt the connection itself, and the task's status writes are compare-and-set so a cancelled or stale run finalizes to `Stopped` without clobbering a newer `Starting`.

The interrupt callback stays installed on the `AVFormatContext` for the container's whole lifetime, so it cannot be keyed on the shared source status: doing so interrupts every read once connected. The per-attempt `closed` flag already carries exactly the right scope, it simply was not reachable from `disconnect` before the container existed. `` `Polling `` now carries it, `disconnect` raises it to interrupt a connection in progress, and the physical identity of that `` `Polling `` value doubles as the generation token that keeps a stale attempt from clobbering a newer one.

Added `tests/regression/GH5389.liq`, the reporter's reproducer reduced to the `on_stop`/`on_collect` assertions. It fails on `main` and passes with this change. `tests/regression/GH4840.liq` covers the connected-read side: it serves a stream over harbor and fails as soon as a connected input interrupts its own reads. Explicit `stop()`/`start()` cycles on `input.ffmpeg` were checked manually and still reconnect.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -105,6 +105,9 @@
 
 ## Fixed:
 
+- Active inputs such as `input.ffmpeg` replaced through `source.dynamic` are now stopped and
+  released. They used to keep their connection, decoder and threads alive for the lifetime of
+  the script (#5389).
 - A transition that drops the incoming source no longer drops that track's announcement with
   it. `cross` had already consumed the metadata into the buffer it hands the transition, so
   the track played on, audible and unannounced, once the transition was over. It is now
--- a/src/core/optionals/ffmpeg/io/ffmpeg_io.ml
+++ b/src/core/optionals/ffmpeg/io/ffmpeg_io.ml
@@ -106,15 +106,16 @@ class input ?(name = "input.ffmpeg") ~autostart ~self_sync ~poll_delay ~debug
     method private connect_task () =
       Generator.set_max_length self#buffer max_length;
       try
-        if self#source_status = `Stopping then raise Stopped;
-        assert (self#source_status = `Starting);
-        Atomic.set source_status `Polling;
+        if not (Atomic.compare_and_set source_status `Starting `Polling) then
+          raise Stopped;
         let opts = Hashtbl.copy opts in
         let url = self#url in
         let closed = Atomic.make false in
         let container =
           Av.open_input
-            ~interrupt:(fun () -> Atomic.get shutdown || Atomic.get closed)
+            ~interrupt:(fun () ->
+              Atomic.get shutdown || Atomic.get closed
+              || self#source_status <> `Polling)
             ?format ~opts
             ~configure_audio_stream:Ffmpeg_decoder_common.configure_audio_stream
             ~configure_video_stream:Ffmpeg_decoder_common.configure_video_stream
@@ -142,11 +143,23 @@ class input ?(name = "input.ffmpeg") ~autostart ~self_sync ~poll_delay ~debug
         List.iter (fun fn -> fn m) (Callbacks.elements on_connect);
         Generator.add_track_mark self#buffer;
         let container = { decoder; remaining; buffer; closed } in
-        Atomic.set source_status (`Connected (url, container));
+        if
+          not
+            (Atomic.compare_and_set source_status `Polling
+               (`Connected (url, container)))
+        then (
+          decoder.close ();
+          List.iter (fun fn -> fn ()) (Callbacks.elements on_disconnect);
+          raise Stopped);
         -1.
       with
         | Stopped ->
-            Atomic.set source_status `Stopped;
+            ignore (Atomic.compare_and_set source_status `Stopping `Stopped);
+            -1.
+        | _ when not (Atomic.compare_and_set source_status `Polling `Starting)
+          ->
+            (* [disconnect] interrupted the connection. *)
+            ignore (Atomic.compare_and_set source_status `Stopping `Stopped);
             -1.
         | e ->
             let bt = Printexc.get_raw_backtrace () in
@@ -156,31 +169,27 @@ class input ?(name = "input.ffmpeg") ~autostart ~self_sync ~poll_delay ~debug
             let err = Lang.runtime_error_of_exception ~bt ~kind:"ffmpeg" e in
             List.iter (fun fn -> fn err) (Callbacks.elements on_error);
             if debug then Printexc.raise_with_backtrace e bt;
-            Atomic.set source_status `Starting;
             poll_delay
 
     method private connect =
       match self#source_status with
         | `Starting | `Polling | `Connected _ -> ()
-        | `Stopping | `Stopped -> (
+        | `Stopping | `Stopped ->
             Atomic.set source_status `Starting;
-            match Atomic.get connect_task with
-              | Some t -> Duppy.Async.wake_up t
-              | None ->
-                  let t =
-                    Duppy.Async.add ~priority:`Blocking Tutils.scheduler
-                      self#connect_task
-                  in
-                  Atomic.set connect_task (Some t);
-                  Duppy.Async.wake_up t)
+            let t =
+              Duppy.Async.add ~priority:`Blocking Tutils.scheduler
+                self#connect_task
+            in
+            Atomic.set connect_task (Some t);
+            Duppy.Async.wake_up t
 
     method private disconnect =
       let stop_task () =
-        match Atomic.get connect_task with
-          | None -> Atomic.set source_status `Stopped
-          | Some t ->
-              Atomic.set source_status `Stopping;
-              Duppy.Async.wake_up t
+        (* A task mid-connection sees [`Stopping] and finalizes itself. The
+           stopped task no longer retains this source. *)
+        if not (Atomic.compare_and_set source_status `Polling `Stopping) then
+          Atomic.set source_status `Stopped;
+        Option.iter Duppy.Async.stop (Atomic.exchange connect_task None)
       in
       match self#source_status with
         | `Stopping | `Stopped -> ()
--- a/src/core/optionals/ffmpeg/io/ffmpeg_io.ml
+++ b/src/core/optionals/ffmpeg/io/ffmpeg_io.ml
@@ -50,7 +50,7 @@ class input ?(name = "input.ffmpeg") ~autostart ~self_sync ~poll_delay ~debug
     val source_status
         : [ `Stopped
           | `Starting
-          | `Polling
+          | `Polling of bool Atomic.t
           | `Connected of string * container
           | `Stopping ]
           Atomic.t =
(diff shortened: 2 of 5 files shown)


oneingan:
Thanks for picking this up so quickly. I tested the exact PR head `e68b3e73461e8382ccaad34baa0858cfc02399fa` (OCaml 5.5, FFmpeg 8.1.2), with my local patch removed.

The lifecycle part works for local inputs:

- the GH5389 reproducer now gets both `on_stop` and `on_collect`;
- 100 replacements blocked inside `Av.open_input` ended with all 100 retired sources stopped/collected and only the current connection remaining;
- 50 explicit `stop()` / `start()` cycles completed (`connects=51 disconnects=50 stops=50`);
- the 3200-second seek regression still returns `3200.0` under a 1 GiB cgroup (239 MiB peak).

However, I found a blocking regression for every successfully connected remote `input.ffmpeg`.

The interrupt callback installed on the `AVFormatContext` is not limited to `Av.open_input`; FFmpeg keeps using it for later reads. As soon as this CAS succeeds:

```ocaml
Atomic.compare_and_set source_status `Polling (`Connected (...))
```

this condition becomes permanently true:

```ocaml
self#source_status <> `Polling
```

so the next network read is interrupted. The input reports partial/corrupt packets and `Ffmpeg_decoder.End_of_file`, then reconnects continuously.

I reproduced it with a generated 60-second H.264/AAC MP4 served by a local Caddy server (including proper HTTP range support):

```text
PR parent 1935cbb, 8 seconds:
RESULT connects=1 disconnects=0 errors=0

PR head e68b3e7, same file/server/script, 8 seconds:
RESULT connects=66 disconnects=65 errors=65
```

The repeated error is:

```text
Packet corrupt
Invalid NAL unit size
partial file
Feeding failed: Ffmpeg_decoder.End_of_file
```

A representative HLS pipeline consequently reached 99 threads within three seconds and segfaulted after four seconds without completing its first media playlist.

Minimal shape of the reproducer:

```liquidsoap
settings.frame.video.default := true
settings.frame.video.width := 640
settings.frame.video.height := 360
settings.frame.video.framerate := 24
settings.frame.audio.samplerate := 48000

connects = ref(0)
disconnects = ref(0)
errors = ref(0)

s = input.ffmpeg(start=true, "http://127.0.0.1:PORT/test.mp4")
s.on_connect(synchronous=true, fun () -> connects := connects() + 1)
s.on_disconnect(synchronous=true, fun () -> disconnects := disconnects() + 1)
s.on_error(synchronous=true, fun (_) -> errors := errors() + 1)

safe = mksafe(s)
clock.assign_new(sync="cpu", [safe])
output.dummy(safe)

thread.run(
  delay=8.,
  fun () -> begin
    print("RESULT connects=#{connects()} disconnects=#{disconnects()} errors=#{errors()}")
    shutdown()
  end
)
```

Checking specifically for `Stopping` / `Stopped` instead of “not Polling” would avoid interrupting normal connected reads. Preserving the immediate stop/start race protection probably needs a cancellation flag or generation token tied to each individual connect attempt, because the shared status can already be `Starting`/`Polling` again for a newer task while the old `open_input` is still returning.

So the cleanup diagnosis and synchronous `on_sleep` change look good, but I cannot validate this head for remote inputs yet.

toots:
Good catch, and the diagnosis is exactly right: the interrupt callback is installed on the `AVFormatContext`, not on `Av.open_input`, so keying it on the shared source status interrupts every read the moment the connect CAS lands.

Fixed in cd0c065. The per-attempt `closed` flag already had the right scope, it just was not reachable from `disconnect` before the container existed, so `` `Polling `` now carries it:

```ocaml
| `Polling of bool Atomic.t
```

`connect_task` allocates `closed` up front and binds `polling = `Polling closed` once, using that same value as the CAS witness for both `` `Starting -> `Polling `` and `` `Polling -> `Connected ``. The interrupt is back to `shutdown || closed` and nothing else. `disconnect` on `` `Polling closed `` raises `closed`, which is what interrupts a connection stuck in `open_input`, then moves to `` `Stopping `` so the task finalizes itself.

That also answers the generation-token point: `` `Polling closed `` is a distinct block per attempt and `Atomic.compare_and_set` compares physically, so an older attempt returning while the source is already `` `Starting ``/`` `Polling `` again for a newer one can never match, and cannot clobber it.

Your reproducer shape, 60s H.264/AAC over a local server, 8 seconds:

| | connects | disconnects | errors |
| --- | --- | --- | --- |
| e68b3e7 | 5 | 4 | 4 |
| cd0c065 | 1 | 0 | 0 |

`tests/regression/GH4840.liq` already covers this, which is the reassuring part: it serves a stream over harbor and reads it back with `input.http`, and it fails on e68b3e7 and passes on cd0c065. So CI would have caught it, no new test needed. `GH5389`, `GH4711`, `GH3303` and `streams/ffmpeg-errors` also pass, and 50 explicit `stop()`/`start()` cycles over harbor give `connects=50 disconnects=50 stops=50`.

Would appreciate a re-test on your side, particularly the HLS pipeline.

toots:
@oneingan sorry I deleted your comment by mistake. Still gonna addressed the issue though. Feel free to re-post if you still have it.

oneingan:
Retested after `cd0c065dda465b79c9872861849d8e1953ffd131`, using the exact merged commit `138d0f8d87e41efd6f69c2a7aa6232297910a055` (OCaml 5.5, FFmpeg 8.1.2) with my downstream patch removed.

The per-attempt token fixes the remote-input regression I reported:

```text
8-second HTTP H.264/AAC test:
connects=1 disconnects=0 errors=0
```

For comparison, the original PR head had produced 66 connects / 65 disconnects / 65 errors in the same test.

The rest of the validation also passes:

- the GH5389-style stop/collection regression test passes;
- 100 connected HTTP replacements: `created=101 connected=101 disconnected=100 stopped=100 collected=100`, with one TCP connection left;
- 100 replacements while `Av.open_input` was blocked: all 100 retired source IDs were observed in both `on_stop` and `on_collect`, no retired input connected, and only the current polling connection remained;
- representative HLS integration: 91 programme switches and 90 seeks over 180 seconds, no fatal/decoder errors, 19–20 FDs, one established TCP connection and no `CLOSE_WAIT`; the six-segment fMP4 playlist decoded as H.264/AAC with an approximately 53 ms A/V start offset;
- the 3200-second seek still returned `3200.0` under a 1 GiB cgroup;
- the script checks and full upstream CI are green.

Memory approached a plateau under the deliberately extreme two-second replacement cadence (about 271 MiB at 87 seconds and 282 MiB at 180 seconds). Thread count oscillated rather than accumulating; the higher absolute range is expected from #5388 making FFmpeg decoding/encoding/scaling use all available cores.

This validates the merged lifecycle fix for the original remote `source.dynamic` workload. Thanks!

PR #5412, opened 2026-09-16
ffmpeg: grow input stream contexts as the demuxer adds streams

charles-rc:
## Problem

Demuxers without a header, such as MPEG-PS, add streams to `format_context->streams` while packets are read. The input side sizes its per-stream arrays from `nb_streams` once:

- `allocate_input_context` allocates `av->streams` at the first read. `close_av` and the decoder flush in `ocaml_av_seek_native` then loop to the current `nb_streams` and read past the end of the array, and `open_stream_index` can write past it for a stream found later.
- On `main`, `av->stream_decoders` is allocated in `ocaml_av_open_input` before `avformat_find_stream_info`, when a header-less input usually reports no streams, so `open_stream_index` reads past its end on the first read.

We hit the first one in production on 2.4.6 (ocaml-ffmpeg 1.3.1): a valid 320 kbps mp3 whose payload happens to contain enough `00 00 01 Cx` sequences to be probed as MPEG-PS. It plays to the end, then Liquidsoap segfaults in `ocaml_av_close` -> `avcodec_free_context` with `nb_streams = 25` and a garbage `stream_t *`.

## Fix

- `av_t` records the sizes of `av->streams` (`nb_allocated_streams`) and `av->stream_decoders` (`nb_stream_decoders`).
- `allocate_input_context` grows `av->streams` to the current stream count, zeroing new entries, and is called before each input-side index. `ocaml_av_read_input` no longer caches the array pointer since opening a stream can reallocate it.
- `close_av` and the seek flush loop to `nb_allocated_streams`; `new_stream` keeps it in step on the output side.
- `open_stream_index` only reads `stream_decoders` for streams that existed when the per-stream configuration ran; later streams use the default decoder.

## Test

`test_input_streams_grow` reads a generated MPEG-PS with one stream at open and four more that start after the probe and end before the duration tail read, checks the stream count grew, then closes the input.

- Before: segfaults on `main` at the first read, and on 1.3.1 at close (5/5 runs).
- After: passes on `main`, and the same change on 1.3.1 closes cleanly. Liquidsoap 2.4.6 built against the patched 1.3.1 also plays the original file repeatedly without crashing, where the stock build segfaults at the first end of track.


charles-rc:
cc @toots a small bug fix for you!

toots (approved):
Makes sense: we assumed the number of allocated streams on the binding side match the one on the ffmpeg side but, in fact, the ffmpeg side can grown from under us leading to reading past the effective number of allocated entries on our own. Thanks!

charles-rc:
@toots can this be backported to 2.4.x?

toots:
> @toots can this be backported to 2.4.x?

It's a bit complicated as `2.4.x` builds on an outdated `ffmpeg` binding `1.3.1`.

I could do a backport but I'm also about to start the `2.5.x` RC process.

Any chance y'all would try with `2.5.x`?

charles-rc:
We can definitely give it a go. The changes from 2.4.x -> 2.5 make it seem like quite a major bump though. So we would want to soak for quite a while.

PR #5427, opened 2026-09-19
Update ladspa.md

StaryMic:
Changed the listed command to a quick one-line grep to list LADSPA plugins since --list-plugins doesn't do that. Thank you Stefan in the IRC chat for giving me the command and for helping me get my own plugins listed!


toots (approved):
Thanks!

PR #5443, opened 2026-09-21
Fix infinite loop reading 64-bit MP4 box sizes

charles-rc:
### The bug

`Faad.Mp4.openfile_fd` never returns on some M4A files. It spins at 100% CPU, forever, in a five-syscall loop:

```
lseek(fd, 0, SEEK_SET)
read(fd, "\0\0\0\34ftyp", 8)
lseek(fd, 28, SEEK_SET)
read(fd, "\0\0\0\1mdat", 8)          <- size field == 1, so a 64-bit largesize follows
read(fd, "\0\0\0\0\0002h\344", 8)    <- the largesize
lseek(fd, 0, SEEK_SET)               <- back to the start, forever
```

### Cause

`mp4ff_read_int64` in `src/modules/synced/faad/src/mp4util.c` accumulates the value out of an `int8_t` buffer:

```c
int8_t data[8];
...
result |= ((uint64_t)data[i]) << ((7 - i) * 8);
```

Any byte above `0x7F` sign-extends to `0xFFFF...FFxx`, and the shift-and-OR then sets every higher bit. `mp4ff_read_int16`, `mp4ff_read_int24` and `mp4ff_read_int32` in the same file all cast `(uint8_t)data[i]` before widening; only the 64-bit reader is missing that cast. Upstream FAAD2 avoids it by declaring the buffer `uint8_t`.

`parse_atoms` then does:

```c
mp4ff_set_position(f, mp4ff_position(f) + size - header_size);
```

and loops `while (mp4ff_atom_read_header(...) != 0)`, which only ends on a short read. A corrupted size makes that sum wrap, and if it lands back on an offset already visited the walk restarts from there and never terminates.

In the file that prompted this, the largesize bytes are `00 00 00 00 00 32 68 e4`. The correct value is 3303652; the sign-extended value is 2^64 - 28. The `mdat` box begins at offset 28, so the next position is `28 + (2^64 - 28)`, which is exactly 0, and the parser walks the same three headers forever.

Landing precisely on 0 is a property of that file, but the size corruption applies to every 64-bit box header containing a byte above `0x7F`. Other files silently get a wrong size and stop the walk early with incomplete tags.

### The fix

Add the missing cast, matching the other readers in the file:

```c
result |= ((uint64_t)(uint8_t)data[i]) << ((7 - i) * 8);
```

I kept the `int8_t` buffer rather than switching it to `uint8_t` so `mp4ff_read_data`'s `int8_t *` parameter still matches without a cast at the call site, and so the function reads the same way as its three siblings.

### Verification

Built the module standalone with a small probe calling `Faad.Mp4.openfile_fd` followed by `Faad.Mp4.metadata`, against a file with a 64-bit `mdat`:

- before: hangs at 100% CPU indefinitely
- after: returns in milliseconds with the expected tags

Two controls confirm the mechanism rather than the file: padding the `mdat` by 128 bytes, so every largesize byte falls below `0x80`, removes the hang; padding by 1 byte, which keeps the sign extension but moves the wrapped target off zero, also removes it. Only the unmodified file loops.

Worth noting this is reachable from untrusted input. Any application that reads tags from user-supplied M4A files can be made to spin a core permanently on a file that is otherwise completely valid.


toots:
Thanks for this report. I think I'm gonna remove the mp4 decoder. It was always a hack and now ffmpeg covers it perfectly.

I'll recommend simply disabling it via the settings.

toots:
I removed faad..

